PR #1, opened 2020-06-20
Adds Resource Files

tomglenn:
This PR removes the requirement for the developer to create auto-loaded `global.gd` and `characters.gd` scripts. The reason for doing so is that they are both very common names which may result in clashes in some projects, and also that creating extra auto-loaded singleton classes to fuel the dialog system does not seem like the best approach.

This PR proposes a cleaner solution using custom Godot Resources.

Namely a `DialogResource` which contains a reference to an external Json file for the dialog, a `Dictionary` of key/value custom variables, and an `Array` of `DialogCharacterResource` references.

And a `DialogCharacterResource` which contains a Name, an Image and Color.

Instead of creating a `global.gd` script that returns custom variables and a `characters.gd` script returning a dictionary of characters, developers can now create individual resource files for each character and dialog.

Video demo here:
https://twitter.com/codewithtom/status/1274151135776051206

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,43 +4,26 @@ A simple Godot dialog addon to use in any kind of project.
 ![Screenshot](https://coppolaemilio.com/godot/dialog-screenshot.png)
 
 ## How to use:
-You first need to create a `global.gd` and a `characters.gd` script for storing variables.  
+You first need to create a `DialogCharacterResource` for each character you wish to have in your dialog.
+To do this, right click in your FileSystem and choose `New Resource`, search for `DialogCharacterResource` and create it.
+It's good practice to put these inside a `Resources/Characters` folder.
+You can assign each character a Name, Image and Color.
 
-Inside your `global.gd` add a dictionary variable called `custom_variables`.  
-`var custom_variables = {}`.
+Once you have created your characters it's time to create a `DialogResource` for your dialog.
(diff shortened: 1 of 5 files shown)


coppolaemilio:
First of all, thank you very much for the PR! I love having the characters as resources because it is so much more user friendly than just having a dictionary. I didn't know you could do that and it is a perfect fit for the add-on.

As a goal I wanted to always have all the placeholders included so you don't have to create extra stuff, in this case we would have to create the three characters resources for the intro so it doesn't fails on a test run.

Regarding the dialog, I'm not sure if having the need for external Json files is something I would like to have. Giving the possibility to set the dialog by code is a must for me.

If having an external resource could give us the chance to create a visual editor (in a similar way to this https://github.com/EXPWorlds/Godot-Dialog-System) I would like to explore the idea of resources for dialogs, but for now I think it just makes you require using external editors to create the json files and it is less user friendly. If both systems can co-exists that could also work.

So, I'll see if I edit a couple of things here and there and commit some extra changes. Overall I really like the direction! Thanks again!

tomglenn:
Hey Emilio, yeah I totally agree with regards to the external JSON. I had ported that functionality over from the existing `dialog_node.gd` but agree that a much better approach would be to have a Node Graph as part of the `DialogResource` that lets you map out the dialog in a visual format.

PR #3, opened 2020-10-04
Add a proper editor icon

willnationsdev:
This updates the editor icon to have a more professional look and feel. Borrowed an icon from flaticon, so I've added a credits document (which is required to freely use their icons).

Here's an example of what it looks like:

dialogic_editor_icon


coppolaemilio:
can you commit this one to the 0.5 branch?

PR #37, opened 2021-01-27
Make sure the mouse is always visible during selection

mindtonix:
No matter what mode the mouse was in before the dialog appeared, it has to be visible in order to select options during dialogs.

These changes makes sure the mouse is visible during selection of options, and reverts it back to the last state after selection is done (which may still be visible, depending on game type)

### DIFF
--- a/addons/dialogic/Nodes/dialog_node.gd
+++ b/addons/dialogic/Nodes/dialog_node.gd
@@ -1,6 +1,7 @@
 tool
 extends Control
 
+var last_mouse_mode = null
 var debug_mode = true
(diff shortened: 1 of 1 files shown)


mindtonix:
Maybe this could be implemented as signals instead. I now realise that i want to tap into several steps in the dialog progress from the level script, and maybe it should all be exposed as different signals emitted?

coppolaemilio:
you can totally add more signals, I think everyone would benefit from it. I'm still heavily working on the editor so I haven't been interacting with the nodes in a real "game"

Issue #38, opened 2021-01-31
Bug with selecting choices in nested questions


Crystalwarrior:
When questions are nested, pressing on one of the buttons can cause a bug
`Invalid type in function 'set_mouse_mode' in base 'InputDefault'. Cannot convert argument 1 from Nil to int.`
Seems like it's simply trying to set a mouse mode to "Null" which gives an error. A simple null check might fix it.

Version: master, commit 83a7a0efcfc99d453495ba91aaa83fe992ba7481

PR #40, opened 2021-01-31
Fix set_mouse_mode being called with null last_mouse_mode

Crystalwarrior:
I am not fully sure if this approach is ideal (as this dialogue system could be used in games without reliance on mouse input), however this should guarantee the mouse mode is not set to "null" at the very least.
Fixes https://github.com/coppolaemilio/dialogic/issues/38


coppolaemilio:
Thanks!

Issue #42, opened 2021-01-31
Using DialogNode makes the dialog object incorrectly sized


Crystalwarrior:
With this setup:
![image](https://user-images.githubusercontent.com/3470436/106386146-afea6700-63e4-11eb-849a-2e97af410211.png)
Whenever the game is launched, the dialog node appears disappeared - however, it is simply incorrectly sized. This is caused by `resize_main()` being called in `_ready():`

![WHXKM4z8MY](https://user-images.githubusercontent.com/3470436/106485963-345ee780-64c2-11eb-90a7-bcaad6476984.gif)

When looking into the scene structure using the `Scene/Remote` tab in the engine, it appears that the dialogue node was not destructed, however, it's size is incorrect and way bigger than it is supposed to:
![Godot_v3 2 4-rc1_win64_cBva1PdGZG](https://user-images.githubusercontent.com/3470436/106485987-3aed5f00-64c2-11eb-94ba-6600ea70c243.png)

coppolaemilio:
Can't reproduce.

Crystalwarrior:
> Can't reproduce.

Will make a video of it

Crystalwarrior:
> 
> 
> Can't reproduce.

I have an idea why - is your Dialogic development project's window size 2048 x 1200?
Since I'm using a window size smaller than that, perhaps that could be it.

AnidemDex:
Can't reproduce on 0.9

AnidemDex:
Hey @Crystalwarrior , I made [resizables dialogs](https://github.com/AnidemDex/dialogic/tree/canvas_hotfix), can you check if it works for you to make a PR that solves this issue?

coppolaemilio:
Responsiveness is not planned at the moment, but with new theme properties to make the dialog size smaller and a fix to the resizing on '2D' mode I will close this ticket for now. If responsive sizing is required later in the future a new issue can be created. Thanks!

PR #43, opened 2021-01-31
Fixes incorrect dialog node sizing when launching program

Crystalwarrior:
Fixes https://github.com/coppolaemilio/dialogic/issues/42


coppolaemilio:
I don't understand what you are attempting to fix by removing this. This was added so the size is properly set from the start, if you don't do this in the ready event it will only happen when being resized, but if the window is never resized it will never adapt to the entire screen.

Crystalwarrior:
It doesn't work as intended on my end, if it works on yours idk why, worth investigating

PR #54, opened 2021-02-17
Modify dialogic.gd to fix a performance issue

### LINKED ISSUE #53: Dialogic plugin makes Godot consume more resources than neccesary
#### Plugin version
[0.9](https://github.com/coppolaemilio/dialogic/releases/tag/0.9)
#### Device
Windows 10 x64 (RAM 4Gb, Intel HD graphics 4000)
## Issue description:
While using the plugin, I noticed that my computer was working slower than usual. I assumed it was just windows background tasks (because windows y'know). After closing them, the slow work still there, so i opened up the task manager and notice something interesting:
image

My ram being drowned by Godot. Things goes worse when i excecute only a `Node` with `Dialog.tscn` as child:
image
> Note: % is the CPU usage.

(Just to be clear, i used the same scenary for all my test. An open `Scene` as the main scene with a `Node` as root and a `Dialog.tscn` as child of `Node`. This is usually the [consumed ram without the plugin](https://media.discordapp.net/attachments/785486110670716928/811615113781444619/unknown.png) and the [consumed ram without the plugin but `Dialog.tscn` as child](https://media.discordapp.net/attachments/785486110670716928/811615359873318963/unknown.png))

Playing a little with the code, I managed to solve this problem ~~(kinda)~~:

* Apparently, the Dialogic's [lines 20 and 21](https://github.com/coppolaemilio/dialogic/blob/master/addons/dialogic/dialogic.gd#L20) could be the cause of the problem (_but I can't figure out why anyway_).
* I don't know if this issue is related to #44 , since reading and wrinting many things on disk can cause some problems.

### Steps to reproduce
1. Create a new project
2. Create a new scene with a `Node` as root
3. Import and activate the plugin
4. Make `Dialog.tscn` child of `Node`

[example_project.zip](https://github.com/coppolaemilio/dialogic/files/5997795/example_project.zip)

AnidemDex:
Fixes #53 

Again, i don't know why this works. But it works.

Those are some test made with the 0.9 version and the version on this branch.

### Profiler
image
### TaskManager
image

### Profiler
image
### TaskManager
image

### DIFF
--- a/addons/dialogic/Nodes/dialog_node.gd
+++ b/addons/dialogic/Nodes/dialog_node.gd
@@ -1,4 +1,3 @@
-tool
 extends Control
 
 var last_mouse_mode = null
--- a/addons/dialogic/Nodes/dialog_node.gd
+++ b/addons/dialogic/Nodes/dialog_node.gd
@@ -1,3 +1,4 @@
+tool
 extends Control
 
 var last_mouse_mode = null
--- a/addons/dialogic/dialogic.gd
+++ b/addons/dialogic/dialogic.gd
@@ -17,8 +17,15 @@ func _enter_tree() -> void:
 	get_editor_interface().get_editor_viewport().add_child(_editor_view)
 	# Hide the main panel. Very much required.
 	make_visible(false)
-	DialogicUtil.init_dialogic_files()
-	get_editor_interface().get_resource_filesystem().scan()
+
+
+func _ready():
+	if Engine.editor_hint:
+		# Make sure the core files exist 
+		DialogicUtil.init_dialogic_files()
+		# Force Godot to show the dialogic folder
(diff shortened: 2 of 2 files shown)


AnidemDex:
I just notice that you need the node inside the theme editor, so it needs to have `tool`. I'll look for another alternative

AnidemDex:
Apparently, using the DialogNode in background (inside the dialogic dock) creates another problem. It's solved in https://github.com/AnidemDex/dialogic/commit/5ba9408b61ed03e73dcec17a34ad449fddec1092

coppolaemilio:
Closing this one since I won't be working on the 0.9 for now. I'll soon merge the 1.0 branch to master as a "release candidate"

PR #59, opened 2021-02-19
Modify dialog_node and portrait script

### LINKED ISSUE #56: Creating a "Character Join" event without "position" gives an error
Maybe the title had to be renamed to `get_character_position() returns null`

## Issue
If you create a `Join` event with a character but without position, it gives you an error and breaks your game.
#### Dialogic timeline:
image
#### Debugger:
image
#### Json file:
image

This is due a problem with the `dialog_node` function `get_character_position()`. This returns null if all positions are `false`, but it should be returning an string to pass that to the `init` function of [`Portrait` node](https://github.com/coppolaemilio/dialogic/blob/master/addons/dialogic/Nodes/Portrait.gd#L13), because the second parameter of `init` expects an string, but the `dialog_node` pass a null reference (_because that's the actual behaviour_).

## Proposed solution
Playing with the code I got 2 possible solutions
### 1
Let the function [`get_character_position()`](https://github.com/coppolaemilio/dialogic/blob/master/addons/dialogic/Nodes/dialog_node.gd#L508) return something:
```python
func get_character_position(positions) -> String:
	if positions['0']:
		return 'left'
	if positions['1']:
		return 'center_left'
	if positions['2']:
		return 'center'
	if positions['3']:
		return 'center_right'
	if positions['4']:
		return 'right'
	return 'none'
```
And then handle that string inside the [`init`](https://github.com/coppolaemilio/dialogic/blob/master/addons/dialogic/Nodes/Portrait.gd#L13) function of `Portrait` node
```python
func init(expression: String = '', position_offset = 'left') -> void:
	if position_offset == 'none':
		return # This line will not work as expected
```
or modify the `positions` variable
```python
var positions: Dictionary = {
	'left': Vector2(-400,0), 
	'right': Vector2(+400,0), 
	'center': Vector2(0,0), 
	'center_right': Vector2(200,0), 
	'center_left': Vector2(-200,0),
        'none': Vector2.ZERO # or whatever you like
}
```
### 2
Handle the null value directly inside the [`init`](https://github.com/coppolaemilio/dialogic/blob/master/addons/dialogic/Nodes/Portrait.gd#L13) function of `Portrait` node
```python
func init(expression: String = '', position_offset = 'left') -> void:
	if position_offset is null:
		return # This line will not work as expected
```

If you ask to me, I prefer the first option modifying the positions variables

AnidemDex:
This is the proposed solution that fixes #56

### DIFF
--- a/addons/dialogic/Nodes/Portrait.gd
+++ b/addons/dialogic/Nodes/Portrait.gd
@@ -6,11 +6,21 @@ var character_data = {
 	'color': Color(0.973511, 1, 0.152344),
 	'file': ''
 }
-var positions = {'left': Vector2(-400,0), 'right': Vector2(+400,0), 'center': Vector2(0,0), 'center_right': Vector2(200,0), 'center_left': Vector2(-200,0)}
+# Changed
+var positions = {
+	'left': Vector2(-400,0), 
+	'right': Vector2(+400,0), 
+	'center': Vector2(0,0), 
+	'center_right': Vector2(200,0), 
+	'center_left': Vector2(-200,0),
+	'none': Vector2.ZERO
+}
+
 var direction = 'left'
 var debug = false
 
 func init(expression: String = '', position_offset = 'left') -> void:
+	
 	rect_position += positions[position_offset]
 	direction = position_offset
 	modulate = Color(1,1,1,0)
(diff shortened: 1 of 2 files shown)


coppolaemilio:
Hey, I've fixed this in a slightly different way so I will close it. But thanks!

Issue #68, opened 2021-03-09
More signals


GammaGames:
It would be nice if you could set the game to emit more signals. Like:

- When a character enters/leaves
- When a character speaks
- During major events (dialogue start, wait, end)

The default signal is very handy for triggering animations, but sometimes I might want my characters to wiggle or have a little text bubble appear above their head when speaking (even custom for different moods!). I could probably do this with the current setup, but it'd clutter my timeline quite a bit.

I know I'm partially using the addon for more than it's original use-case, but it's got so many features and is so flexible!

GammaGames:
This also might be a duplicate of #67? I can move my comments over there if I need to

Issue #67, opened 2021-03-09
More than one signal/function?


choinheap:
Hey, started using dialogic for a quick game jam, seems really good for cutscenes, only that I'm not making a visual novel but rather a 3D RPG with a lot of animations. 

Love that I can combine my own functions via signals inside the timeline but unless I'm missing something it seems like I can only set the value and not which function or different signal to use unless it's a different node. Been using if statements for whichever value I set to do different things within the function. But I'm wondering if I'm approaching this wrong and not understanding the intent here.

coppolaemilio:
At the moment there is only that one signal event, so what you are doing with conditionals is how it is supposed to be done. Of course this is not ideal so I'll see what I can do to trigger more signals from the dialog events and more custom "emit signal" calls

PR #70, opened 2021-03-09
WIP signals and debug mode variable

GammaGames:
* Add `.gitattributes` so windows doesn't mess up line endings
* Expose `debug_mode` in scene, add variable to `start` function (default true)
* Add some signals:
  * `dialogue_start()` (NOT WORKING YET)
  * `dialogue_end()` (NOT WORKING YET)
  * `text_event(event)`
  * `wait_event_start(seconds)`
  * `wait_event_timeout()`

This should, for the most part, address #68 (and maybe #67).

More signals should be added to make it more complete, and I'm not sure where to put the start and end signals to make them work. Should be understandable if I read the code a bit!

I ran into a bug where I would try to connect the `text_event` signal in my scene's `_ready` function, but the first event wasn't connected in time. I should be adding the dialogue via code anyway, but it's something to keep in mind.

My editor (vscode) also strips all whitespace automatically, so there'll be some tabs cleared out. I can try to add them back if you want them.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+* text=auto
+* text eol=lf
--- a/addons/dialogic/Nodes/dialog_node.gd
+++ b/addons/dialogic/Nodes/dialog_node.gd
@@ -2,7 +2,6 @@ tool
 extends Control
 
 var last_mouse_mode = null
-var debug_mode = true
 var input_next: String = 'ui_accept'
 var dialog_index: int = 0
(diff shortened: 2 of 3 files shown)


coppolaemilio:
Thanks! but I wasn't going to keep pushing changes to the master branch. Do you want to start trying to work with v1.0? (make sure to make backups of your timelines/characters/glossary files before trying the latest version)

PR #71, opened 2021-03-09
signals and debug mode variable

GammaGames:
* Expose `debug_mode` in scene, add variable to `start` function (default true)
* Add a bunch of signals
  * Signals that pass the `event` dictionary have `_event` as their suffix
  * Not every event has a signal (`condition`, `scene`, `change_timeline`)

This should, for the most part, address #68 (and maybe #67).

If you connect the `timeline_start` signal via the `_ready` function for a dialogue scene that exists in the editor tree, it might not fire correctly because of timing. My solution for this is to add the node via the `Dialogic.start()` method.

Note: I have VSCode set to strip all trailing whitespace automatically, I can try to undo that if you want.


GammaGames:
Hmm, something that would make the signals more useful is if we have a `get_resource` function in the utils that returns the JSON for a given id (`character-1615177272.json`). Would make it a bit easier to get whatever info I wanted from a signal.

GammaGames:
Also, if you feel that the signals should be emitted at the *end* of their `event_handler` match, that can be fixed. I'm not sure how Godot handles signals by default

coppolaemilio:
There are way too many signals now. I think it should be just one for all the events and send what event is happening on an argument instead of in the name for the signal.
Something like `event_start` and send `text_event`

GammaGames:
Thank you! I knew it was complicated and a bit wonky, keeping the signal count low is good. It now uses an `event_start` and `event_end` signal, the type is just `text`, `wait`, etc.  
`event_start` passes the event (or timeline id), while the end signal only passes the type.

It'll be a bit cleaner to use a dispatch map in 4.0, but for now a simple `match` statement works fine :)

coppolaemilio:
Main issue with the previous approach would be that I'll make it as easy as possible to make new events later on, so they need to be as self contained as possible

GammaGames:
I agree. For example, the `event_end` signal only has a few types because adding extra internal signal connections would clutter it and I'm not sure how useful most of them would be anyway. This felt like a nice starting point that could be expanded upon in the future if it was needed.

coppolaemilio:
Thank you so much!

Issue #80, opened 2021-03-15
Allow more IF conditions


arnaudvergnet:
* Dialogic version: rc_1.0, commit 853f991d8089777b9aff2607310c7251c04086da

# Summary

![Screenshot_20210315_183139](https://user-images.githubusercontent.com/80701113/111195572-b0b61180-85bc-11eb-90f6-a9834283f311.png)

The menu button does not allow choosing other conditions than `Equal to`.

# The problem

## What is the current behavior?

We can only make `Equal to` conditions

## What is the expected correct behavior?

Use more conditions such as `<`, `=<`, `>`, `>=`

arnaudvergnet:
Fixed in https://github.com/coppolaemilio/dialogic/commit/cf1b272d7c62cf7d59126a561dbb35a70f6995ca

PR #81, opened 2021-03-16
Add more 'if' conditions

arnaudvergnet:
This pull request adds support for more conditions in if statements, as requested in issue https://github.com/coppolaemilio/dialogic/issues/80.

This adds a new scene, `ConditionPicker` to allow selecting conditions, used only in the `IfCondition` Scene. It follows the same structure as the `DefinitionPicker` scene.

Conditions can be saved and loaded back like definitions and values.

Supported conditions are:

```
==
!=
>
>=
<
<=
```

Screenshot_20210316_184140

### DIFF
--- a/addons/dialogic/Editor/Pieces/Common/ConditionPicker.gd
+++ b/addons/dialogic/Editor/Pieces/Common/ConditionPicker.gd
@@ -1,0 +1,58 @@
+tool
+extends MenuButton
+
+var options = [
+	{
+		"text": "[ Equal to ]",
+		"condition": "=="
+	},
+	{
+		"text": "[ Different from ]",
+		"condition": "!="
+	},
+	{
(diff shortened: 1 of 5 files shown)


coppolaemilio:
Looks great! I think that for the `>`, `>=`, `<`, `<=` it should try to transform the variable to ints/floats since they are all stored as strings now. Do you want to make that change as well or should I take over?

arnaudvergnet:
Oh yes didn't think about that (not used to dynamic languages)! I don't have time today to edit it so if you want to go faster you can take it from here.

I think the best approach would be to try to convert to numbers on both sides, if one fails stick to string. This should also be documented to avoid problems.

coppolaemilio:
Will do! Thanks for the commit!

PR #85, opened 2021-03-17
Add dialog_scene_path param

GammaGames:
This allows users to use their own custom dialog scene. The scene should probably be inherited from the base `Dialog.tscn` to ensure compatibility.

I also set the `debug_mode` variable to false by default and made it the third param. If a user wants to set it to `true` and not pass in a custom scene, then user can set it to the node that is returned by the method. It would be a lot easier if Godot had kwargs, but it doesn't yet.

### DIFF
--- a/addons/dialogic/Other/DialogicClass.gd
+++ b/addons/dialogic/Other/DialogicClass.gd
@@ -2,7 +2,7 @@ extends Node
 class_name Dialogic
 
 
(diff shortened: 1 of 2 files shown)


GammaGames:
If you'd rather `debug_mode` left as true I can undo that change, it's not *really* important but I tend to use the console for lazy debugging and prefer a cleaner default.

coppolaemilio:
Thanks! I think that debug off by default is good 👍🏻

PR #89, opened 2021-03-19
fix number comparison

arnaudvergnet:
If condition and definition values are valid floats,
then make the conversion on floats.
Otherwise, fallback to string comparison.

This adresses the issue we talked in the [PR](
https://github.com/coppolaemilio/dialogic/pull/81#issuecomment-800489767)


coppolaemilio:
thanks!

PR #92, opened 2021-03-19
Show glossary defs in questions

arnaudvergnet:
Adds the ability to show glossary definitions inside a question text.
This behaves exactly like in a classic text event


arnaudvergnet:
When this PR is accepted, I have an other one coming (which depends on this one) allowing the use of inline variables

Issue #84, opened 2021-03-17
Inline glossary values (e.g. player character’s name)


JapanYoshi:
It would be great if Dialogic let me plug values in the Glossary into lines of text, kind of like this:

- 💬 **What do you want to have for lunch?**
- ❓
  - { **Pizza**
    - 📝Set **FOOD** to be **pizza**
    - 📝Set **TOPPING** to be **pineapple and anchovy**
  - { **Tacos**
    - 📝Set **FOOD** to be **tacos**
    - 📝Set **TOPPING** to be **red onion and cilantro**
- }
- 💬 **[[FOOD|cap]], yum! I’ll make sure to order some with lots and lots of [[TOPPING]].**

It would be even more useful if free answer questions were supported in Dialogic natively.

PR #93, opened 2021-03-19
add inline variables support

arnaudvergnet:
Allows showing inline variables using `${variable_name}` inside a text message.

Parsing takes place before each text update, not just in the initial timeline loading.
This allows showing dynamic variables inside a dialogue.

Variable values are inserted before glossary parsing, allowing dynamic glossary text.

For example, if `test_value`'s value is `test_gloss`, then putting `${test_value}` inside a text/question will show test_gloss and enable the glossary popup on the word.

Might help solve #84.

### DIFF
--- a/addons/dialogic/Nodes/dialog_node.gd
+++ b/addons/dialogic/Nodes/dialog_node.gd
@@ -235,7 +235,7 @@ func _insert_variable_definitions(text: String):
 	for d in definitions:
 		if d['type'] == 0:
 			var value = DialogicUtil.get_definition_value(d, runtime_id)
-			final_text = final_text.replace('${' + d['name'] + '}', value)
+			final_text = final_text.replace('[' + d['name'] + ']', value)
 	return final_text;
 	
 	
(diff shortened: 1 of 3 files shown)


addons/dialogic/Nodes/dialog_node.gd
```diff
@@ -219,28 +218,27 @@ func parse_branches(dialog_script: Dictionary) -> Dictionary:
...
+	final_text = _insert_variable_definitions(text)
+	final_text = _insert_glossary_definitions(final_text)
+	return final_text
 
+func _insert_variable_definitions(text: String):
+	var final_text := text;
+	for d in definitions:
+		if d['type'] == 0:
+			var value = DialogicUtil.get_definition_value(d, runtime_id)
+			final_text = final_text.replace('${' + d['name'] + '}', value)
```
coppolaemilio, line 238: Was there any reason to use the `&{` and `}`? I would prefer something simpler like just regular brackets `[` `]`
```suggestion
			final_text = final_text.replace('[' + d['name'] + ']', value)
```

arnaudvergnet:
I used `${}` in the first place because I saw some programming languages (such as JS) doing it, but yes this is easy to change and we could even make this a setting the user could change

PR #97, opened 2021-03-21
Improve resources handling and allow saving definitions

### LINKED ISSUE #76: Variables and saves.
Right now Dialogic is not aware of what variables to save for savegames, so I need to implement a better `definition` state and provide with better functions to save/load sets of `definitions`.

### LINKED ISSUE #94: Definitions broken on HTML5?
I can't make it work, but there are no console errors!
image
If later, at the same timeline, I try to use "a" with
image
Nothing happens.

### LINKED ISSUE #95: Variables not working on exported projects.
It seems that when you save files to a location in `res://` instead of  `user://` they cannot be changed in exported games. I think we should make a copy of the definitions.cfg in the `user://` path to allow for persistent data. 

I'm tagging @arnaudvergnet since you were working on this part of dialogic already

arnaudvergnet:
Greatly improve modularity by regrouping all resources related handling inside a static class.

What changed:
* Greater abstraction: paths, configs, json are all handled by `DialogicResources` to prevent errors
* No more `DialogicSingleton`: It won't be useful as we are migrating to saves in `user://`
* Saves: Able to save definitions, and to reset them, from code or inspector.

Tell me what you think of these changes and if you find any bugs.

Closes #95
Closes #94 
Closes #76


coppolaemilio:
I'm going to test this one now

arnaudvergnet:
ah yes did not try on an empty project, will fix that

coppolaemilio:
Other than those 3 I can't find much more, but I'll keep testing to find more stuff. Other than that the PR looks great! and exported game resources seem to work, but the variable bug is still there.

arnaudvergnet:
I fixed the reported errors but it might need more testing

coppolaemilio:
I'll get to test right away 👍

coppolaemilio:
Still getting error messages (print messages like `Error while accessing path: ... - Error 31`) on startup but it works now but it works almost perfectly! 
image

Another error I get is when using the `Change Timeline` event is:
` res://addons/dialogic/Editor/Pieces/ChangeTimeline.gd:31 - Invalid get index 'working_timeline_file' (on base: 'HSplitContainer (TimelineEditor.gd)').`

But I'll probably merge this changes so `main` is in a better place than now.

Thank you so so much!

PR #103, opened 2021-03-23
Implement two-step start to allow timeline data manipulation

linux-man:
I can imagine a lot of situations where "adding one more feature" to the visual editor will be too much, and too many features will make this plugin hard to use for "normal" users. This is a kind of "advanced user" tool. Dialogic is called with `stop_on_load=true`, when timeline and basically all dialog_node data can be edited before a real "start".
Code example - dynamic portrait change:
```GDScript
	var new_dialog = Dialogic.start(timeline, "res://addons/dialogic/Dialog.tscn", false, true)
	for item in new_dialog.dialog_script.events:
		if item.get("action") == "join" && item.get("portrait") == "p1":
			item.portrait = $"/root/Player".sprite_string
	new_dialog.start()
```

### DIFF
--- a/addons/dialogic/Nodes/dialog_node.gd
+++ b/addons/dialogic/Nodes/dialog_node.gd
@@ -19,6 +19,7 @@ var current_theme
 export(String, "TimelineDropdown") var timeline: String
 export(bool) var reset_saves = true
 export(bool) var debug_mode = true
+export(bool) var stop_on_load = false
 signal event_start(type, event)
 signal event_end(type)
 signal dialogic_signal(value)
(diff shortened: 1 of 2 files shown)


coppolaemilio:
I tried to understand it but I don't know why you need this 🤔

linux-man:
Maybe it's just me...
I want the player to select his image.
I have dialogs that are randomly used by several Characters (with different portraits).
At this time is there any way to implement such behaviours on Dialogic?

coppolaemilio:
I will add an option to set a scene instead of an image for a portrait. That way you can make a custom scene with whatever values you want or the player pick and dynamically show something different. Would that work for you?

linux-man:
Of course. Until then I will use my fork. Keep up the great work.
Over and out.

PR #104, opened 2021-03-23
Synchronize definitions and timeline saving

### LINKED ISSUE #102: Implement a better saving system
* **Dialogic version**: rc_1.0
* **OS**: Ubuntu 20.04

## Summary

It is currently quite difficult to implement a robust saving system.

## The problem

**Steps to reproduce**

* Create a new game with this plugin
* Create a new definition `value = 0`
* Create a new timeline with a `if value == 1` then a `set value = 1`
* Play a first time and exit before the end, it won't go in the `if` branch
* Play a second time, it will go inside

**What is the current behavior?**

Definitions are saved each time they change. This can work in most cases, but as we cannot go back in the middle of a timeline, it can break the integrity of values.

**What is the expected correct behavior?**

Definitions should be saved at the end of a timeline, or when specified by the user.

## Solutions

**Possible fixes**

A possible solution could be to store current definitions in memory in a singleton, and only write to the config file in `user://` at the end of a timeline or when the user demands it.

This way the user will be able save the current timeline id, and get back to it without breaking values, even if the player quits in the middle of a dialog.

arnaudvergnet:
Definitions and current timeline are kept in memory inside a singleton, and saved to disk when needed.

This data is only saved to the config file on timeline end/start.

While in the editor, default values are loaded as I can't seem to make the saves singleton load.

User can request a definitions reset either by a dedicated function, or by passing an argument to the start method.

Current timeline is saved so user can read it while running or load it back on next run. This feature combined with definitions saving makes a complete save/restore system, in which users can exit mid-timeline without negative effects.

I also added a lot of utility functions in `DialogicClass.gd`, to allow the user to edit and query variables and glossary definitions. It only repeats functions in the singleton, but this way the user only has one class to worry about and it prevents him from accessing unsafe methods.

Closes #102

### DIFF
--- a/addons/dialogic/Other/DialogicDefinitionsUtil.gd
+++ b/addons/dialogic/Other/DialogicDefinitionsUtil.gd
@@ -1,0 +1,55 @@
+extends Node
+class_name DialogicDefinitionsUtil
+
+static func get_definition_key(config: ConfigFile, section: String, key: String, default):
+	if config.has_section(section):
+		return config.get_value(section, key, default)
+	else:
+		return default
+
+
+static func set_definition_variable(config: ConfigFile, section: String, name: String,  value):
+	config.set_value(section, 'name', name)
+	config.set_value(section, 'type', 0)
+	config.set_value(section, 'value', str(value))
+
+
+static func set_definition_glossary(config: ConfigFile, section: String, name: String,  extra_title: String,  extra_text: String,  extra_extra: String):
+	config.set_value(section, 'name', name)
+	config.set_value(section, 'type', 1)
+	config.set_value(section, 'extra_title', extra_title)
+	config.set_value(section, 'extra_text', extra_text)
+	config.set_value(section, 'extra_extra', extra_extra)
+
+
+static func add_definition_variable(config: ConfigFile, section: String, name: String, type: int, value):
+	config.set_value(section, 'name', name)
+	config.set_value(section, 'type', type)
+	config.set_value(section, 'value', str(value))
+
+
+static func delete_definition(config: ConfigFile, section: String):
+	config.erase_section(section)
+
+
+static func definitions_config_to_array(config: ConfigFile) -> Array:
+	var array := []
+	for section in config.get_sections():
+		var type = config.get_value(section, 'type', 0);
+		if type == 0:
+			array.append({
+				'section': section,
+				'name': config.get_value(section, 'name', section),
+				'value': config.get_value(section, 'value', ''),
+				'type': type,
+			})
+		else:
+			array.append({
+				'section': section,
+				'name': config.get_value(section, 'name', section),
+				'title': config.get_value(section, 'extra_title', section),
+				'text': config.get_value(section, 'extra_text', section),
+				'extra': config.get_value(section, 'extra_extra', section),
+				'type': type,
+			})
+	return array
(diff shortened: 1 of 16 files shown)


addons/dialogic/Other/DialogicDefinitionsUtil.gd
```diff
@@ -0,0 +1,55 @@
...
+		return default
+
+
+static func set_definition_variable(config: ConfigFile, section: String, name: String,  value):
+	config.set_value(section, 'name', name)
+	config.set_value(section, 'type', 0)
+	config.set_value(section, 'value', str(value))
+
+
+static func set_definition_glossary(config: ConfigFile, section: String, name: String,  extra_title: String,  extra_text: String,  extra_extra: String):
```
coppolaemilio, line 17: While reviewing the code I noticed that we could maybe save this glossary type variables as an array on the `value` key instead of creating the three `extra_title`, `extra_text` and `extra_extra`. What do you think?

arnaudvergnet: That is a good idea as it would remove differences in cfg structure between glossary and variables. The drawback is that it will break compatibility with previous version.

arnaudvergnet: But if we are making breaking changes, might as well use json right away instead of converting between json and config files.

coppolaemilio: We can keep it as is, but I was planning on breaking the compatibility as much as needed now before the actual 1.0 release. The time is now! haha
As you prefer, I'll keep testing this branch and see if there are any errors.

arnaudvergnet: Let's convert definitions to json then, it will be more in line with the rest of the plugin. Only settings and themes will be using cfg.

coppolaemilio: Sounds good to me 👍🏻

arnaudvergnet:
@coppolaemilio definitions are now saved in json. Hopefully I did not break anything. I didn't notice any bug but it needs more testing as this change is quite important

arnaudvergnet:
While I was at it, I also saved the state (current timeline) in json as well.

coppolaemilio:
Sorry I wasn't able to test until now. I'll do the usually test and QA stuff and try to merge soon 👍🏻

coppolaemilio:
Starting from an empty project:
image

arnaudvergnet:
As usual I do not test on empty project :) It is fixed

coppolaemilio:
Testing an empty project with no variables and just a timeline that has one text event:
image

arnaudvergnet:
Fixed both bugs!

The first one was because I initialized saves with an empty dictionary. Instead I initialized it with `{'variables': [], 'glossary': []}`, just like I did for default definitions.

As for the dialog not working with the editor node, this was because it returned the file name without `.json`. I changed the start method to allow for timeline name or filename, so I changed how the timeline variable was set with the start method but forgot about the inspector.

All

coppolaemilio:
Yes! they look fixed on my end 🎉 I'll merge now and keep looking for stuff. I'll need to really go back over all the changes because it is starting to feel too complex. I would love if we could simplify more stuff after v1.0 is stable

Issue #111, opened 2021-03-25
Better resources UI


arnaudvergnet:
**Is your feature request related to a problem? Please describe.**
When dealing with a lot of timelines, characters, definitions and themes, it takes a lot of place in the screen and you quickly get lost.

For example, in my project I have 13 timelines, and I am only at the beginning. Splitting my dialog into several timelines makes it easier to organize and maintain (and allows for the autosave to trigger often). However, timelines are sorted by creation date (filename using timestamp), and in a simple list. This can be difficult to find the correct timeline in a long list. 

**Describe the solution you'd like**
In my opinion, the best approach would be to let the user create 'folders' for resources and let him drag-and-drop resources to organize them.

We would also need to edit the pickers in the timeline. It should display a dialog showing the relevant resource's tree (like the `Create New Node` dialog) instead of showing a simple list.

To implement this behavior, a possible solution would be to create folders in the filesystem, and move resource files when the user moves them in the tree. This way the tree structure would be easy to do, and we could sort files by name (instead of filename) to make it easier to find resources.

**Describe alternatives you've considered**
I tried to use naming conventions with specific prefix for resources of a certain type, but as we cannot move them in the list, everything becomes mixed up as your project grows.

coppolaemilio:
The folders was something that was planned initially when I started unifying the item lists, but there is also another feature I want to implement that I never openly spoke about which is the "chapters" view.
The main idea is that each folder (chapter) would host a bunch of timelines and when you select one can see a different view with each chapter as a node in a graph node panel. That way you can better organize your story or branches, and also use new events like "jump to next chapter" etc... 

Moving items would be nice and could probably be accomplished, but haven't tried yet. I suppose you have to add an extra parameter with the place where each item is and load them in that order instead of alphabetically. 

I wasn't planning on implementing any of that for the v1.0 since I believe i18n is more requested and probably will break compatibility, so I'm researching that right now

arnaudvergnet:
I don't think I fully understand how you want to implement chapters, but I guess you'll be able to explain a bit better when you start working on it. Just be careful to keep the plugin compatible for general use, and not specific to visual novels.

I'll work on a PR to at least sort resources by name so we can organize them using a prefix system. I'm working a game (visual novel) with a lot of choices, I'll report back to you to see if it is easy to use.

Sorting using a tree seems like a lot of work as we would have to remake pickers, so I'll leave the idea for now.

coppolaemilio:
[...]

PR #112, opened 2021-03-25
sort resources by name if available

arnaudvergnet:
This makes it easier to find resources using prefix, as discussed in #111.

Trees are rebuilt each time a resource name changes. I tried to squash every bugs I could find, but as always, test and report anything you find!


coppolaemilio:
I'm getting all these errors on my existing test project
image

coppolaemilio:
I'm a bit worried that every commit lately is breaking the existing projects, and this can be a big issue when people would be updating from 1.0 to 1.1

I feel like the singleton and having so many classes is starting to be a bit hard to manage, but I'll investigate to see what steps you need to make to upgrade without having issues.

arnaudvergnet:
This commit should not be breaking changes, as it only impacts display. I managed to reproduce the issue and am working on a fix.

Having many classes is not especially a bad thing. It is better to have several classes with each their purpose than to regroup everything in one file.

The singleton seems to be quite annoying. It can only be used while running and if you change its name, you have to remove the previous one from project settings. But I don't know how you can keep saved data in memory without it.

coppolaemilio:
> The singleton seems to be quite annoying. It can only be used while running and if you change its name, you have to remove the previous one from project settings. But I don't know how you can keep saved data in memory without it.

Yeah, I didn't want to add any singletons at first but when I started thinking about solutions to save the state of the custom made variables it became apparent that it would be impossible without actually having a persistent node in the game. Maybe there is a way to remove previous ones from code if they exists around here: https://github.com/coppolaemilio/dialogic/blob/25de662f782430689acae36ef08b0ca9d6989663/addons/dialogic/dialogic.gd#L11

arnaudvergnet:
This should be fixed. For some reason godot was able to find the class used for sorting while I was working, but after a restart it couldn't find it anymore.

And if it can reassure you, I am working on my own game while contributing to the plugin, so any breaking change will impact me as well so I try to minimize them.

> Yeah, I didn't want to add any singletons at first but when I started thinking about solutions to save the state of the custom made variables it became apparent that it would be impossible without actually having a persistent node in the game. Maybe there is a way to remove previous ones from code if they exists around here: 

Might be worth investigating!

PR #113, opened 2021-03-25
Improve timeline editor controls

arnaudvergnet:
This PR adds buttons to fold/unfold all nodes in a timeline. It also adds the possibility to select pieces. When selecting a piece, it will change color and adding a node will add it just bellow the selected piece. Clicking again a node will deselect it.

I did all the logic inside `TimelineEditor.gd`, so this will work on any piece as long as it has a `PanelContainer` node just bellow the root.

I think we could do the same for the `DragController`: instead of adding the node on each piece, control everything from `TimelineEditor.gd`, this will be easier to maintain.

If you want to change how selected items appear, you can edit the style in `res://addons/dialogic/Editor/Pieces/selected_styleboxflat.tres`

This does not touch resources handling so it is not breaking.

### DIFF
--- a/addons/dialogic/Editor/EditorView.gd
+++ b/addons/dialogic/Editor/EditorView.gd
@@ -6,6 +6,7 @@ var editor_file_dialog # EditorFileDialog
 var file_picker_data: Dictionary = {'method': '', 'node': self}
 var current_editor_view: String = 'Master'
 var version_string: String 
+onready var master_tree = $MainPanel/MasterTree
 onready var timeline_editor = $MainPanel/TimelineEditor
 onready var character_editor = $MainPanel/CharacterEditor
 onready var definition_editor = $MainPanel/DefinitionEditor
@@ -24,6 +25,8 @@ func _ready():
 	definition_editor.editor_reference = self
 	theme_editor.editor_reference = self
 
+	master_tree.connect("editor_selected", self, 'on_master_tree_editor_selected')
+
 	# Toolbar
 	$ToolBar/NewTimelineButton.connect('pressed', $MainPanel/TimelineEditor, 'new_timeline')
 	$ToolBar/NewCharactersButton.connect('pressed', $MainPanel/CharacterEditor, 'new_character')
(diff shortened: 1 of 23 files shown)


coppolaemilio:
I really like it! I was going to do something similar with the themes, I wanted to make the outline change only, but godot themes don't really allow you to do that per instance, so all the same kind of events changed and ended up abandoning that branch haha

A few things: 
- Fold and unfold should be like it was in previous dialogic versions, in the toolbar instead of where the event buttons are
image

- When you have an item selected and you add a question or if condition the order of those new events gets reversed. I'm fine with removing the auto-adding of the choices and closing branch nodes when you click on those for this version.
image

- When selecting the text node an extra gap gets added to the left (because the theme for the theme is different to the rest so the text area is as wide as the event. I'm okay with leaving it like this for now, just wanted to flag it
image

arnaudvergnet:
> Fold and unfold should be like it was in previous dialogic versions, in the toolbar instead of where the event buttons are
image

I knew those functions were not here unused for no reason! Why did you remove this functionality ?

> When you have an item selected and you add a question or if condition the order of those new events gets reversed. I'm fine with removing the auto-adding of the choices and closing branch nodes when you click on those for this version.

Didn't see that, maybe I can fix this.

> When selecting the text node an extra gap gets added to the left (because the theme for the theme is different to the rest so the text area is as wide as the event. I'm okay with leaving it like this for now, just wanted to flag it

I pulled my hair out on this, and then abandoned. A dirty fix would be to create a selected theme specific for this control, but yeah this isn't a top priority bug.

I also noticed some controls did not allow you to click on them on empty space. While digging in the pieces, I saw some had a spacer with pass mouse events, and other with stop. So I created a new `Spacer` node passing mouse events, and used it in all pieces. So the number of files modified in this PR greatly increased but this didn't change much.

coppolaemilio:
> I knew those functions were not here unused for no reason! Why did you remove this functionality ?

I removed them when I unified the editor and completely forgot to re-add it later 😅 

> I pulled my hair out on this, and then abandoned. A dirty fix would be to create a selected theme specific for this control, but yeah this isn't a top priority bug.

I agree

> I also noticed some controls did not allow you to click on them on empty space. While digging in the pieces, I saw some had a spacer with pass mouse events, and other with stop. So I created a new Spacer node passing mouse events, and used it in all pieces. So the number of files modified in this PR greatly increased but this didn't change much.

Sounds good as well! 

Thanks again for your amazing work 🎉

arnaudvergnet:
> I removed them when I unified the editor and completely forgot to re-add it later

Understandable :)

> When you have an item selected and you add a question or if condition the order of those new events gets reversed. I'm fine with removing the auto-adding of the choices and closing branch nodes when you click on those for this version.

It is now fixed. The issue was present for if and choices as it added elements after the selected item, not at the end. I just told the editor to reverse the adding order if a piece is selected.

PR #118, opened 2021-03-26
add basic background music controls

### LINKED ISSUE #117: Control background music
Right now, we can only control audio event, but it would be nice to control background music as well.

We could create a new Audio node, just like the Audio Event node, that would play the given music in a loop and create crossfades when changing track.

arnaudvergnet:
This adds a new timeline piece allowing to set background music.

Music crossfade on change, and fade out on stop.

To stop music, simply put a BackgroundMusic node without a file selected.

Closes #117

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ The plugin is not production ready, this means that it will not work in your gam
     - Add offset to the portrait
   - Timeline Editor:
     - New `Theme event` to change the theme in the middle of a timeline
+    - New `Background Music Event` to play music in your dialog. Music can crossfade when changing track and fade in/out when starting/stopping.
     - Re-enabled the `Scene Event`
   - Theme Editor:
     - You can now add multiple themes
--- a/addons/dialogic/BackgroundMusic.gd
+++ b/addons/dialogic/BackgroundMusic.gd
@@ -1,0 +1,24 @@
+extends Control
+class_name DialogicBackgroundMusic
+
+onready var _anim_player := $AnimationPlayer
+onready var _track1 := $Track1
+onready var _track2 := $Track2
+
+
+func crossfade_to(audio_stream: AudioStream) -> void:
+	if _track1.playing and _track2.playing:
+		return
+	
+	if _track2.playing:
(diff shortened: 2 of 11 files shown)


coppolaemilio:
I have a few notes, but if you don't want to work on those I can merge and I'll fix later today: 
1. The text on the timeline editor button should have two spaces at the start (amazing padding styling, I know) so it should be:
     `··Background Music` instead of `Background Music`.
2. The theme should be the same color as the `Audio Event` to avoid the rainbow effect, but I don't really mind the difference since it is subtle.
3. When adding this kind of features it would be nice to add to the changelog so we don't forget to report what new features the plugin has.
4. The empty state of the event looks different to the empty state of the `Audio Event`
image
So ideally both should say "None" or none of them should have it. Both should have the delete icon.

arnaudvergnet:
> The text on the timeline editor button should have two spaces at the start (amazing padding styling, I know) so it should be:
··Background Music instead of Background Music.

Fixed!

> The theme should be the same color as the Audio Event to avoid the rainbow effect, but I don't really mind the difference since it is subtle.

I didn't want to use the exact same color to prevent errors, as the icon is already the same. I took the base color for Audio Events, but slightly darker to show it is a background sound, not a foreground sound. But yeah I can use the same color if you want.

> When adding this kind of features it would be nice to add to the changelog so we don't forget to report what new features the plugin has.

True, it's easy to forget!

> The empty state of the event looks different to the empty state of the Audio Event

This is because they behave differently. An empty audio event does nothing, while an empty background event stops the music. So there is no need to clear an audio event's track, but it might be useful for background. The ui is not that clear about this but I don't know how to explain it simply.

arnaudvergnet:
> The empty state of the event looks different to the empty state of the Audio Event

i implemented the same behavior in audio event and background music: an empty audio event stops the previous one. So both foreground and background events have text explaining their behavior when no track is selected. What do you think ?

coppolaemilio:
That looks good! I wasn't able to test with audio yet, but I'll merge and test later 👍🏻

PR #127, opened 2021-03-30
Translations proposal

AnidemDex:
# What is this?
This is my proposal to deal with internationalizations with Dialogic.

> I'll probably edit this request to add more information

I'll use [what Godot gives to us](https://docs.godotengine.org/en/stable/tutorials/i18n/index.html): the TranslationServer and its powerful translation importer (that just import translation files likes csv or po files to use them later as .translation resources).

I've worked with PO files, but for now, I'll focus test with CSV files (because it's faster than creating PO files just for test). You can see the progress of them in [this Google spreadsheet](https://docs.google.com/spreadsheets/d/1vOX4ffFq7AyYVm-KYDrctqdmJK2arUGcJrsSvj83cYI/edit?usp=sharing)

## Why?
If you're going to make a game with dialogues, you'll probably need to translate those text later to get more people involved into your game, or at least, let your fandom translate them for you (without decompiling your game for that).

> This is not my best argument, I'd have to admit

## Translations doesn't work inside the editor
Since [TranslationServer doesn't work in tool mode](https://github.com/godotengine/godot/issues/46271), I've worked on a [workaround to solve it](https://github.com/AnidemDex/Godot-TranslationService) and see the translations inside the editor. The version on this PR is a modified version to work with dialogic without messing things up.

## Problems
Now we have two main problems with this approach:
* Timelines text (and every text that can be translated) probably **will be broken with this approach** ~~(unless?)~~.
* Apparently, this approach is unmaintenable for larger projects. (I really don't know how)

## What's the plan?
1. Make the plugin itself be a multilingual plugin (the first on its class, since i can't find another plugin with more languages on it).
This can be done easily with the translation tool and the translation files. The hard part is to automatically import those translations.

2. Modify the Dialogic timelines to let them be _translatables_ without making them exploding in the process.
Yep, this is the hardest part.

_____
I really wanted to this be included with the 1.0 version, but I guess it's too late for that, so I'll let this PR here, open to your opinions.

At the moment that I upload this draft there's a proof of concept included with the recent commits. You can try an empty dialogic node without any timeline and see the result when there's no translation, one translation or more of them. (They doesn't add automatically to the translations, you had to add them by hand)


arnaudvergnet:
I never used translations inside Godot so I can't comment on the code itself, but while looking at the commits I didn't find any reference to glossary definitions. Did you find a way to handle this a well?

AnidemDex:
> I never used translations inside Godot so I can't comment on the code itself, but while looking at the commits I didn't find any reference to glossary definitions. Did you find a way to handle this a well?

@arnaudvergnet Yeah, i have a brief look to handle them using [DialogicDefinitionsUtil](https://github.com/coppolaemilio/dialogic/blob/main/addons/dialogic/Other/DialogicDefinitionsUtil.gd).
I'll translate the plugin first before dealing with the .json timelines.

This is a draft, so i'll keep commiting this branch until I got a working prototype to make the PR

AnidemDex:
I have paused the development of this branch in favor of an implementation of the code in [this branch](https://github.com/AnidemDex/dialogic/tree/resources_reformat).

The status of the current branch translates the plugin and the texts sent to the Dialogic node in the form of ID. More information [here](https://github.com/AnidemDex/dialogic/blob/translation/addons/dialogic/Nodes/dialog_node.gd#L330) and [here](https://github.com/AnidemDex/dialogic/tree/translation/addons/dialogic/Other/translation_service)

wpshadow:
Would it be a possibility to put a little checkbox with a label "translate" there and write only the translation-id inside the timelines text event?
**test_01** and checked would be enough with a method that gets the correct string from the translation server/service?
It's not the nicest solution but perhaps a quick workarround for now

AnidemDex:
> Would it be a possibility to put a little checkbox with a label "translate" there and write only the translation-id inside the timelines text event?
That was the first idea that i thinked, but making those text as ID would change totally the way that text is saved inside JSON and how the definitions are formatted, so i stopped there, at least, for now.

zarmath:
Could you use the same system that godot uses?, i.e. put for example [greeting] using square brackets in the text box and have it translated using the translation csv file?

AnidemDex:
> Could you use the same system that godot uses?

@zarmath It uses Godot TranslationServer, the same server that translates text when you use the `tr()` function. The problem it's just the way that the text will be saved. For now, I don't know if it's going to be saved the text as key for the CSV/PO files or there'll be a `translation_key` property in the JSON file.

PR #130, opened 2021-03-31
Added call-node feature

ellogwen:
Hi guys, I hope you are not mad at me, but I created a new event node for the timeline to call arbitrary node methods, because I needed it badly. I couldn't decide if I just create a workaround or just contribute instead. I do not want to take over your project, but maybe you find this handy and are considering to add it? It was not hard to do, so I am not mad if you decline it.

**Summary**:
I created a new event node call "Call Node". You can setup the node to call with a default Godot NodePath and a method name. In addition you can add arguments as well. The Timeline will execute those methods and even wait for completion, if the method in question is async and/or yielding.

**Usage**:
Add "Call Node" to your timeline. Change "Target Node" in GodotNodePath format. (e.g. `/root/Level/Player` or `./Triggers/DoorTrigger`) depending on your Scene setup. Add the name of the method to call as a string, as you would with Godots builtin `call` function. Optionally, add up to 99 arguments. Arguments get stored as strings, but you can cast them in your scripts like you do with defitinions. 

**Important**:
If you add arguments, the method in question gets called with an array, containing the arguments, as first parameter. But if you ommit the arguments, the method gets called without (an empty) arguments array. This is due to the way Godots `call` methods work. This way you can call parameterless builtin functions as well. As a future improvement, if I find a way to dynamically to spread the args instead of passing an array, I would prefer doing it that way (like the `...` operator in JavaScript). For now you cannot call builtin functions with paramters (e.g. `set_visible(true)`) and have to built a wrapper function to do so `func set_visible_arr(args): set_visible(bool(args[0]))` Does anybody know a way around this with the current state of GDScript?
I know there is the callv method (or the FuncRef.call_funcv) that kind of does what I mean, but this still leads to casting issues.

If you want to wait to complete a stack of tasks in sequence, just create a yielding method like this, and just call it. The timeline will yield until the called yielding function completes:

```
func do_craft_animation():
    $AnimationPlayer.play("crafting")
    yield($AnimationPlayer, "animation_finished")
    $Hat.hide()
    yield(get_tree().create_timer(4.0), "timeout")
    $AnimationPlayer.play("crafting")
    yield($AnimationPlayer, "animation_finished")
    $Hat.show()
```
I provided an example video that shows the new feature in question (Sorry for the odd video aspect ratio. OBS kind of dislikes me today):

https://streamable.com/s79dsk

Here is an example of the timeline:

image
image

I tried to follow your conventions as much as possible, and tried to not alter any existing stuff. However, I needed to change a small piece of the dialog_node.gd in order to make yield-waiting work. This change should not have any sideeffects.

Please let me know what you think and if I need to change something.

### DIFF
--- a/addons/dialogic/Editor/Pieces/CallNode.gd
+++ b/addons/dialogic/Editor/Pieces/CallNode.gd
@@ -25,9 +25,9 @@ func load_data(data):
 		if (event_data['call_node']['arguments'][i] == null):
 			event_data['call_node']['arguments'][i] = ''
 	
-	$PanelContainer/VBoxContainer/Target/LineEdit.text = event_data['call_node']['target_node_path']
-	$PanelContainer/VBoxContainer/Method/LineEdit.text = event_data['call_node']['method_name']
-	$PanelContainer/VBoxContainer/Method/ArgumentsSpinBox.value = event_data['call_node']['arguments'].size()
+	$PanelContainer/VBoxContainer/Properties/TargetNodeEdit.text = event_data['call_node']['target_node_path']
+	$PanelContainer/VBoxContainer/Properties/CallMethodEdit.text = event_data['call_node']['method_name']
+	$PanelContainer/VBoxContainer/Properties/ArgumentsSpinBox.value = event_data['call_node']['arguments'].size()
 	
 	_create_argument_controls()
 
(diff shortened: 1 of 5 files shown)


coppolaemilio:
Love the potential for integration with this node 👍🏻 
The only edit I would make is:
- For the `Call Node` event, make the fields `Target Node`, `Call method`, and `Arguments` in the same line. New arguments can be added below that node as it is right now

ellogwen:
Did the changes as you requested

image

PR #132, opened 2021-03-31
Improve event dragging and selection

arnaudvergnet:
Handle all drag events in TimelineEditor.
Drag offset takes element sizes into account: no more weird behavior with large items.


coppolaemilio:
While switching timelines I was able to get this error on every step while dragging an event: 
image

Can't reproduce again, and can't really know the origin 🤷‍♂️

arnaudvergnet:
It is fixed, I messed up the out of bounds guard, it went 1 too far.

Issue #46, opened 2021-02-01
Better keyboard support.


coppolaemilio:
Writing text requires to use the cursor way too many times, so I'll be adding a few hotkeys for better keyboard navigation.

## Some ideas for the Text Event are: 

- [x] `Shift` + `Enter`: Create a new text event with the same settings as this one? (or maybe alternating between this and the previous, so you can create a conversation between two characters easily)
- [x] `Tab`: Jump to the next Text Event text edit field
- [x] `Shift` + `Tab`: Jump to the previous Text Event text field
- [x] `Alt` + `up` / `Alt` + `down`: Move selected event up or down

ellogwen:
Just working on this issue, Alt up and Alt down is working already. Tab to jump to the next field is "blocked" because tab is a default character, i would propose to use Control + Up, Control + Down in unisence with the move shortcuts

ellogwen:
Implemented Movement and Selection in PR#137 as a draft

PR #137, opened 2021-04-01
shortcuts for timeline navigation as of issue #46

ellogwen:
This PR implements two parts of the issue #46

selecting previous and next event in the timeline with `CTRL + UP` and `CTRL + DOWN`
moving currently selected event up and down the timeline `ALT + UP` and `ALT + DOWN`
This PR aswell automatically focuses the textbox when you select a TextEvent
This PR aswell automatically selects the event when you click into a TextEvent textbox

Preview Video:
https://streamable.com/i6nvah
(yes, I know there are typos in the video :P)

Edit: 
Indentation is now working on shortcut movement
`CTRL DELETE` removes the currently selected event node and selects the next/last event node
`CTRL T` creates a new text event node below the currently selected and focus it's textbox to continue writing

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,11 @@ Create dialogs, characters and scenes to display conversations in your Godot gam
 v1.1 - WIP
 - Improved event dragging and selection 
 - Timeline:
-  - New event `Call Node`: Call a Godot NodePath and a method name. In addition you can add arguments as well. The Timeline will execute those methods and wait for completion, if the method in question is async and/or yielding. By [ellogwen](https://github.com/ellogwen)
-
+  - New event `Call Node`: Call a Godot NodePath and a method name. In addition you can add arguments as well. The Timeline will execute those methods and wait for completion, if the method in question is async and/or yielding.
+  - You now can drag and drop events into the timeline! 
+- Dialog node
+  - Fixed a bug where Glossary Definitions plain text was being added to the name label.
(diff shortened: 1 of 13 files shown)


ellogwen:
Indendation is missing when you move the block with shortcuts. I will fix this real quick

arnaudvergnet:
Would it be possible to remove a selected item with the delete key? Or would it interfere with the rest of the editor?

ellogwen:
> Would it be possible to remove a selected item with the delete key? Or would it interfere with the rest of the editor?

The delete key is a tricky one. A lot of interactions withing the editor uses the delete key and, inside a Text Event the delete key is part of a normal text edit workflow. Since this commit automatically selects the event panel in the timeline when you enter a Text Event text field, I dont know if it would be a good choice to chose the delete key.

ellogwen:
Do not merge: Something is odd with the character selection and portrait selection of text events. Don't know if its my fault or the problem comes from somewhere else

ellogwen:
Added `CTRL T` for creating a new text node below the current selected node and focusing it's textbox

ellogwen:
There is a wonky behaviour with the alt down sometimes switching from the timeline down to the characters. not that big of a deal tho and i think the issue is within a default control behaviour of another element, not touched by this PR

coppolaemilio:
Thank you so much!

Issue #144, opened 2021-04-02
[Enhancement] Add edit timeline button next to the TimelineDropdown control in the Editor


ellogwen:
**Is your feature request related to a problem? Please describe.**
No problem, but convenience enhancement. When you place a dialog node into a scene you can select the timeline at the property panel. It would be nice if there would be an edit button that straights jumps into Dialogic timeline edit with the corresponding timeline selected

**Describe alternatives you've considered**
The current way is: read and remember the name of the timeline you have selected, switch to dialogic tab, go to timeline, search for the correct one and click it.

ellogwen:
@coppolaemilio merged, can be closed :)

PR #145, opened 2021-04-02
edit timeline button in timeline inspector plugin

ellogwen:
This PR for #144 implements a button next to the timeline selection inspector plugin
to directly jump to the timeline editor and selecting the corresponding timeline.

This addon as well adjusts some visual/sizing issues with the original timeline selection MenuButton

Please check that this plugin does not interfere and works as you like before mergin.

Showcase: 
https://streamable.com/yggbm7


coppolaemilio:
Another great contribution! 
Just one thing:
- Make the button disabled when no timeline is selected, and if a timeline is selected make it enabled. That way it will bring the attention to the user and they will understand better what the button does.

PR #146, opened 2021-04-02
Replace the animation tween with a text timer

Crystalwarrior:
Will allow us to dynamically speed up/slow down/pause/etc. the timer in the future


coppolaemilio:
Works great! Only issue I find is that if you set the text speed to be 0.
The text speed set to 0 should make the text to show up instantly on the screen, but it is not working properly right now. I think that because this is not a tween you will have to make a special conditional when starting the timer and if the speed is 0 just show all and set it to finished instantly. Once that is fixed we can merge :)

PR #161, opened 2021-04-05
[Proposal] next indicator offset theme settings

ellogwen:
This PR adds offset_x and offset_y settings to the theme indicator to offset it and give the user more fine grain control how the indicator is positioned. You can even position it outside of the box to fit a wider range of indicator/box designs

### DIFF
--- a/addons/dialogic/Dialog.tscn
+++ b/addons/dialogic/Dialog.tscn
@@ -87,9 +87,9 @@ anchor_left = 0.5
 anchor_top = 1.0
 anchor_right = 0.5
 anchor_bottom = 1.0
-margin_left = -455.0
+margin_left = -250.0
 margin_top = -207.0
-margin_right = 455.0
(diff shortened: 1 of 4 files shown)


ellogwen:
@coppolaemilio pointed out that it may conflicts with changing themes multiple times during timeline execution. I will check this more thorough tomorrow

coppolaemilio:
The offset needs to be applied in a different way as we discussed. You can't use "margins" because those are being used by the "up and down" animation. A solution might be a wrapper and reset the wrapper position, but not sure what the best is.

Another improvement that you can make is using only 1 value of a `Vector2()` (check `button offset` or `theme_shadow_offset_x`) and use only one signal for saving the values when edited.

Last thing: I wouldn't lock this to be an `int` since people that are working on very small resolution games might also want to add floats? so leave it as reading from and writing to a `Vector2` variable. 

Thanks!

PR #177, opened 2021-04-07
Updating the theme-editor design

Jowan-Spooner:
- Creating multiple columns for all the tabs
- Creating a section title for each column
- general sorting and reordering of the settings/options
- Renaming the "Auto Color" option to "Use character Color"
- Renaming Text tab to Dialog Text
- Cleaning and reordering the beginnging of the ThemeEditor.gd script (The reference part)
- Fixing the refrences to all the elements in the ThemeEditor.gd script
- Making the tabs always go to the bottom
- moving the "text input" field and the "preview changes" button below the preview and switch them

Preview:

### DIFF
--- a/addons/dialogic/Editor/ThemeEditor/ThemeEditor.gd
+++ b/addons/dialogic/Editor/ThemeEditor/ThemeEditor.gd
@@ -18,62 +18,62 @@ var loading = true
 # a way to set an id and then access that node via id...
 # Here you have paths in all its glory. Praise the paths (っ´ω`c)♡
 onready var n = {
-	# Text
-	'theme_text_shadow': $VBoxContainer/TabContainer/Text/Column/GridContainer/HBoxContainer2/CheckBoxShadow,
-	'theme_text_shadow_color': $VBoxContainer/TabContainer/Text/Column/GridContainer/HBoxContainer2/ColorPickerButtonShadow,
-	'theme_text_color': $VBoxContainer/TabContainer/Text/Column/GridContainer/ColorPickerButton,
-	'theme_font': $VBoxContainer/TabContainer/Text/Column/GridContainer/FontButton,
-	'theme_shadow_offset_x': $VBoxContainer/TabContainer/Text/Column/GridContainer/HBoxContainer/ShadowOffsetX,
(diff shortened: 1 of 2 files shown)


Jowan-Spooner:
Btw. I renamed the auto-color option only in the editor. Inside the script and the theme-config it is still called "auto-color". I just didn't want to touch this because I was sure I would break something. In the future, it would make sense to rename that too.

coppolaemilio:
My feedback so far is:
- The section titles should take the color from the editor Theme. You can use this plugin to explore the values and pick one that fits the role: https://github.com/pycbouh/godot-editor-theme-explorer This is important to support custom/white themes.
- Make the column separators not go all the way down if the rest of the column is empty:
image

> Btw. I renamed the auto-color option only in the editor.

That's fine! I like the new description.


Once those things are fixed it should be good to go!

Jowan-Spooner:
Hi @coppolaemilio. 
1. The current design already works quite well with custom and light themes. it is a transparent grey. I will still look into the plugin you send. 

Here are some images of diffrent themes:
dialogicA
dialogicB
dialogicC
dialogicD

2. I'm not quite sure how to stop the vertical separators from going that far down. I would have to add a new Control element, which would brake all the References. This is because the size of the tabs it controlled by the tab-container and the separator has to be set to fill. I could fix this but it would require some changing of the nodestructure. 

One could consider making the tabs not expand to the bottom but I liked it more this way, because it doesn't resize all the time when changing tab.

coppolaemilio:
> The current design already works quite well with custom and light themes. it is a transparent grey. I will still look into the plugin you send.

Even if it looks good on your examples, it needs to be as close to the editor theme as possible since we can't really know if that transparent color will always work. We can make sure that the color will be consistent if we use the same properties that the editor uses to set those colors.

Here is the snippet with the change for that one. I selected the color using the plugin I tould you about:

```gdscript
var title_style = $"VBoxContainer/TabContainer/Dialog Text/Column/SectionTitle".get('custom_styles/normal')
title_style.set('bg_color', get_color("prop_category", "Editor"))
```

Jowan-Spooner:
> Even if it looks good on your examples, it needs to be as close to the editor theme as possible since we can't really know if that transparent color will always work. We can make sure that the color will be consistent if we use the same properties that the editor uses to set those colors.

I totally understand this, I just wasn't sure where to put the color change, because I don't know what function is triggered when someone changes their theme.

PR #180, opened 2021-04-08
Image folder organization

Jowan-Spooner:
The new folder structure looks like this:
> Images
>> Example Assets
>>> backgrounds
>>> portraits
>>> next-indicator

>> Plugin
>>> Context Menus
>>> Event Icons
>>>> Main Icons

>>> Pieces
>>> Resources
>>> Toolbar
>>> Tutorials

grafik

There are still some images in the Images folder but I think they aren't used at all.  I left them there in case there is a plan I don't know of. Otherwise they can be deleted.

I also duplicated the X - image (for the set-value event), because the Event was using the one from the resources folder. There was also an Icon called set-value that was somehow not used. It is renamed to set-value old now.

On that note, I renamed all the event icons to have the exact name of their event.

I also corrected all places inside scripts that I could find. The rest should have corrected itself, just caused a lot of changes in a lot of the files.


Jowan-Spooner:
Now that I think about it, it could make more sense to move the "Example Assets" folder up in the hirarchy and instead also add the "Fonts" folder as a child of it. One of the things I wanted to achieve, was to make clear what files are used for the plugin and which for the content. What do you think?

coppolaemilio:
> Now that I think about it, it could make more sense to move the "Example Assets" folder up in the hirarchy and instead also add the "Fonts" folder as a child of it. One of the things I wanted to achieve, was to make clear what files are used for the plugin and which for the content. What do you think?

Yeah, go for it! Tell me when it's ready and I can merge. Since this changes many files I'm a bit worried to keep working on other stuff because of possible merge issues.

Jowan-Spooner:
I will create a new correct one.

PR #183, opened 2021-04-08
Image organization and creation of the Example Assets folder

Jowan-Spooner:
This is the better version of the #180 PR. 

This is the new folder structure:

> adddons/dialogic
>> Example Assets
>>> Fonts
>>> backgrounds
>>> portraits
>>> next-indicator

>> Images
>>> Context Menus
>>> Event Icons
>>>> Main Icons

>>> Pieces
>>> Plugin
>>> Resources
>>> Toolbar
>>> Tutorials

>> Editor, Nodes, ...

grafik

Also I deleted images that weren't used. 

I renamed all the event icons to have the exact name of their event.

I also corrected all places inside scripts that I could find. The rest should have corrected itself, just caused a lot of changes in a lot of the files.


coppolaemilio:
I'm getting waaaay too many errors. I know they are mostly because of this update moving files and Godot not really knowing about it, but I'm worried about all the people that will notify me about this. Not sure how to proceed with this...

Jowan-Spooner:
It looks like the path replacer doesn't look for the portraits? Is this on purpose?

coppolaemilio:
Yes! I don't think anyone will use those portraits in their games since characters are the most personal stuff and I don't think anyone would want to use those other than for testing. 
I also didn't want to do the path redirect for portraits since it was going to add additional code for avoiding just an error to a small group of people. I added a warning and will mention it on the video.

The default theme on the other hand will be on every instance of dialogic, so that one was a must 😅

Jowan-Spooner:
Okay just wanted to make sure. This makes a lot of sense. Happy you got it to work!

PR #186, opened 2021-04-09
move start methods in dialog_node

arnaudvergnet:
This PR moves the `start` and `start_from_save` methods from `DialogicClass` to `dialogic_node`.

This allows the DIalogicNode to be persistent: no need to delete and create a new node to play a new timeline.

I had to add a new var named `runtime_id` to check if the timeline changed between 2 events. This was necessary when calling start from a signal fired by Dialogic: the new timeline would get loaded, then the index incremented when reaching the end of the signal event, from the previous timeline, skipping the first event from the new timeline.

To make this work I had to refactor a bit the event loading system.

This may need more work, as the DIalogicNode was not designed to be reset.

Example usage:

With DialogNode added to the tree:
Screenshot_20210409_191505

The triggering timeline:

Screenshot_20210409_191555

After receiving the start_minigame signal, the dialog closes itself, the minigame starts, and on gameover, the next timeline starts. All this using the same Dialogic node.

### DIFF
--- a/addons/dialogic/Dialog.tscn
+++ b/addons/dialogic/Dialog.tscn
@@ -8,8 +8,6 @@
 [ext_resource path="res://addons/dialogic/Nodes/dialog_node.gd" type="Script" id=6]
 [ext_resource path="res://addons/dialogic/Nodes/BackgroundMusic.tscn" type="PackedScene" id=7]
 
-
-
 [sub_resource type="StyleBoxFlat" id=1]
 bg_color = Color( 1, 1, 1, 0 )
 expand_margin_left = 10.0
@@ -89,9 +87,9 @@ anchor_left = 0.5
 anchor_top = 1.0
 anchor_right = 0.5
 anchor_bottom = 1.0
-margin_left = -250.0
+margin_left = -455.0
 margin_top = -207.0
-margin_right = 250.0
+margin_right = 455.0
 margin_bottom = -40.0
 __meta__ = {
 "_edit_use_anchors_": false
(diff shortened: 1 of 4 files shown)


coppolaemilio:
I'm a bit weary because right now the test project I have fails to run because the function `start` doesn't exists. Is there a way you could program in a fallback for it to not break compatibility? otherwise we would have to wait until v2.0 to merge this

arnaudvergnet:
I added back the previous functions (`start` and `start_from_save`) and marked them as deprecated. So this PR should not be breaking anymore.

I also added a new function add_node() which adds the dialogic node to the given parent, and checks if the parent is of type Control. If it isn't, it throws an error.

When closed, the dialog now hides, and shows itself back when calling the start method. This allows to add the node at the start of the game, hide it, and only show it again when triggering a dialog.

We will have to update the documentation to tell people to use this function now.

arnaudvergnet:
As I exposed start methods in dialog_node.gd, I named the class to improve typing when using the node and I separated public from private functions. So I renamed all private functions to add the `_` prefix, which is why this PR has a lot of changes.

arnaudvergnet:
I just noticed it broke the theme editor. This is fixed now, but seeing how the preview dialog is handled, I think it need a bit of refactoring. The dialog_node should not be marked as a tool as it is used in the final game. We should create a new script (eg: dialog_node_preview) extending this base class and adding preview related stuff to it, and only this class should be marked as a tool. But this is the work of another PR.

coppolaemilio:
I'm still getting errors on the theme editor. I can't open it but I'll see if I can manage to clean up the preview thing a bit on this same PR

coppolaemilio:
Well, it seems like it was an issue with the project I had running and not with the changes you made 😅

Issue #167, opened 2021-04-06
In game: Select/Focus the first choice button


Jowan-Spooner:
**Is your feature request related to a problem? Please describe.**
While it is totally possible to advance text events only using the keyboard right now, at each question you get stuck and have to use the mouse to select a choice. 

**Describe the solution you'd like**
I think it should be possible to navigate the choice buttons via UP and DOWN arrow and then accept the focused one with the action key (could be a prolem if you have mouse button as action key) or ENTER. 

The easiest way would be to just focus the first choice button, right?

**Describe alternatives you've considered**
I don't know any.

**Additional context**
For me it's just a convenience issue, but I think it would be good for accessibility in general.

timkrief:
I got it working but it was also pressing the button from the last input. Will try to open a PR tomorrow!

PR #187, opened 2021-04-09
First choice grabs focus, keyboard/controller can be used to answer

timkrief:
Peek 2021-04-09 16-25
(recommended to be used with native buttons from https://github.com/coppolaemilio/dialogic/pull/185 )
Solves #167


timkrief:
How do we link issues? I solves https://github.com/coppolaemilio/dialogic/issues/167

PR #193, opened 2021-04-11
Fix the portrait focus order

AnidemDex:
TL;DR: this fix a no reported issue related to the order of the portraits

Apparently there's a bug that causes lost the focus order:
image

This PR fixes it:
image


coppolaemilio:
Gracias :)

PR #107, opened 2021-03-24
Rearrange and refactor dialog_node.gd to be split up into multiple scripts

Crystalwarrior:
Having a single big script file that accounts for a million things is going to be a huge bottleneck extremely quickly, like a snowball effect.
To avoid it developing into a 3000+ line nightmare, it must be split up into multiple single-responsibility scripts early on.
Additional benefit is it's much easier to interface with dialogic and implement new things without majorly altering pre-existing behavior, as well as easier development and resolving of merge-conflicts if any even arise.

Closes https://github.com/coppolaemilio/dialogic/issues/45

PR #197, opened 2021-04-11
move text bubble in its own scene

### LINKED ISSUE #45: Unhardcode everything and split behaviors into digestible scene structure
Currently, Dialogic interface is hard-coded and difficult to expand with extra systems any developer would like to have (for example, my attempts to implement an Ace Attorney cross-examination system are hell right now.)
image
*example of current hard-coded nature*

My proposal is to expand and unhardcode the system in such a way that it's split into several digestible scenes, all of which can be replaced and overriden by developers while Dialogic still provides out-of-the-box default versions of the behaviors. This way, beginners will have an immediately working interface - what Dialogic is very good at right now - and also provide an alternative for advanced users who want to use the backend of Dialogic while handling the frontend themselves.
image
*example of a dynamic and nice version*

In simple terms, the parser node is the big focus of this issue report. A very good principle to hold for this is as such:
image
*example how a parser node only acts as an intermediary for anything connected to it to interpret the data*

This approach will also solidify Dialogic's position in the Godot plugin scene and might even make it the #1 go-to plugin for dialogue interfaces 🤔

arnaudvergnet:
This PR is based on [Crystalwarrior](https://github.com/Crystalwarrior)'s work in #107. It has been updated to work on current main.

> Having a single big script file that accounts for a million things is going to be a huge bottleneck extremely quickly, like a snowball effect.
To avoid it developing into a 3000+ line nightmare, it must be split up into multiple single-responsibility scripts early on.
Additional benefit is it's much easier to interface with dialogic and implement new things without majorly altering pre-existing behavior, as well as easier development and resolving of merge-conflicts if any even arise.

This frees about 100 lines from dialog_node.gd. This should not change the behavior of any node, it only regrouped text bubble logic in the same place.

@coppolaemilio tell me when you want to merge this and I'll remove the draft. In the meantime I'll make sure this branch is free of merge conflicts until you are ready for 1.2.

Closes #45


coppolaemilio:
tested and works good on my end 👍🏻 I want to merge it now so I can keep refactoring and simplifying code

PR #198, opened 2021-04-11
Fixed dialog_node.gd and Portrait.gd

Lecrea:
Fixed Portrait.gd- $Tween.connect("tween_all_completed", self, "queue_free" error
Fixed dialog_node.gd- Error calling method from signal 'timeout': 'Control(dialog_node.gd)::_on_close_dialog_timeout': Method expected 2 arguments, but called with 0..
  <C++ Source>  core/object.cpp:1260 @ emit_signal()
  
They caused the game to freeze after a dialogue finished.


Lecrea:
Dont mind the number of commits, this is my first time making a PR

addons/dialogic/Nodes/Portrait.gd
```diff
@@ -47,6 +47,7 @@ func _ready():
 	if debug:
 		print('Character data loaded: ', character_data)
 		print(rect_position, $TextureRect.rect_size)
+		$Tween.connect("tween_all_completed", self, "queue_free")
```
arnaudvergnet, line 50: This should be moved outside of the `if debug:`, otherwise the signal wont be received without enabling debug

coppolaemilio:
I'm closing this pr because the change on what it was based was reverted.

PR #203, opened 2021-04-12
[Proposal] 2.0 Version, flavored with classes

### LINKED ISSUE #10: Supporting more and improved data serialization options
From what I can tell, you used to use your character resources for data, but you've since migrated to using JSON for everything.

However, there are distinct pros and cons to each data format.

- Resources:
    - visual editing with drag/drop support
    - type-safe use of properties
    - support for methods, signals, constants
    - automatic data-sharing at global scale (separately loaded resources with the same path share the same instance; this is not true of Dictionaries which will produce duplicate when loading the same file multiple times).

- JSON:
    - doesn't have any of the 3.2.x frustrations with exporting user-defined resources, namely because it's all done via code anyway.
    - can easily edit data in any text editor with a commonly understandable format that does not break as a result of Godot API changes.
    - easier to export to or import from alternative editor software.

It'd be better if multiple serialization options could be supported simultaneously, presenting users with options to suit their preference/scenario.

I can also see that you manually parse and put together the data in the `editor_view.gd` script which performs many other tasks as well. Might I suggest the following changes?

- Use the Strategy pattern to define serialization strategies (or just use an enum inside some Saver class). Allow JSON and Resource as options. Make the selection configurable from the UI.
- Switch to exclusively using Resources for in-editor work, rather than Dictionaries.
- Switch to using `inst2dict` and `dict2inst` to automatically convert resources to and from JSON formats.
- Write a separate class that exclusively handles the data for the app, e.g. a DialogicDB class. Have it handle your "iterating over directories" work, building the object instances, and storing them in Arrays/Dictionaries/whatever for use elsewhere. Then, have your `editor_view.gd` just create an instance of the DialogicDB class and initialize it. This will help clean up the `editor_view.gd` script.

Also, long-term, because editing arrays of resources in the Inspector kinda sucks, we could consider implementing a spreadsheet-like editor in the bottom panel for editing multiples of these resources at once (with each row being a different instance), at which point, the entire collection of resources could be saved into a single `dialog_db.tres` file. Give people the option of saving each resource out to a separate file in a directory, but also give the option to just have it all saved to a single resource file that stores an array of the other resources internally. There's actually already a proposal I wrote about it at godotengine/godot-proposals#13. Not a requirement of the above changes though.

### LINKED ISSUE #45: Unhardcode everything and split behaviors into digestible scene structure

AnidemDex:
Ok, take a deep and slow breath. 

---
# Dialogic 2.0 Proposal
This is my __open__ proposal to be the Dialogic 2.0 version. It includes:
- A new way of handling events [DONE]
- A new editor interface [IN PROGRESS]
- A new way of storing timelines [DONE], characters [DONE], themes, portraits [DONE], definitions, configurations [DONE] and events [DONE]
- A new `DialogicNode` class

## New events
The new event class ([`DialogicEventResource`](https://github.com/AnidemDex/dialogic/blob/resources_reformat/addons/dialogic/Resources/EventResource.gd)) let you create custom events without modifying the base [`DialogNode`](https://github.com/AnidemDex/dialogic/blob/resources_reformat/addons/dialogic/Nodes/DialogNode.gd) class. It let you create, modify, or do whatever you can do in script, in game.

Every event is stored inside a [`timeline`](https://github.com/AnidemDex/dialogic/blob/resources_reformat/addons/dialogic/Resources/TimelineResource.gd), the timeline takes care of excecuting them in order. `DialogNode` just need to take a timeline and [`start`](https://github.com/AnidemDex/dialogic/blob/resources_reformat/addons/dialogic/Nodes/DialogNode.gd#L42) it.

Every event can return a [`DialogEditorEventNode`](https://github.com/AnidemDex/dialogic/blob/resources_reformat/addons/dialogic/Nodes/editor_event_nodes/EditorEventNode.gd) to be displayed inside the `DialogicEditor`, and can also register a custom button to be displayed inside the `CustomEvents` tab in the `Timeline Editor`

The new editor interface loads automatically every timeline, character, definition or theme, and takes care about displaying its correspondent editor, without hardcode the editor path.

### **Every editor takes care about itself, and nothing more.**

Every editor shows the selected resource. Modify the resource and save it. The main editor don't need to know it.

## New way of storing Dialogic files
Now, every Dialogic related file is saved as a [`Resource`](https://docs.godotengine.org/en/stable/getting_started/step_by_step/resources.html).
Resources can be exported as JSON files without problems if you need it, and be converted from JSON to resources.

Resources let you modify it with the editor, or load/edit them in runtime, without using a custom parser, those are files optimized to work with Godot.

Resources can be expanded without breaking old values.

## New DialogicNode class
DialogicNode is a new kind of `Control` node, expanded from the actual `DialogicNode`.

Its composed with 2 subclasses (nodes) and can be expanded with more:
- **Dialog node**:
It'll display text, you can modify it if you want a different Dialog Node.
- **PortraitsNode**:
It'll display the portraits. Every position is a `Control` node, so you can change the positions if you want without hardcode them.

This can potentially fix #10 , fix #45 and most of [bugs](https://github.com/coppolaemilio/dialogic/issues?q=is%3Aopen+is%3Aissue+label%3Abug)


AnidemDex:
I will be editing the information as development progresses. Support is welcome.
I propose this because I want to see if it is useful.

PR #205, opened 2021-04-12
Scene event rename 2

Jowan-Spooner:
Now it should be correct. I had some trouble doing the rebasing of my branch. Don't mind the useless commits. 

I added two spaces in front of the text of the Change Background Event Button in the timeline editor.


coppolaemilio:
Massive changes, thanks! 🤣

PR #207, opened 2021-04-12
Improve Character and portrait picker abstraction

### LINKED ISSUE #195: Join Event sometimes remembers portraits of previous characters
## The problem

**Describe the bug**
When you change the character of a join event it will remember the set portrait of the old character. Because often the other character doesn't have that portrait, the default Dwarf shows up. 

**To Reproduce**
1. Create an empty Dialogic project
2. Add two characters, one with two portraits and one with only one.
3. Create a new timeline. Add a join event.
4. Select the character with two portraits. Select the second portrait.
5. Now change the character to the one wit only one portrait.
# The portait selector will disappear -> it is impossible to select the default portrait
6. Run the game. Admire the dwarf.
 
**Expected behavior**
When selecting a different character, I would expect the default portrait to be selected.

 - OS: Linux
 - Godot Version: 3.2.3
 - Dialogic Version: 1.0

## Solutions

**Workaround**
Delete the old event and create a new one.
What you tried to get the feature working

## Furthermore
I will have to investigate if the same happens to text events.

arnaudvergnet:
This PR moves character and portrait pickers into the same node to improve abstraction.

The newly created node can be easily used into any component to allow selecting a character and its portrait, with the help of a signal.

Closes #195


coppolaemilio:
Merci :)

PR #208, opened 2021-04-12
Mirror portraits setting for characters and the Join event (updated)

### LINKED ISSUE #168: In game: Add option to mirror portraits on the right positions
If you have a sprite that is slightly looking to the right, then it looks very strange to position them on one of the right positions. 
dialogic2

It would be cool to have an option to mirror the image when it is on one of the right positions.

Make seperate portraits and use them accordingly. 

**Additional context**
Not important, but I think it would make a good addition.

Jowan-Spooner:
This is a simple implementation adding a setting for each character. If enabled the portraits will be mirrored when they are shown on the right or the center-right positions. 

There was a discussion about how to implement this, so if you don't like it, don't merge it.

It should fix #168

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,14 @@ v1.1 - WIP
   - You can now select a character to ask a question in the `Question Event` [[Tim Krief](https://github.com/timkrief)]
   - Added very basic syntax highlighting to the `Text Event` editor.
   - Fixed an indenting bug when removing events 
+  - The `Character Join` event no has a mirror option [[Jowan-Spooner](https://github.com/Jowan-Spooner)]
   - Shortcuts added! [[ellogwen](https://github.com/ellogwen)]
     - Selecting previous and next event in the timeline with `CTRL + UP` and `CTRL + DOWN`
     - Moving currently selected event up and down the timeline `ALT + UP` and `ALT + DOWN`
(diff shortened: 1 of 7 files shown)


coppolaemilio:
So, my feedback would be this:

1. Add the mirror option to the `Character Join` event:
image
_Super high fidelity mockup_ (Use `get_icon("MirrorX", "EditorIcons")` for the mirror icon)

2. Remove the "on right positions" and move it to where the offset is. This might be because the asset added is not looking to the proper side by default so leaving the option to mirror always will be handy. 
image

So the portrait would check if to flip based on the character data and then again if to flip based on the `Character Join` event. That way regardless of how your sprites are set up, you have full control of where they are looking at. If you want all your sprites to be looking left but your artists sent you some character's sprites looking the other way you can flip them. And then when you add them on your timeline you can choose to flip them again at any time with the `Character Join` event.

Jowan-Spooner:
Okay I updated this and tested it on a little project made in 1.0. 

Here is the setting on the join event:
Join Event

Here is the setting in the character editor:
Character Editor

I also updated the readme.

coppolaemilio:
amazing changes! thanks a lot 🙇🏻

PR #209, opened 2021-04-12
Add conditions to choices

### LINKED ISSUE #165: Choice event: Adding condition-logic to Choices
In many cases I would want to create questions where I can only choose some ansers under certain conditions. If the condition is false, the answer should be hidden or greyed out (depends on the game creators needs, so this could be a setting or similar).

The cleares to see in the timeline editor would be to have a condition behinde the choice much like the IF-event.
IDEA:
dialogic

Create if statements and multiple versions of each question inside them.
Very annoying and space and time consuming.

**Thoughts**
I could think of cases where I would want multiple conditions for one choice, especially for inbetweens (show choice if VARIABLE is less then X and greater then Y). But then you would have to deal with possibly infinitly long events and ANDs/ORs between them.
I think it would make much more things possible, but be a great deal of work.

arnaudvergnet:
This PR adds conditions to choices just like in if branches.

The condition is hidden by default and can be activated with a checkbox.

This is still a draft as this hasn't been thoroughly tested, but it seemed to work on my minimal testing.

Closes #165

Screenshot_20210412_230547

### DIFF
--- a/addons/dialogic/Editor/Events/Choice.gd
+++ b/addons/dialogic/Editor/Events/Choice.gd
@@ -31,10 +31,10 @@ func load_data(data):
 	$PanelContainer/VBoxContainer/Header/CustomLineEdit.text = event_data['choice']
 	if event_data.has('condition') and event_data.has('definition') and event_data.has('value'):
 		condition_checkbox.pressed = not event_data['condition'].empty() and not event_data['definition'].empty() and not event_data['value'].empty()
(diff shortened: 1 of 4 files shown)


coppolaemilio:
Nevermind, I think it is an issue with my local project.

coppolaemilio:
I'm seeing a double "New definition" option in the picker
image

PR #210, opened 2021-04-12
Add contextual menus to add timelines, characters, definitions and themes

timkrief:
Peek 2021-04-12 23-39


coppolaemilio:
I'm getting this error:  `res://addons/dialogic/Editor/MasterTree/MasterTree.gd:318 - Attempt to call function 'get_metadata' in base 'null instance' on a null instance.`

timkrief:
> I'm getting this error: `res://addons/dialogic/Editor/MasterTree/MasterTree.gd:318 - Attempt to call function 'get_metadata' in base 'null instance' on a null instance.`

When are you getting this error?

coppolaemilio:
This happens when I right click on one of the folders. I'm using the latest 3.3 rc.

coppolaemilio:
I think the issue happens because the item never gets selected, so it cannot retrieve the metadata.

timkrief:
I know what's happening, I think they change when the signals are triggered so now I end up deselecting before the rmb is triggered, and at this point, yes, nothing is selected so there's a bug. I'm on it

timkrief:
Should work now and fixes the editable settings (and future item that should not be edited) at the same time

Issue #211, opened 2021-04-13
Add volume and bus control to audio events.


coppolaemilio:

arnaudvergnet:
Linked to #151

Issue #151, opened 2021-04-04
Be able to specify audio buses for audio events and background music


Sslaxx:
**Is your feature request related to a problem? Please describe.**
I use separate audio buses for sound effects, music etc. Dialogic plays all of these through the default Master bus.

**Describe the solution you'd like**
Being able to specify which bus(es) Dialogic uses for audio events and background music.

**Describe alternatives you've considered**
Could edit the relevant scene files myself, or use signals to get functions to play sound effects and/or music.

PR #215, opened 2021-04-13
Audio Event and Background Music event update

Jowan-Spooner:
grafik


Jowan-Spooner:
Should fix #211 and #151

coppolaemilio:
I found some bugs if you really want to mess with them. But for what it's worth I'll approve it as is and then if someone really wants to break them they can create the bug reports. 

The only thing I would like changed is that they are open by default when you create them (like the text events) right now it seems weird that when you create an event you have to open it as well to start editing

Jowan-Spooner:
@coppolaemilio You are right that is annoying. I did this because I found it annoying to have them be unfolded on default (when you open a timeline), because they just make it longer. I wanted to find out wether I can just open it when it's newly created, but didn't finish it. What do you think?

Do you like to be more specific about the bugs? Do you mean the problem if you delete the audio bus?

coppolaemilio:
Ideally it should be like this anyway:
image
so even if it is closed you get the most important part of the editor, and if you want the extra options you can open it.

> Do you like to be more specific about the bugs? Do you mean the problem if you delete the audio bus?

It wasn't about the audio buses (didn't test it too much) It was regarding the order of the events. Like, if you create an event with background music 2s fadeout and the next event is another song with 0s fade, you will hear some weird things happening and the first event will "prevail". But again, I don't know why would anyone do that 🤷🏻‍♂️

Jowan-Spooner:
Okay I updated it so that they are open by default.

PR #217, opened 2021-04-14
fix: add a static proxy for C# projects

### LINKED ISSUE #55: Make Dialogic compatible with C# and add the examples on the documentation.
I never used C# yet but it seems like a good time to try it out.

mscharley:
Fixes #55 

I haven't tested this extensively yet as I'm just getting started with Dialogic but this works for me to:

* Set and retrieve variable values.
* Show a dialog.

This could use some cleanup like duplicating across the documentation in C#'s format but this seems to be functional at least. It operates in the same way that the static `Dialogic` class works for GDScript.

Example usage (no `using` statements necessary):

```
  public override void _Ready()
  {
    var dialog = DialogicSharp.Start("Greeting", false);
    AddChild(dialog);
  }
```

### DIFF
--- a/addons/dialogic/Other/DialogicSharp.cs
+++ b/addons/dialogic/Other/DialogicSharp.cs
@@ -1,0 +1,77 @@
+using Godot;
+using GC = Godot.Collections;
+using System;
+
+public static class DialogicSharp
+{
+  private static Script _dialogic = GD.Load<Script>("res://addons/dialogic/Other/DialogicClass.gd");
+  private const String DEFAULT_DIALOG_RESOURCE = "res://addons/dialogic/Dialog.tscn";
+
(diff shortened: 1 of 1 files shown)


coppolaemilio:
I'm not able to test it either but I'll accept it and see if other C# users can confirm if it is working or not. Thanks for the initial work 🙏

Keyvan-Rostyslav:
@mscharley It looks like conflict issues in the Start and StartFromSave methods due to type casting (T) to Godot.Variant

MrNavaStar:
Yes this is broken in the current version of godot

jjolmo:
Do we have a c# wrapper? shall we create it? not sure if this is on radar or if we're on our own to start a new one

PR #221, opened 2021-04-15
Add event templates

arnaudvergnet:
This PR adds the support for an event template, to be used to quickly create new timeline events.

These new events behave exactly like current ones, but are easier to create and to maintain.

Currently only the `Change Background` event has been ported, and is fully compatible with other events.
Porting new events without breaking the timeline is simple: move the header/body logic inside separate scenes, replace the content of  the event to be ported by the template, then fill in the required fields in the inspector.

The event's root node should be a container with the actual event's script, which can then communicate with the template and its body/header using `get_body()` and `get_header()`.

Each root event's script must have the following line to allow the timeline to access the template: 
```gdscript
onready var event_template = $EventTemplate
```

Here is an example with `Change Background`:

Screenshot_20210415_092552

Once the event is done, we need to edit `TimelineEditor.gd` to let it know the event now uses the template. Go to the function `_has_template` and add checks to detect the event with its `event_data`, like it was done in `load_timeline`.
Screenshot_20210415_092715


coppolaemilio:
When creating a new background node I get this error:
image
I assume you saved the node while open and it retained some of the properties set by your tests?

coppolaemilio:
It failed when selecting a scene as a background
No preview or text updated on the node:
image
And when running the game:
image

arnaudvergnet:
> When creating a new background node I get this error:
> image
> I assume you saved the node while open and it retained some of the properties set by your tests?

yes exactly! I'll remove those

> It failed when selecting a scene as a background
> No preview or text updated on the node:
> image
> And when running the game:

I forgot we could select scenes as backgrounds my bad

arnaudvergnet:
How would do you want to handle scenes? When running the main version, selecting a scene simply clears the image

arnaudvergnet:
I simply added the name of the timeline to the picker, but hid the preview as I do not know how to get it from a scene. Seems to work on my end.

PR #225, opened 2021-04-16
Use inheritance for template

arnaudvergnet:
I finally understood how inheritance works in Godot, so I improved the template system.

Events do not need a root node anymore, they only need to instance the template scene as an inherited scene (by using the link button next to the +)
Screenshot_20210416_094250

The template scene does not have a default script attached. Instead, you need to create a new one. The attached event script needs to extend the `EventTemplate.gd` class by adding the following at the top:

```gdscript
tool
extends "res://addons/dialogic/Editor/Events/Templates/EventTemplate.gd"
```

Now events can directly access template methods without needing to store an `event_template` variable. This also simplifies interactions in the Timeline Editor, as it does not need to access this variable and can do everything from the event itself.

Editing event variables with the inspector is still possible, but only after attaching the script inheriting from the class `EventTemplate.gd`. Just remember to save and press `Scene->Reload Saved Scene` if you want to see your changes in the editor.

Events can also detect when they are getting selected/deslected, by overriding the function `on_timeline_selected(selected: bool)`. This function is called by the TImeline Editor and does nothing by default, but could be useful for text events to focus the input on selection.


coppolaemilio:
I'm testing this and the event works properly (opening images, scenes, saving loading) but It doesn't show up in game. 🤔 
No error on the terminal, nothing to really know whats wrong at the moment. Any idea of what it could be?

coppolaemilio:
maybe it is related with the other bug I had that didn't display the options in the proper place. I'll investigate

coppolaemilio:
No... it seems like the background node is never created in the DialogNode

arnaudvergnet:
just saw that, this is strange, scenes work but not images

arnaudvergnet:
Seems it is broken on main as well

coppolaemilio:
Probably yes. I'll fix it asap 👍🏻

coppolaemilio:
fixed it on main

PR #228, opened 2021-04-17
Manual save data export/import

### LINKED ISSUE #224: Add a way to manage saves manually.

My game has multiple save slots which the player can save or load from. This doesn't seem to be supported by the current autosave mechanisms in Dialogic.

The ideal situation I can think of consists of three parts:

1. An opt-out scheme for the autosaves, perhaps `Dialogic.set_autosave(false)`. Nothing should be written into `user://` until a conversation dialog is shown at the earliest to give some time to the developer to call this function.
2. A `Dialogic.export() -> data` function that returns a single data structure representing the current Dialogic state.
3. A `Dialogic.import(data) -> void` function that takes a data structure from a previous export and updates Dialogic's state to those settings.

Using an opt-in scheme for autosaves could make the Dialogic code simpler at the cost of being a breaking change to developers using this addon.

**Additional context**

The current setup doesn't seem to be tenable for any style of game having more than one save state.

mscharley:
Fixes #224.

This is my proposal for manual save management. This allows developers to manually handle save data management if they need something more complicated that the autosave provides. This also allows for multiple save states using whatever save mechanism the developer is using already.

Example usage:

```gdscript
# Disable the normal autosave mechanism.
Dialogic.set_autosave(false)
# To save your game, call this and serialise the results however you like.
var dialogic_data = Dialogic.export()
# To load a game, call this with the unserialised value from #export() above.
Dialogic.import(dialogic_data);
```

## Changes to existing behaviour

Dialogic should continue to work the same way unless `set_autosave(false)` is called with one exception:

1. Dialogic will no long pre-emptively create files and folders in `user://dialogic`, instead waiting to create them just in time as it saves state for the first time.

### DIFF
--- a/addons/dialogic/Other/DialogicClass.gd
+++ b/addons/dialogic/Other/DialogicClass.gd
@@ -77,7 +77,22 @@ static func get_definitions() -> Dictionary:
 ## 
 ## @returns						Error status, OK if all went well
 static func save_definitions():
-	return DialogicSingleton.save_definitions()
+	# Always try to save as much as possible.
+	var err1 = DialogicSingleton.save_definitions()
+	var err2 = DialogicSingleton.save_state()
+
+	# Try to combine the two error states in a way that makes sense.
+	return err1 if err1 != OK else err2
+
+
+## Sets whether to use Dialogic's built-in autosave functionality.
+static func set_autosave(save: bool) -> void:
+	DialogicSingleton.set_autosave(save);
+
+
+## Gets whether to use Dialogic's built-in autosave functionality.
+static func get_autosave() -> bool:
+	return DialogicSingleton.get_autosave();
 
 
 ## Resets data to default values. This is the same as calling start with reset_saves to true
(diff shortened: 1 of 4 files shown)


mscharley:
Converting this back to a draft for now - it seems a little buggy. I originally wrote this against the 1.0 release but the latest master has changed a few things.

coppolaemilio:
Thanks for the contribution, I will be able to test today or tomorrow, but in the mean time. Have you tested it with exported projects? From previous experiences there are some differences between testing Godot projects "running them" from the editor and running exported games. 
If you haven't please give it a go and confirm that things are working as expected. Just in case 😅

mscharley:
I've only tested the new manual workflow, I haven't tested the old workflow much. I hadn't tested with an exported project but I just did that and I didn't see any issues. `user://dialogic` was still not created, and the import/export functionality worked with my saved game system.

mscharley:
Something that might be nice to do that I only just noticed was possible would be to have a setting in `res://dialogic/settings.cfg` that controls the default value for autoload. For now though, setting it to false in an `_ready` in a singleton works well though.

coppolaemilio:
@arnaudvergnet can you test this one as well since you made the original saving?

arnaudvergnet:
@coppolaemilio I saw you merged it, do you still need me to test it?

coppolaemilio:
@arnaudvergnet yes, since you did the other part of it I believe you would be better fit to test it properly. I did some tests and it seems to run fine but if you could take an extra look I would really appreciate it 🙇🏻

arnaudvergnet:
I tried it a bit and this is a breaking change. Nothing gets saved on my own project, I'll investigate a bit more to see what changed

arnaudvergnet:
Never mind this is working as expected, I forgot I edited my project so it wasn't working with the last version either

PR #233, opened 2021-04-18
add timeline start and end signals

arnaudvergnet:
This PR moves start and end timeline signals into their own signals, to separate them from events. This will make it easier to use.

### DIFF
--- a/addons/dialogic/Nodes/dialog_node.gd
+++ b/addons/dialogic/Nodes/dialog_node.gd
@@ -23,8 +23,13 @@ export(bool) var reset_saves = true
 ## Should we show debug information when running?
 export(bool) var debug_mode = true
 
+# Event end/start
 signal event_start(type, event)
 signal event_end(type)
+# Timeline end/start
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
Doesn't this break compatibility for people who have listend to the signals the old way?

arnaudvergnet:
Yes it does, so it might not get merged for 1.2, but it is a necessary change as I feel the current system isn't that intuitive

Jowan-Spooner:
Couldn't you just leave the old emit_signal statements in? And then it would keep working? And we could delete them for 2.0, whenever that might be...

coppolaemilio:
Yeah, I would leave them and add a comment to those like `# Delete in 2.0 `
Once we arrive at 2.0 we can remove all the lines marked with that comment.

arnaudvergnet:
> Yeah, I would leave them and add a comment to those like `# Delete in 2.0 `
> Once we arrive at 2.0 we can remove all the lines marked with that comment.

yep, will do

PR #248, opened 2021-04-21
add support for custom choice buttons

### LINKED ISSUE #247: Allow using custom choice buttons
The current options for choice buttons are too limited, and the amount of possibilities someone could request is too great. For this reason, we should let the user use its own component for buttons. 

In the theme editor, `Choice Buttons` tab, create a new field letting the user select a custom button scene, telling him the only constraint is that is has a `pressed` signal available.

I tried using native buttons but we can't customize them, same if using only a texture.

**Additional context**
I first planned on adding more customization options, but then saw all the possibilities and ditched the idea.

arnaudvergnet:
This PR lets the user use custom choice buttons. The custom scenes can be selected from the theme editor.

The selected scene must implement the `pressed` signal and have the `text` property available, like any component extending `Button`.

No checks are performed, the user is only told in the button/label tooltip in the theme editor. Selecting a scene which does not meet the requirements could crash the game.

Like with native buttons, activating custom buttons in the theme editor disables other controls.
Screenshot_20210421_113115
Screenshot_20210421_113136
Screenshot_20210421_113215

Closes #247


arnaudvergnet:
### Update

This new option is now hidden by default, user must enable `advanced theme options` to see it.
Screenshot_20210421_122015

Screenshot_20210421_121930
Screenshot_20210421_121942

PR #258, opened 2021-04-26
Removed print signal when choice buttons were created

Jowan-Spooner:
It's a small thing, but was annoyingly visible in the output console in your newest video @coppolaemilio.


coppolaemilio:
thanks!

PR #262, opened 2021-04-27
Builtin Documentation

Jowan-Spooner:
I created a system to have a built-in documentation. This is not ready to be merged, but very ready to be tested. I will try to make a video about it this afternoon, because there is quite a lot of functionality.

I explain most of the features and how to work with it in this video. It was just easier then writing it all down ;)
https://www.youtube.com/watch?v=La1HFWdg9us

How it looks in the master tree:
grafik

An example page. Pages are saved as markdown files:
grafik

Longer pages have a content menu. There is always a back to the top button:
grafik

It works okay with the light themes:
grafik

Here is the file system the documentation is made from:
grafik

## Things the markdown parser/page viewer supports right now:
- Bold text \*\*
- Lists (+-*)
- Mono \`\`
- Underlines \_\_
- Striketrhough ~~
- Headings 1-5
- images with github path, relative paths or godot paths (no sizes right now, all )
- links (are opened in the browser or emit a signal that can be used to open another docs page in godot) can use www-paths, godot paths, relative paths or section paths (with a #)
- The page viewer has a back to top button and a content menu that contains all ##heading2s 
- The page viewer has an edit mode that enables a refresh and edit button that are VERY usefull when editing the documentation.

## There are some more things that come with this documentation:
- A helper singleton that is used to create a documentation tree (or add it to an existing tree) from the documentations content folder.
   - The helper looks through the addons/dialogic/Documentation/Content folder and creates a tree from that.
   - Folders with only Images in them are ignored and not shown in the documentation tree.
 
## Integration into dialogic
There is not a lot of changes to dialogic:
- The DocsHelper ist added in the dialogic.gd script, when the plugin is enabled
- There is a new DocumenationViewer (that is like the other Editors, but editor was a bad word for it)
    - The DocumentationViewer mostly only consists of the PageViewer and a bit of code.
- And of course there are changes to the MasterTree Script, adding a new docs root node. All the children of that item are created by the DocsHelper
- 

## Notes:
This documentation was designed to work both in godot AND on github. The best way to work with it is to use only relative paths for images and links!
This documentation was also designed to be as little integrated into dialogic as possible for it to be easily reused for other purposes and plugins.


coppolaemilio:
I made some changes on the icons, I'm a bit worried about how long it takes to load each page of the documentation. We might need to improve how the data is stored. Maybe building it once from md to godot's bbcode and then save the bbcode version of it instead? that way we won't need to generate them every time on runtime because I have a really powerful rig and it still takes a solid second to load.

coppolaemilio:
I think that most of the issues are coming from the autoloads. I think it would be ideal to not use any autoload because of how many issues they generate. I know this might be a bit harder to refactor but I would love it if you could find a solution without using those 😬

Jowan-Spooner:
I will close this, as I added a newer version without an autoload  #278

PR #271, opened 2021-04-29
pretty print resources json with sorted keys

### LINKED ISSUE #269: Make resources more "git friendly"
Resources such as definitions and timelines are saved minimized, thus on the same line.
This makes working with git harder as each change will trigger a merge conflict.

Pretty print JSON in the files. This will increase file size but maybe we can add a minify script that users can run before packing the game.

Right now I make sure only one person edits a timeline at the same time, but this is not practical.

arnaudvergnet:
This PR makes timelines and definitions easier to read and to handle with git. It also sorts object keys to prevent unnecessary changes when simply opening the timeline without modifying it in the editor.

The drawback is the file size, which increases because of the added characters. This PR does not provide any minifying script to run before packing the game.

Closes #269

### DIFF
--- a/addons/dialogic/Other/DialogicResources.gd
+++ b/addons/dialogic/Other/DialogicResources.gd
@@ -219,7 +219,7 @@ static func set_json(path: String, data: Dictionary):
 	var file = File.new()
 	var err = file.open(path, File.WRITE)
 	if err == OK:
-		file.store_line(to_json(data))
+		file.store_line(JSON.print(data, '\t', true))
 		file.close()
 	return err
 


Sslaxx:
How much of a difference does it make file size?

coppolaemilio:
negligible, we are talking less than a small sprite

Sslaxx:
Pretty much what I thought, was wondering why it was listed as a concern in all honesty.

coppolaemilio:
I believe Arnaud comes from webdev so in a website every byte counts :)

PR #278, opened 2021-05-02
Builtin Documentation

Jowan-Spooner:
Second try, now without a singleton.

I created a system to have a built-in documentation. 

How it looks in the master tree:
grafik

An example page. Pages are saved as markdown files:
grafik

Longer pages have a content menu. There is always a back to the top button:
grafik

It works okay with the light themes:
grafik

Here is the file system the documentation is made from:
grafik

## Things the markdown parser/page viewer supports right now:
- Bold text \*\*
- Lists (+-*)
- Mono \`\`
- Underlines \_\_
- Striketrhough ~~
- Headings 1-5
- images with github path, relative paths or godot paths (no sizes right now, all )
- links (are opened in the browser or emit a signal that can be used to open another docs page in godot) can use www-paths, godot paths, relative paths or section paths (with a #)
- The page viewer has a back to top button and a content menu that contains all ##heading2s 
- The page viewer has an edit mode that enables a refresh and edit button that are VERY usefull when editing the documentation.

## There are some more things that come with this documentation:
- A DocsTreeHelper that is used to create a documentation tree (or add it to an existing tree) from the documentations content folder.
   - The helper looks through the addons/dialogic/Documentation/Content folder and creates a tree from that.
   - Folders with only Images in them are ignored and not shown in the documentation tree.
 
## Integration into dialogic
There is not a lot of changes to dialogic:
- The DocsTreeHelper ist added as a child of the MasterTree
- There is a new DocumenationViewer (that is like the other Editors, but editor was a bad word for it)
    - The DocumentationViewer mostly only consists of the PageViewer and a bit of code.
- And of course there are changes to the MasterTree Script, adding a new docs root node. All the children of that item are created by the DocsTreeHelper

## Notes:
This documentation was designed to work both in godot AND on github. The best way to work with it is to use only relative paths for images and links!
This documentation was also designed to be as little integrated into dialogic as possible for it to be easily reused for other purposes and plugins.


coppolaemilio:
I would love to revisit this one!

coppolaemilio:
Fixed the merge issues and it should up to date with main, still haven't tested if it actually works or not, but at least I resolved all the stuff

coppolaemilio:
Well, it seems that I'm able to see the tree with the documentation, but the editor doesn't work. I'll leave this one here for now but at least it is closer to working again than before! :)

coppolaemilio:
When opening an existing project with this branch I see the editor added below:
image

Error messages:
image

coppolaemilio:
It would be nice to remove the white outline to the documentation reader and make it integrate better:
image

Make the max-width of the text not be 100% so when your screen is too wide the lines don't go on forever on the screen. The regular documentation already does this and adds a left margin:
image

I would also try to style it the same way as the regular Godot documentation is. That means link colors, title colors, font sizes, etc... We need to get the styles from the actual theme so that if the user modified things they would also reflect on Dialogic's documentation. You can use this plugin to get any reference to the internal theme variables there are: https://github.com/pycbouh/godot-editor-theme-explorer

And lastly I would modify the root item to be called "Help" instead of "Documentation".

After all those changes I think it is safe for us to start writing the actual content. It doesn't have to be much, but at least the basics.

coppolaemilio:
the font is too small on retina displays (Currently using a macbook air)
image

coppolaemilio:
I'm merging so I force myself to write the docs hahaha

PR #284, opened 2021-05-04
Simple Translation Setting

bojjenclon:
This change simply adds a new setting to always treat text as a translation key, instead of displaying it directly. When on, text is sent through `tr()`  before any additional checks are performed on it.

This behavior isn't far off from how it would appear to code raw labels directly in Godot if using translation files, so I think this is an acceptable behavior for Dialogic as well.


AnidemDex:
Have you tried using it inside the editor? Or with special characters?

bojjenclon:
It doesn't translate in the editor, that wasn't my goal. It only translates in game, which is what tr() does anyway so I felt this behavior should be normal for most Godot users.

coppolaemilio:
I'll review it and probably add it with an "experimental" tag on the settings because I want the translations to be integrated also in the plugin. But this is a super clever and simple solution in the mean time 👏🏻

AnidemDex:
> It doesn't translate in the editor, that wasn't my goal. It only translates in game, which is what tr() does anyway so I felt this behavior should be normal for most Godot users.

Just take care. `tr` doesn't accept special characters since translation keys doesn't accept them

bojjenclon:
> > It doesn't translate in the editor, that wasn't my goal. It only translates in game, which is what tr() does anyway so I felt this behavior should be normal for most Godot users.
> 
> Just take care. `tr` doesn't accept special characters since translation keys doesn't accept them

My thought process was basically: someone is either going to localize their entire project, or not use localization at all. So if they have this setting on, they're already using translation keys and thus should be able to handle the `tr` conversion. Otherwise, they just leave it off and `tr` will never be hit.

coppolaemilio:
There seems to be a problem with creating the settings scene: 
image
this happens after opening the godot editor

coppolaemilio:
It seems like the settings panel was not properly saved before doing the commit?

coppolaemilio:
I merged main and did a few changes and I see it working now. I'll continue testing it.

coppolaemilio:
Love the update! Merging now.

bojjenclon:
Sorry about that, I think it was due to how I created this branch. I did all the work in my game's repo then copied files over to the dialogic repo folder, and I probably forgot to update some stuff in the process.

Issue #275, opened 2021-04-29
Wait seconds event unfolded


arnaudvergnet:
## The problem

**Describe the bug**
When click on the unfold button on the top left, the wait seconds event is also unfolded while it has no body.

**To Reproduce**
1. Open a timeline
2. Add a wait seconds event
3. Click on unfold all events on the top left
4. See the unfolded event

**Expected behavior**
It should not be possible to unfold events without a body

**Screenshots**
![Screenshot_20210429_194439](https://user-images.githubusercontent.com/80701113/116595513-b1fc8d00-a923-11eb-9e37-8e2e12c71855.png)

**System (please complete the following information):**
 - OS: Ubuntu 20.04
 - Godot Version: 3.2.3
 - Dialogic Version: latest main

## Solutions

**Possible fixes**

Block unfolding when event has no body

Jowan-Spooner:
This was fixed by the Timeline eventblocks rework.

Issue #237, opened 2021-04-20
Question text going out of screen in editor


arnaudvergnet:
## The problem

**Describe the bug**
As the question text is one line only, it can be too long and go out of screen, making an horizontal scroll appear.

**To Reproduce**
1. Open a timeline
2. Create a question
3. Write a long line
4. See the scroll appear

**Expected behavior**
The line should wrap if too long, like in a text edit.

**System (please complete the following information):**
 - OS: Ubuntu 20.04
 - Godot Version: 3.2.3
 - Dialogic Version: latest main 7970e027f9226de14aa29cc9eaa95e2c9dc03a77

## Solutions

**Possible fixes**

We could use the new template to reuse the text input content, so the behavior would be the same, but limiting to one line only.

Jowan-Spooner:
This is basically fixed on the main branch. The Question event now is more similar to the text event.

PR #287, opened 2021-05-06
Timeline eventblocks rework

Jowan-Spooner:
What should come with this pull request are:
- Mutli selection of events (SHIFT for range, CRTL for adding/removing from the selection)
- Refined Copy-Pasting
- Change of Shortcuts to
    - DEL (without CTRL)
    - UP/DOWN for selecting the next/previous event (withou CTRL)
    - Added CRTL+V for pasting
    - Added CRTL+C for coping
    - Added CRTL+X for cutting
    - Added CRTL+D (although it's not working correctly yet)
    - Added CRTL+A to select all (needs some work, because it overrides CRTL+A in text fields, which is annoying..
    - Added CRTL+SHIFT+A to deselect all

Now to make this possible I started recreating all the Events from the Event templates. This also meant changing the way the EventTemplate works. 

All events are now inherited scenes from the EventTemplate scene. Also they don't have individual scripts anymore but only one big EventBlock.gd script. They are made unique through their exported variables (event_data, style, name, header_scene and body_scene, etc.). This should make creating new events easier, as well as changing the behaviour of all EventBlocks at once by minimizing the amount of duplicated code.

To make this possible I had to recreate all the now so called EventParts (the contents of the Events) and also create a big amount of new ones. They all extend a class EventPart, so they all work very similar now. If you want to create a new EventPart then start with a VBoxContainer and add a script to it. Then copy the ExampleEventPartScript's content into it. Build your EventPart from there!

Note: Because I was a bit scared about redoing everything, I deleted pretty much nothing (I renamed the old stuff). This means one of the next commits before merging needs to be a lot of cleanup of now unneeded files...


Jowan-Spooner:
Should fix #275 and this #237 (by making question more similar to text event).

Jowan-Spooner:
Oh and this also includes some fixes for the event_id's ;)

Issue #191, opened 2021-04-10
Add Set Glossary event


arnaudvergnet:
**Is your feature request related to a problem? Please describe.**
We should be able to edit glossary data during runtime, like we do with variables.

**Describe the solution you'd like**
We could add a "Set Glossary" event, to change a glossary definition's data. We could also add a new property to enable/disable the glossary entry. This would help when the entry must be unlocked to be seen.

**Describe alternatives you've considered**
My current approach is to change the name of the entry so that it doesn't appear in the text, then change it back when it is unlocked. But the drawback is that you have to emit a signal and do it by code.

PR #290, opened 2021-05-07
Add SetGlossary event

Jowan-Spooner:
Adds a new event that allows changing a glossary variable during the game. The only unfinished thing is that there is no event-icon!


Jowan-Spooner:
This should fix #191

PR #332, opened 2021-06-01
Inline anchors refs don't work

clemens-tolboom:
Ie this https://dialogic.coppolaemilio.com/#credits is not working due to the dash `-`

I had to remove the ❤️ from credits to make it work.

I also made inline links work.


coppolaemilio:
thanks! I will hopefully be able to finish the new site soon 😅

PR #340, opened 2021-06-04
ConditionPicker rework as EventPart

### LINKED ISSUE #335: Implement the ChoicePicker as a better EventPart
## The problem

**Describe the problem**
Right now the ChoicePicker works differently then the other EventParts. This was to quickly fix things but for the future it would be better to reimplement it the "proper way". 

One reason more is, that right now the DefinitionPicker isn't used so it doesn't used the folders. Now that's bad.

## Solution
I would want to have this as soon as possible. Maybe in the Version 1.2?

Jowan-Spooner:
I have tried. Please test if everything works as you want it to. 

If it works this fixes #335


coppolaemilio:
The `If condition event` should not display the "has condition?" part as it would always have one:

image

coppolaemilio:
The conditional also compares strings, but I'm getting this warning "The selected operator requires a number!"
image

Jowan-Spooner:
yeah I didn't test the condition event enough. it works for the choice. I'm working on it ;/

coppolaemilio:
Same with the choice `[Equal to]` warning:
image

PR #344, opened 2021-06-07
Added Audio bus selection for Typing SFX and looping sounds effects support

timkrief:


timkrief:
I also fixed this issue I stumbled upon in this PR 
image

PR #345, opened 2021-06-08
Modify loading function to permit direct loading from GDscript

eladar2000:
When loading the timeline directly with GDscript as descriped in the FAQ :

```
func _ready():
	var gdscript_dialog = Dialogic.start('')
	gdscript_dialog.dialog_script = {
		"events":[
			{ 'event_id':'dialogic_001', "character": "", "portrait":"", "text": "This dialog was created using GDScript!"}
		]
	}
	add_child(gdscript_dialog)
```
the loading of the timeline is not complete because most of the loading is done in the set_current_dialog function, and that function is only called if the "timeline" attribute is set.

I tried in this PR to create a separate loading method that is called by set_current_dialog and by itself if the timeline is created by GDscript (with empty timeline attribute).

### DIFF
--- a/addons/dialogic/Nodes/dialog_node.gd
+++ b/addons/dialogic/Nodes/dialog_node.gd
@@ -48,7 +48,7 @@ func _ready():
 	
 	# Checking if the dialog should read the code from a external file
 	if not timeline.empty():
-		dialog_script = set_current_dialog(timeline)
+		set_current_dialog(timeline)
 	elif dialog_script.keys().size() == 0:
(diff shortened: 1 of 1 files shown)


coppolaemilio:
Thanks for the PR, I'll review it and try to merge it today 👍🏻

coppolaemilio:
Sorry for the delay, got many things going on and I wasn't able to test this one properly. I will improve this in 1.3 so I expect to revisit it after the release of 1.2

PR #353, opened 2021-06-10
Randomness feature for the Set Value event

Jowan-Spooner:
There is now a dice symbol that (when toggled) will reveal to boxes for a minimum and a maximum number to choose. 

Worked well in my test but we will have to see.


coppolaemilio:
There is an error (not sure where) but if you set the random value `to be` random from `3` to `3` the result is 5. I think some numbers are getting rounded up and end up being bigger than the possible value?

PR #389, opened 2021-07-06
Remove unused files

miere43:
I don't think these files are supposed to be checked out into repository...


coppolaemilio:
thanks!~

Issue #393, opened 2021-07-09
Possible fix to the export settings requirement?


coppolaemilio:
https://docs.godotengine.org/en/stable/classes/class_editorexportplugin.html

In this class there is a method called `add_file`

Maybe it could be used and add all the .json and .cfg files, so that we don't have to make the user do that manually on the export dialog.

Still haven't tested any of it, but if you are reading this issue and feel adventurous give it a try :)

Lucas-Romao:
I managed to get it working, I'll open a pull request

Jowan-Spooner:
I will test this now and close the issue if it works.

Jowan-Spooner:
It works, but only if you have a theme (as reported in #466). It could very well be a different thing that produces the error though.

Great work @LuRomao !!

Jowan-Spooner:
I would say this is solved. Thanks again @LuRomao

PR #396, opened 2021-07-10
Fixing requirement to manually configure export dialog

LuRomao:
Should fix proposal #393


coppolaemilio:
Hey! this looks really nice, thank you for making it! One thing I don't understand: Is it adding every file? We only need it to add `.cfg` and `.json` files, but I don't see any filter to pick only those in the script 🤔

LuRomao:
I think I forgot it, i'll add a filter

LuRomao:
@coppolaemilio Here you go!

coppolaemilio:
Thank you! I'll review it and merge it tonight 👌

coppolaemilio:
It seems like some things are exporting properly, but the definitions don't seem to be working as intended. I will keep testing and see if there is any solution to that

PR #398, opened 2021-07-11
Update template bug_report.md

theludovyc:
In system
Remove "please complete the following information"
Update dialogic default version to 1.2.4


coppolaemilio:
Hey! Thanks for the commit, but I won't be merging this one.
The "please complete the following information" is important for people to read.
And it is not a default version number, it is just an example for them to fill out since we don't really know what version they are running.

If they end up just sending the defaults to us we can tell, otherwise it will seem like they specifically wrote the "1.2.4" and might lead to some bug hunting confusion.

Hope you understand 🙃

theludovyc:
I understand :upside_down_face:

PR #399, opened 2021-07-15
More flexible saving system

Jowan-Spooner:
# New functions !
## dialog_node
The dialog_node has now two new functions:

### get_current_state_info()
returns a dict with info such as
+ current timeline and event
+ current background
+ current background music
+ current portraits

### resume_state_from_info(state_info)
+ that loads in all of the above and starts the timeline 

These should NOT be used directly. 

## Dialogic Class
Instead the DIALOGIC CLASS has two new functions too (actually more):
### save_current_state(save_name)
saves the state (of the last added dialog node) and the current definitions to the save with the name save name (replaces or creates the save folder)

### resume_from_save(save_name, dialog_scene_path, debug_mode, use_canvas_instead) 
is similar to start() but get's all info from the given save folder and also loads the definitions
 
there is also a get_save_names_array() function in the DialogicClass that is usefull for creating save/load menus.

I have yet to test all the problems that come with this.


Sslaxx:
Been wondering if changing how definitions are stored (as a Dictionary) and serialised (avoiding the restrictions with JSON) would make some of this more flexible, but at what cost.

Jowan-Spooner:
This is old, so I restarted and there is a new pull request (a draft too, but largely updated).

Issue #362, opened 2021-06-18
Dialogic.start's CanvasLayer gets hidden by CanvasLayers with a higher layer number


Sslaxx:
## The problem

**Describe the bug**
`Dialogic.start` adds everything to a `CanvasLayer` on layer 1. If you have a `CanvasLayer` with a higher layer than this, then it will display over the top of the layer supplied by `Dialogic.start`.

This is because `use_canvas_instead` is `true` by default, and this means the layer is given the default value of 1. There may be reasons developers may not require the start function to provide a `CanvasLayer`.

**To Reproduce**
Create a `CanvasLayer` with a layer value of 128, create a `ColorRect` with any colour as a child of it.

Use a script along the lines of:
```
onready var dialogic_test = Dialogic.start ("sslaxx_test_talk")

func _ready () -> void:
	add_child (dialogic_test)
	return
```

**Expected behavior**
Unfortunately, due to the way `CanvasLayer` works this *is* the expected behaviour.

**Screenshots**
![362_Demonstrator](https://user-images.githubusercontent.com/324679/122575301-8440e400-d048-11eb-8bf1-236a3c7bf868.png)

**System (please complete the following information):**
 - OS: Linux
 - Godot Version: 3.3.2
 - Dialogic Version: 1.2

## Solutions

**Possible fixes**

* Have the `use_canvas_instead` argument be `false` by default.
* If `use_canvas_instead` is true, allow an optional `use_canvas_layer` type argument to specify a layer to use.
* It may also be worthwhile considering a `start_with_canvas` function, which can specify a layer value.

coppolaemilio:
@Sslaxx can you check the lasted merged change? It enables an option in the settings to fix this. If this is still not fixed let me know and I can reopen the issue 👍🏻

Sslaxx:
Will do.

Sslaxx:
![326_Demonstration](https://user-images.githubusercontent.com/324679/126365272-c0c79824-ec38-4c85-86ce-e41b466a0948.jpg)

This is a `Label` attached to a `CanvasLayer` with a layer of 32. The Dialogic layer is set to 128.

Sslaxx:
Fixed by #403.

PR #403, opened 2021-07-19
Custom Canvas Layer in settings

RedXGames:
This is a fix to [Issue]https://github.com/coppolaemilio/dialogic/issues/362)
In the settings editor, you can now choose the layer index which the dialogic's canvas layer will use.
This should prevent dialogues from being rendered behind UI with higher layers.


coppolaemilio:
thank you! I'll try to review it and merge it soon

coppolaemilio:
Thanks again!

Jowan-Spooner:
@RedXGames. I'm trying to fix #427 right now and wonder if there is a reason why the canvas layer box is a LineEdit instead of a input made for numbers. Any reason or just by accident?

Issue #401, opened 2021-07-16
Voice support.


Sslaxx:
**Is your feature request related to a problem? Please describe.**
Dialogic has no support for voice acting built in. This means there is no (easy) way to add voiced lines.

**Describe the solution you'd like**
Not sure. Maybe like Parrot does? https://github.com/deep-entertainment/parrot/blob/main/docs/resources.md Or AGS - https://adventuregamestudio.github.io/ags-manual/MusicAndSound.html#voice-speech

**Describe alternatives you've considered**
Not really sure there can be any, without hacking into Dialogic's source code.

RedXGames:
I'm working on implementing voice support for my game, I will pull a request when it's ready.

RedXGames:
@coppolaemilio  Hey, Emilio, I'm almost done with the voice support, It is working perfectly except for one tiny bug.
I couldn't find the part of the code that creates extra messages when entering new lines in text events.
I need to find it because somehow the voices don't play if the text event has multiple lines with the option to create new messages is on.
Do you know where can I find it in the code?

coppolaemilio:
@RedXGames it should start here: https://github.com/coppolaemilio/dialogic/blob/main/addons/dialogic/Nodes/dialog_node.gd#L211

RedXGames:
I must be blind, I spent 30 minutes in that script and didn't see it xD.
Thanks for the help

Jowan-Spooner:
Hey @Sslaxx. I've finished @RedXGames pr and merged it. Feel free to test it. Of course make sure to have a backup of your project!! Main is always WIP and surely contains bugs.

I'll write a little tutorial on the feature later today but it's easy to use: Just enable Voices in the Settings and (if you want) pick a default voice audio bus!

Hope it works as you wanted it.

Jowan-Spooner:
Let us know if this fixes this issue for you, so we can close it.

Jowan-Spooner:
I'll close this for now. If the feature doesn't work correctly or more is needed, we can reopen this or add new issues.

PR #406, opened 2021-07-20
added basic voice support

RedXGames:
I added basic voice support for text events and questions regarding [Issue 401](https://github.com/coppolaemilio/dialogic/issues/401)
I also fixed some bugs with 'editor_reference' being null on some events.

**How to use** 
-Enable dialogue voices from the settings (It's disabled by default because most people wont use it)
-Create a timeline and add a text event
-You can now load a sound file (.ogg or .wav) to play when the text event gets executed.

**IMPORTANT NOTE**
If you have "New lines will create extra messages Enabled" then the voice will only play for the first line, To select voices for the other lines, press the 3 dots button and load all the audio files at once(Files need to be named by order so they match the line order in text)

Final note : 
This is not the final version but this is what I was able to make in my limited time, I will try to tweek and improve it asap.

VoiceSupport

### DIFF
--- a/addons/dialogic/Dialog.tscn
+++ b/addons/dialogic/Dialog.tscn
@@ -1,7 +1,8 @@
-[gd_scene load_steps=8 format=2]
+[gd_scene load_steps=9 format=2]
 
 [ext_resource path="res://addons/dialogic/Nodes/TextBubble.tscn" type="PackedScene" id=1]
 [ext_resource path="res://addons/dialogic/Example Assets/backgrounds/GlossaryBackground.tres" type="StyleBox" id=2]
+[ext_resource path="res://addons/dialogic/Nodes/CharacterVoice.tscn" type="PackedScene" id=3]
 [ext_resource path="res://addons/dialogic/Example Assets/Fonts/GlossaryFont.tres" type="DynamicFont" id=4]
 [ext_resource path="res://addons/dialogic/Nodes/glossary_info.gd" type="Script" id=5]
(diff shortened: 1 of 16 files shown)


Jowan-Spooner:
Hey thanks for you work so far, this looks like it will be of great use.

However I would think it would be better to move the voice line control below the text editor and use the already existing AudioPicker for it as it already comes with controls for Volume, AudioBus, and a preview functionality. 

Also I'm not too happy about the way you select multiple files right now (for multiple lines), as there is no way to find out whether they are in correct order (besides testing it). But I understand that all other ways would mean a lot more work...

Please tell me if you see things differently then I do and if you would like to work on this further, as otherwise I could work on this too. But for now thank you so much for working on this and contributing!

RedXGames:
I was thinking the same thing about moving the audio picker below the header This will make it much easier to pick more than one audio file and it will be much more organized, I will definitely update that in the near future.
And man, At first I tried using the already existing AudioPicker but there was a big that prevented the editor file dialog from appearing. So I made my own node, Luckily I managed to fix the bug with the AudioPicker, It had something to do with editor_reference not getting assigned properly for PartEvents.gd on certain events.
I think now I have a clear idea on what I need to do and hopefully I can make it.
Thanks for your advice and I will definely work on this more, I'll give it a crack this evening and see where it goes.

RedXGames:
I closed this pull to improve the user experience. I will pull another one when it's ready.

Issue #402, opened 2021-07-16
Timelines are deleted if folders are nested too far


zaknafean:
## The problem

**Describe the bug**
When moving subfolders into one another, timelines are deleted if the nesting goes too far. The timelines no longer exist within the Godot file system or the Windows Explorer.

**To Reproduce**
1. Create 3 folder with a timeline in each. /A, /B, /C with timelines tA, tB, tC
2. Move folder /A into folder /B. /B/A/tA and /B/tB are fine
3. Move folder /C into folder /A. /B/A/tA and /B/tB are fine, but /B/A/C/tC no longer exists

Attached below is an animated Gif demonstrating the problem

**Expected behavior**
I would ideally expect the folder rearrangement to just happen. If thats not possible an error message. Bare minimum I'd like the timelines to not be deleted, so I can still use them rather than have to recreate them all from scratch.

**Screenshots**
![TimelineDeletionBug](https://user-images.githubusercontent.com/7741797/125954651-995b803c-e05d-485a-8ff4-22d0090986cc.gif)

**System (please complete the following information):**
 - OS: Windows
 - Godot Version: 3.3
 - Dialogic Version: 1.2.5

## Solutions

**Workaround**
You can move folders up to 2 deep without issue. So **Timeline/FirstFolder/SecondFolder** is fine but **Timeline/FirstFolder/SecondFolder/ThirdFolder** is bad. Interestingly you can successfully CREATE a timeline in that hypothetical **ThirdFolder** without much issue. Its only the moving of the subfolders that seems to break.

**Possible fixes**
Possible fixes for this are beyond my skill in coding.

zaknafean:
I think I found the issue, thankfully seemed pretty simple. 
Line 202-203 of DialogicUtil is a recursive call dealing with removing folders within folders, and is called by the the **move_folder_to_folder** function. 
Notably this **remove_folder** function takes an argument to delete files, and naturally **move_folder_to_folder** passes a _false_ here. Unfortunately the recursive call defaults to _true_, and that seems to be the source of the bug. All sub folders in the folder being moved wipe out the files recursively. I don't think this was intended.

The simple fix is to simply pass along the _delete_files_ argument to the recursive **remove_folder** call. Thus if you don't want to delete the files in the parent folder, you don't delete the files in the child folders, and if you do want them gone, they also go. 

[...]

PR #408, opened 2021-07-21
Fix for Issue #402

zaknafean:
Fixed issue where nested folders would delete contained files


coppolaemilio:
Works great! Thank you so much for finding and fixing the issue 👏🏻

PR #409, opened 2021-07-23
Dialogue Voice Support

RedXGames:
I Have updated the voice support implementation from my previous pull after suggestions from @Jowan-Spooner.
This  feature was also suggested on [Issue 401](https://github.com/coppolaemilio/dialogic/issues/401)

This will allow users to play character voices during a text/question event event.

**How to use** 
-Enable voices from settings.
-You will be able to load audio files depending on your text event

VoiceSupport


Jowan-Spooner:
I'll try it out this afternoon. Looks great already!

One last thing I thought might be cool would be a default audio_bus in the settings, so if you want to have your voice be in a special AudioBus you would not have to select it everytime.

What do you think? Gonna try the other things later!

Jowan-Spooner:
Okay so I'm trying it right now:

Functionality wise:
-  I noticed that the voice line will play untill it either ends or another voice line replaces it. I would think it should also stop if another text event without a voice line is shown. 
- Still would like the "Default Voice-Line Audio Bus" selector in the settings
- Also, THIS IS A REQUIREMENT, please move the CharacterVoice.gd script into the Nodes folder. 

UI-wise:
- What does the "[Multiline Mode]" mean? It seems to be there even when I have only one line so it's either unnecessary or not working correctly.
- If possible I think it would be cool to have a label with like "Line 1: " in front of the corresponding audio selector, as counting through could get tedious. (Could be cool to fetch the text from the text-input field so that you could hover over the label and see the content of that line, but this is pretty unnecessary)
- I would like the label in the settings to have a better text. My suggestions is "Enable voice-line selectors" but I'm not very proud of that either so feel free to suggest something better.

Otherwise this looks great. Thanks so much for your contribution. If you need any help feel free to ask here or on emilios discord server!

RedXGames:
That's weird. voices are stopping when new text events appear with not audio (at least from my end), this might be a bug. I'll look into it.

I will add the default bus and the labels this afternoon. (Did not notice the CharacterVoice being on the wrong folder!.. OOps)
[Multiline mode] is active when you have "New lines will insert new messages on", I think I'll remove it because it may confuse people.

Jowan-Spooner:
Oh I see. You can leave the MultiLine Mode in if you explain it better. I will retry maybe it was a mistake on my side that kept the sound playing! Good luck implementing!

RedXGames:
I managed to find the problem with the voice not being interrupted. I'll commit the changes in a bit.

Jowan-Spooner:
Hm I can still get the sound to keep playing:
Some more info for debugging: 
- The sound in question is a pretty long .wav file
- Her it is used (the female merchant file):
grafik
*Please ignore what sounds I choose for thesting and my terrible mouse-writing XD*

RedXGames:
It should be fixed now. let me know if you're still having problems

Jowan-Spooner:
Works perfectly!

RedXGames:
I Have converted this pull into a draft as I'm going to add more refinements to it sometime this week.

Jowan-Spooner:
Very good! I have added the Milstone 1.3 so we don't forget to add this to the new version. There is so much other stuff to do till then that I'm sure it will make it in!

Jowan-Spooner:
Hey @RedXGames have you managed to work on this? If you don't have the time right now (or just lost interest for now) I would gladly do the final work as I think it's only small stuff missing. Would love to have this on main soon.

RedXGames:
Hello @Jowan-Spooner, I really couldn't get any free time lately so I didn't work more on it.
If you wanna take a crack at it feel free to, it's mostly just small things left such as adding the default audio bus and saving settings.

Jowan-Spooner:
Cool! I'll give it a shot today and see what I can do.

Jowan-Spooner:
@RedXGames Once again huge thanks for this contribution. Feel free to test and critique the changes I made. I've merged it for now, so more people can easily try it out and find problems.

Keep up the good work!

RedXGames:
@Jowan-Spooner Thank you for finishing the job! I'm working on a tight schedule these days.
Will give it a try this afternoon
And thanks again for your help.

PR #410, opened 2021-07-23
A modular Custom events implementation

Jowan-Spooner:
This is a draft as this is early in developement. Still wanted to let people know I'm working on this.

Here is how it works:
- Select a folder in the Settings menu (my recommendation: res://dialogic/custom-events/):
grafik
- In that folder create a folder for your custom event. give it a good name. 
- To work correctly, this folder must contain at least two files:
1. A file called EventBlock.tscn that extends the template scene at `"res://addons/dialogic/Editor/Events/Templates/EventTemplate.tscn"`.
     - You don't need to fill out all of the variables of that event. BUT make sure to set a UNIQUE id. Best practice is to use your unique user name and a number (e.g. "rabloe_000"). This means when you give your event to someone else, they won't likely have the same id!
2. Then you need to have a script file (.gd) that is named EXACTLY like the event_id in the EventBlock scene (except with a .gd in the end)!!! This is how it will be called so don't mess this up.
     -  this .gd file needs to have a `handle_event(event_data, dialog_node)` function

Once you are happy with your events, go back to the settings and hit the Refresh button next to the Folder selector.
When you are back in the TimelineEditor, at the bottom you will find your custom events.

As an example, here is what I decided to do with it:
grafik

Here are the two events, so feel free to try them out:
[QuitEvent.zip](https://github.com/coppolaemilio/dialogic/files/6869586/QuitEvent.zip)
[PrintEvent.zip](https://github.com/coppolaemilio/dialogic/files/6869587/PrintEvent.zip)


zaknafean:
2 things of note so far though:
A) I think having a custom-events folder just be created in the same dialogic folder as characters, themes, etc makes alot of sense, and will make compatibility for newbies alot easier (IE: just take this folder and dump it in that folder, and you wont have to fix dependencies)
B) Can't drag and drop custom events into the editor at this time
C) A mostly blank 'example' template folder would also be helpful to have

zaknafean:
B) Was due to the new custom button not inhereting the EventButton script. I solved it like so:

image

Could be better ways, but this seems to do the trick.

Jowan-Spooner:
Thanks for testing and your suggestions.
- A custom-events folder will now be created automatically (if not existent). Do you think I should the possibility to change the folder completely?
- I implemented B. Works now.
- I would think I would write a little entry in the documentation about creating a custom event. Additionally I would provide the two events I made in a different repo (and leave a link to that in the docs). What do you think?

zaknafean:
Functionally its all great (aside from a print statement of 'Set' when you opt in). Edge case bug I found however.

1. Successfully set up a custom event. Add it to your timeline. It works yay!
2. Disable custom event opt-in
3. Restart editor or Navigate back to timeline and hit refresh custom events
4. Your custom event is erased from the timeline completely, even in the JSON. The data is lost.

I assume that if an event_id isn't recognized the auto save removes the 'bad' block to keep the json valid. This probably becomes a tricky situation as you definitely want the timeline to be self correcting, but I can also see valid situations where valuable data could be lost if it cleans up a temporarily broken or corrupted custom event, such as during an upgrade, code migration, etc.

Assuming what I think is happening, is happening, than I believe that the proper behavior is probably to just ignore the block if its the id doesn't start with 'dialogic_', rather than remove it. You'll assume its a custom event, and that it could return at some point. It could get messy, but hopefully its not a common case.

zaknafean:
It all looks good, and I haven't been able to break it!

zaknafean:
Found a bug. I'm not sure if this existed previous to the dummy build or not, as it was an accidental find. But seems we get some memory leaks.

image

We get around 21 orphan nodes that appear RIGHT as the dialog box appears. This appears to be related to having any event in the custom-events folder when you've opted in. Both my event and your print event trigger the same problem. Triggers regardless of Node use directly or singleton.

coppolaemilio:
@zaknafean was this issue fixed after the latest commit?

zaknafean:
It was fixed. I've found no further flaws with this branch.

coppolaemilio:
That's great! @Jowan-Spooner feel free to merge it to main whenever you think it's ready 👍🏻

coppolaemilio:
Fixed all the merge conflicts. Still would love you to test one last time and see if it is good to merge. Feel free to do so whenever!

zaknafean:
Sounds good. I'll pull from main and do a thorough testing come this weekend.

Out of curiosity is there a spot that would make sense to toss community made custom events? I'm sure others would like to use my camera shake or camera pan events, let alone what other people come up with. Things that don't belong in every game but could be pretty handy for some

coppolaemilio:
I want to enable a section in the dialogic website to share events, but at the moment I guess just posting it on google? I can imagine enabling a way of installing custom events from within dialogic, but that ofc is a lot of work 😅 .
We'll see!

Issue #99, opened 2021-03-21
Change Speed/Settings of a Dialog During Timeline Execution


ellogwen:
I just fiddled around a bit with your amazing plugin, but stumpled on something I couldn't find out how to implement.
I have a dialog with default speed of 2. While the timeline is running, I want to change the speed of the text to 100 but just for one sentence. Is there a way to do this? I tried to use bbcode, but couldnt find the right vars. I want to accomplish something like that

```
This is strange, [speed=100]very strange[/speed]
```
If there is no bbcode tag available, is there a way to access the current dialogs settings with using the Set Var node? But this would only work for whole chat nodes and not for single sentences. I will try it with a custom RichtTextBox effect in the meantime.

Maybe you can add something like this `[dialogic speed=100]test[/dialogic]`

Edit: Okay, you cannot slow down / access the speed inside a RichTextEffect. Darn.

ellogwen:
Ok, since you use a tween for determing the total length times the character count, changing the speed inbetween would not work aswell.

arnaudvergnet:
You could create a new theme with a high text speed, then trigger a theme change just before the chat node you want sped up, then after this node switch back to previous theme. This will allow different speeds between nodes, but yeah I don't think it is possible mid-sentence

coppolaemilio:
Like Arnaud said, the theme is the one dictating the speed not the bbcode, so it is not possible to change using any custom tag. I could explore adding more granular customization inside the text messages in the future, but for now you will have to make different themes for different messages.

Crystalwarrior:
Why not ditch bbcode and do something like this instead?
`\t{100}\e{armscrossed}`That thief `\t{40}`you ran into `\t{100}\e{normal}\s`should be your `\f`real suspect, `\t{100}\e{smug}`wouldn't you say?

\t = time, ms per letter OR percentage of global theme speed like 0.5
\e = emote, aka change character portrait mid-text
\f = screen flash
\s = screen shake

ellogwen:
@Crystalwarrior then problem ist not the syntax. furthermore, Godot has builtin functionality for bbcode and RichTextEffects per character to do shake and complex text effects. It would gain nothing to ditch this already done work in favor of a new syntax. 
The problem however, is not related to bbcode or any text rendering at all, it's how the animation is implemented within the plugin, using a Tween and calculating the length for the whole text box in advanced.
[...]

PR #422, opened 2021-07-27
Add text speed multiplier

Jowan-Spooner:
This is usefull to have a settings slider. I'm not quite sure if it's what was meant but it could be a fix for #99

It adds two functions to the DialogicClass:

### Dialogic.set_text_speed_mutilplier(float)
0 = text shown instantly
smaller 1 = slower
1 = the normal speed (this is default)
bigger 1 = quicker
## Example
ezgif com-gif-maker
*the slider goes from 0 to 2*


coppolaemilio:
Looks great! the only thing I would say at the moment is that 0 should be the slowest, and 2 should be instant. It is a bit weird to go back from super slow to instant so close in the slider and in the oposite side of what speed means

Jowan-Spooner:
Hm yeah I could do that (would do it tomorrow). I choose this way as I thought it was what renpy's speed setting did. Also a speed of 0 (would take infinite time to show up). But we can discuss this again tomorrow.

coppolaemilio:
@Jowan-Spooner This one was merged https://github.com/coppolaemilio/dialogic/pull/523, so I guess the branch would need an update.
Let me know if you want me to take over

Jowan-Spooner:
This is way outdated. I personally still like the idea though, so maybe I'll find time to redo this in the future.

PR #424, opened 2021-07-28
Rework of the character name coloring 

### LINKED ISSUE #359: The problem with the color of Names in the text
## The problem

**Describe the bug**
Depending on the order in which the names are written, they behave incorrectly. Since I am extending this via google translator, it is better to show it so that you understand.
Monika1 - Character Name
Monika2 - Display Name
Monika3 - Nickname
The player too.

The test lines look like this.

Here's what we get in the game:

The following order also behaves strangely:

As I understand it, DisplayName was not intended to change the color in the text at all.

I also wanted to note that bbcode does not work in the glossary. Perhaps this was intended, but it would be convenient if it was possible.

Version 1.2 release

Jowan-Spooner:
Hopefully fixes #359. Uses some regex now.


coppolaemilio:
Issue is still not fixed:
image

image

image

coppolaemilio:
nevermind, I tested in main 🤣

Issue #426, opened 2021-07-29
undo/redo feature for timeline editor


Schwender-exe:
some form of undo/redo within the timeline editor would be helpful, I've accidentally deleted entire nodes of dialogue accidentally by pressing backspace a few too many times and with no way of undoing it I end up having to re-write it.
obviously not an _asap_ feature, but one that would be very much appreciated!

coppolaemilio-farmersfridge:
yeah, it should be fixed asap.

coppolaemilio:
Yeah, it should be fixed asap. I'll add it to the 1.3 milestone

Jowan-Spooner:
Kind of a duplicate of #321. But I agree. Maybe I can do this today. Already implemented it halfway once.

Jowan-Spooner:
Small status update:
- Deleting, Cutting and Pasting can now be redone on my branch.

Will be working on Event Adding, moving and duplicating next.

Jowan-Spooner:
Hey @Schwender-exe (and @coppolaemilio) please check out my attempt at implementing this #428. As always when testing things, make a copy of your project or create a new one. 

Hope it works!

coppolaemilio:
Tested this a lot and tried to break it but it seems pretty solid. The only issues I found were mixing some copy and paste from text and events but it is better to see those errors than to lose an event. 

Thank you so much once again @Jowan-Spooner 🙇🏻 

Closing the issue now, but feel free to re-open or comment

Issue #321, opened 2021-05-28
UndoRedo support for deleting events


Jowan-Spooner:
**Is your feature request related to a problem? Please describe.**
Often I delete a event or even more by accident. It's really annoying that there is no way to get them back,

**Describe the solution you'd like**
Just be able to use CRTL+Z and CRTL+SHIFT+Z as everywhere else.

**Additional context**
I will look into this and how it works, but I could imagine it to be a bit harder.

Jowan-Spooner:
This was actually solved quite some time ago ;)

PR #428, opened 2021-07-29
UndoRedo support for the TimelineEditor

Jowan-Spooner:
Created to solve #426 and #321.

Implemented with a local instance of godots UndoRedo.

By the time of writing this, it should work for:
- deleting events (with DELETE key)
- pasting events (CRTL+V)
- cutting events (CRTL+X)
- moving events (by dragging them)

When it messes up:
- if you do a not undoable action between these actions, the undo might do unexpected things (move the wrong item, etc.)

Future ideas:
- would love to make adding of events via the buttons and CRTL+T undoable, same goes for duplication. Also the option buttons and movement using ALT+Arrow keys. 

Please test this!


Schwender-exe (commented):
Implemented changes and seems to work, though will have to test it a bit more to be 100% certain. Other than the CTRL D and CTRL Y command not implemented yet, works well!

Schwender-exe (changes requested):
- CTRL D and CTRL Y commands
- as you stated before, undo/redo support when adding events to timeline
- (misc, but related to original issue) grab_focus when selecting choice event textbox

Jowan-Spooner:
Thanks for testing. 

About Redo (CRTL+Y):
Actually in godot (at least for me, not sure if it's the same for you) Redo works with CRTL+SHIFT+Z. That should actually work already.

Schwender-exe:
ah, that is in actually didn't notice, guess it's just because I'm using CTRL+Y for redoing actions, thanks though!

Jowan-Spooner:
I've pushed a fix for CRTL+Y. Don't think it hurts.

Jowan-Spooner:
Okay I finally managed to implement all the features I wanted for this now. Feel free to test again. Adding in events should be undoable as well as duplicating. I'll be away the next couple of days so if there's a problem you'll need to wait a bit.

coppolaemilio:
I will test this now and see if I can merge to main

PR #433, opened 2021-07-29
Change the Selected Event Style to only have blue borders and not change event color.

### LINKED ISSUE #431: Selected event is very dark in the white theme
Needs to be updated because the text is hard to read.

Jowan-Spooner:
Supposed to fix #431.


coppolaemilio:
Nice! finally the selection style as it should be 😅

PR #442, opened 2021-08-02
Refactor insert_variable_definition

theludovyc:
Working, it's can be merged =) !


coppolaemilio:
What is the practical improvement over the previous existing code? Can you provide an example where the previous version wasn't working and this new one is?

theludovyc:
Old version and New one works. It's just performance :)

theludovyc:
Look at the code of replace https://github.com/godotengine/godot/blob/0cee8831b25377567bff8f46f7eaad8cb203dbf8/core/string/ustring.cpp#L3434

Sslaxx:
Would rather see variable entries replaced with Dictionaries to allow more datatypes to be stored in them.

PR #445, opened 2021-08-06
This tells git to ignore certain nonessential files from export allowing for faster installing from AssetLib

quentincaffeino:
Not sure if this is actually welcome, but this gives a faster and easier install flow where user doesn't need to select what to keep.

Note: this works by removing ignored files from zip downloaded from github. This also affects releases archives, so you might want to keep something like README or LICENSE. But other than that folders like .github and files like .git* are really have no reason to be exported.

Awesome project :)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+.git/           export-ignore
+.github/        export-ignore
+docs/           export-ignore
+.gitattributes  export-ignore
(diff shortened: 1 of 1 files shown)


coppolaemilio:
I belive the readme is quite important to include, same with the docs, but the other git stuff can be left out :)

Issue #454, opened 2021-08-10
Regex name compiler throws errors when not escaping special characters


zaknafean:
## The problem

**Describe the bug**
Regex compiler isn't escaping question marks (and probably other reserved characters)

**To Reproduce**
1. Make a character with a ? in their name
2. Use them in a timeline
3. Get errors like below

**Expected behavior**
This probably isn't a high priority issue, as I think things still work, but it does look bad in the debugger. I'd expect to not get errors

**Screenshots**
![image](https://user-images.githubusercontent.com/7741797/128926745-3b5a42cf-06ef-434d-aaa8-1e2202904b7c.png)

**System (please complete the following information):**
 - OS: [Windows, ]
 - Godot Version: [3.3.3]
 - Dialogic Version: [1.3 w/pr410]

## Solutions

**Workaround**
It is probably ignorable, and I haven't seen it break anything yet, but it probably should be addressed.

**Possible fixes**
Error manifests in dialog_node.gd line  231, when it is parsing characters via regex.
![image](https://user-images.githubusercontent.com/7741797/128926952-4ca2519a-de0e-42a4-983c-df21180c7a10.png)
I spat out some output of 'regex_thing' and it game me this:
(????|Blue)\b
(Frei)\b
(Chuck)\b
(???|Doc)\b
(Beryl)\b
(????|Blue)\b
(Frei)\b
(Chuck)\b
(???|Doc)\b
(Beryl)\b
(????|Blue)\b
(Frei)\b
(Chuck)\b
(???|Doc)\b
(Beryl)\b

????, Frei, Chuck, Beryl, and ???  are all names of characters i'm using. These need to translate to \?\?\?\? and \?\?\? to avoid the issue. I imagine this also needs to be the case for any other reserved character that could be a Character name.

zaknafean:
Best way to see this in action is actually to make a character with the name **...** and give it a specific color. It'll really mess up the formatting of the dialog. With some sleuthing I have a solution though, which I'll pull request for review. To better state the problem:

If you have a character name that contains any special character for regular expressions ( any of these **(+&|!(){}^\"~*.?:\-** )
then formatting and parsing can break, and many errors appear in the console. This isn't a big deal for most of those characters except the & or ?, particularly if name changing is on the horizon.

[...]

PR #455, opened 2021-08-11
Escape regular expression characters in names

zaknafean:
Fix for #454
New regular expression added to escape regular expression characters that could appear in a name.


coppolaemilio:
Thank you so much 🙏🏻

PR #481, opened 2021-09-01
Implement start and stop time to charactervoice and audiopicker

KvaGram:
This may be a bit of a hack, I'm sure there are more elegant ways of doing this.
But I think this should work just fine.
Let me know what you think.


Jowan-Spooner:
Hey @KvaGram Thanks for the PR, I will look at it now!

Issue #164, opened 2021-04-06
Timeline: Scrolling to newly created events


Jowan-Spooner:
**Is your feature request related to a problem? Please describe.**
When creating new text events after the last event in a timeline, the new event is invisible (when your timeline is to long to fit the screen). You will have to scroll down to see it. While writing Dialog, I often have both hands on the keyboard, so this is quite annoying.

**Describe the solution you'd like**
It would be cool if the scroll could reveal/go to any new event that is added.

**Describe alternatives you've considered**
Ignoring it. Works okay too.

**Additional context**
This could also be interesting in case #163 is fixed, because right now when scrolling there will always be an event selected that is visible. If that would not be, then this here could be very useful too.

PR #486, opened 2021-09-06
TimelineEditor: Scroll to newly created events

Jowan-Spooner:
This is a huge milestone for me because #164 (the original issue) was the very first thing I did on this github ;).  Glad I came so far to finaly trying to remove it again.


Jowan-Spooner:
Noticed that this also means, when you open a timeline you will land at the very end.... But that could be a good thing. What do you think @coppolaemilio ? Maybe you could try it out? See how it feels?

coppolaemilio:
hmmmm I think it could be confusing, but maybe we can add an extra argument for the scroll to not trigger when the event is created on the loading of the timeline?

Jowan-Spooner:
probably! will see to it!
maybe not today ;)

Issue #363, opened 2021-06-18
Scrolling dialog text with keys


moonwolf-github:
If the text does not fit into the box, a scroll bar appears so you can scroll the text with the mouse. Can the keys be used for this purpose?

Jowan-Spooner:
Maybe if we focus the text element. Would need to try it out. Good idea though

coppolaemilio:
I think we should make a setting to select if you want the game input to be mouse, keyboard or both. That way we can handle all this kind of interactions depending on each project. This would fix many other issues of people having issues with not wanting one or the other as input methods.

PR #488, opened 2021-09-07
TextBox: Scrolling with arrowkeys 

Jowan-Spooner:
I manged to get it to work by enabeling selection on the RichTextLabel. Then I added a empty Control node in front of the textbox so you cannot ACTUALLY select the text. Would be glad if someone could try it out. 

Should solve: #363


Jowan-Spooner:
great!

Issue #478, opened 2021-08-30
Add a sound effect every time you advance in the dialogues.


EmmanuelHuazo:
I would like to add more aural feedback to the player so it occurred to me that you could add the ability to add sound effects every time the player presses the space bar to advance the dialogues.

I could even supply you with some free effects to try

Jowan-Spooner:
This was fixed by #503!

PR #503, opened 2021-09-11
Add audio picker and audio effects: wait, pass, hover, select

### LINKED ISSUE #260: Add sound effect options in theme settings
i sort of did it manually for text advancement but i think it breaks something lol

But, basically a function that plays sound effects when advancing text, selecting a dialogue choice, etc.

timkrief:
Closes  #260 and #478


Jowan-Spooner (approved):
It seems to work great now! Only thing that could be changed as I said before is changing "waiting" to "Text completed" and "passing" to something like "Next Action", but I'm open for debate and think it's something that could be changed later.

PR #505, opened 2021-09-11
Fix the negative pitch issue, and add some validation to the audio picker fields

timkrief:


Jowan-Spooner:
I'm confused: doesn't the allowed volume range start from like -80db?

timkrief:
This was made less confusing with the last commit

PR #512, opened 2021-09-12
Fixed "Don't change" setting.

TimurBahadir:
Currently the "Don't change" setting isn't handled in the `set_portrait` method and will reset the portrait to the default one.


Jowan-Spooner:
Oh god. Pretty dumb that that didn't work. Thanks so much for finding and fixing it!

Issue #443, opened 2021-08-03
Hide the Text Event Character picker if there are no character resources in the project


coppolaemilio:
Kinda weird to see it there if there are no options to select.

zaknafean:
Pretty easy fix using the DialogicUtil class to check if a character exists. The event part is hidden fine and reappears when you add a character properly. Only real question would be should we have a message about adding a character up there rather than just having it be blank?

coppolaemilio:
it could be interesting to add, but not sure if it would add much to the plugin tbh. If you want to add a character I feel like the character folder is right there for you to add. My main issue is: imagine someone is using dialogic but doesn't want to deal with characters: why should they see the prompt of "create a character! click here!" on every text event 🤔

zaknafean:
Not going to try to figure out what this does...
![image](https://user-images.githubusercontent.com/7741797/133128705-9f029f85-ffe6-4d97-aec6-62487291eea2.png)

But I wonder if we may not need it now. Added your idea to Character Join and Character Leave
![image](https://user-images.githubusercontent.com/7741797/133129284-07138497-dce9-49ef-bc9d-0ec7d370b306.png)

coppolaemilio:
that's great, can you add maybe a button that would create a character as well?

zaknafean:
Seems reasonable. I'll see how well it flows.

coppolaemilio:
Thanks! you can also probably replace the get_node('../../.././././././....././././.') or whatever with a find_parent 🤣

zaknafean:
I think the whole .../........../.. is most easily solved by not having Character Leave and Text Event share an Event Part. They seem to have different wants, but that's out of scope for this at the moment. 

The button thing works better than I'd have thought! Only real question is if 'flat button' is better than regular button. Below 'Join' is flat, 'Leave' is an adjusted.

![image](https://user-images.githubusercontent.com/7741797/133141185-188fdb7e-500d-4a19-97a6-51ea50f73431.png)

coppolaemilio:
I would do:
You haven't created a character yet. Click `[here]` to create one.

zaknafean:
mmm ya that looks better. Will send the pull request up for a once over.

![image](https://user-images.githubusercontent.com/7741797/133143479-ab0b87ac-5efa-471e-b121-1d825aac4866.png)

PR #516, opened 2021-09-13
Hide the character/portrait when no characters exist

zaknafean:
We count characters found and hide the event part if its 0


zaknafean:
In reference to enhancement #443

Issue #473, opened 2021-08-23
More range for dialog text speed


Pilvinen:
I would like to have **more range for dialog text speed**.

Currently you can set it to 0 to stop the typing. Since only integers are accepted the next setting is 1 which is, in my opinion, way too slow. Then you have range to adjust it all the way up to 10 to make it even slower.

What I'd expect to see in 1 would be text being typed really fast, almost instantaneously, but still typed. And from there you could make it as slow as you like.

I suggest refactoring the code to accept **float values** for this setting in order to avoid breaking backwards compatibility with existing code of end users.

In other words you could set Speed to values like 0.01, 0.33, 1, 2.5, 3.0, 5 and so on.

zaknafean:
So behind the scenes '1' equates to a wait speed between letters of '.01' seconds, so the current range for waiting is .01 seconds to .1 seconds between letters.

I've adjusted the editor to allow steps of .1, so now the range is technically  '.0001' to '.01' seconds. Simple enough change but...

I wasn't seeing much different in anything sub 1, so I set up some tests with the classic 'Lorem ipsum' dummy test. The completion speeds in milliseconds were as follows:

- **Speed 0.5:**  7413, 7412, 7411
- **Speed 1:**  7410, 7411, 7412
- **Speed 1.5:**  7414, 7412, 7413
- **Speed 2:** 8930, 8912, 9813
- **Speed 2.5:** 11144, 11145, 11127
- **Speed 3:** 13379, 13380, 13362
- **Speed 3.5:** 15611, 15612, 15596
- **Speed 4:** 17811, 17829, 17845

The variance is annoying, but given the methods I used, I can't get much more accurate. 
What we're pretty much seeing is the difference per letter below speed 2 (.02) seconds per letter is negligible compared to the general overhead in displaying the letter and the other background signals. After these findings I wanted to see about where the difference started, and the answer is...

- **Speed 1.6:** 7412
- **Speed 1.7:** 7562
- **Speed 1.8:** 7995
- **Speed 1.9:** 8480

So it seems below somewhere between 1.6 and 1.7, the speed you pick won't matter, at least not with how individual letters are currently handled. That said I think there is still some value in this range upgrade, and its a pretty minimal change set, so I'll put up a pull request for review.



You can try my tests with the attached zip file.
[dialogic-473-test.zip](https://github.com/coppolaemilio/dialogic/files/7190288/dialogic-473-test.zip)

coppolaemilio:
[...]

PR #523, opened 2021-09-18
More range for dialog text speed #473

zaknafean:
Propgated **text_complete** signal for testing and future functionality
Changed theme speed to be a float, stepped by .01
Cleaned up some tabbing in canvas_dialgog_node to match the rest of dialogic


zaknafean:
This is more of an alternative to Jowans, **Add text speed multiplier**, thats simpler, though certainly not better as far as functionality goes. As his is more 'live' speed and this is more 'set theme speed', I don't think there is much intersection functionally speaking, but I imagine merge wise they wouldn't be quickly compatible.

Issue #194, opened 2021-04-11
ability to view dialog history(Backlog)


univeous:
I think it's pretty common in dialog-based games like visual novels.
![image](https://user-images.githubusercontent.com/40604180/114298402-d1548700-9ae8-11eb-8556-2dbc730c3e12.png)
(screenshot from [ATRI -My Dear Moments-](https://store.steampowered.com/app/1230140/ATRI_My_Dear_Moments), in case anyone is interested)

PR #524, opened 2021-09-18
Text History

zaknafean:
Basic Text History functionality

### DIFF
--- a/addons/dialogic/Nodes/DialogNode.gd
+++ b/addons/dialogic/Nodes/DialogNode.gd
@@ -9,6 +9,7 @@ var waiting_for_answer: bool = false
 var waiting_for_input: bool = false
 var waiting: bool = false
 var preview: bool = false
+var record_history: bool = true
 var definitions = {}
 var definition_visible: bool = false
 var while_dialog_animation: bool = false
@@ -45,6 +46,7 @@ var audio_data = {}
 onready var ChoiceButton = load("res://addons/dialogic/Nodes/ChoiceButton.tscn")
 onready var Portrait = load("res://addons/dialogic/Nodes/Portrait.tscn")
 onready var Background = load("res://addons/dialogic/Nodes/Background.tscn")
+onready var historyLabel = $HistoryPopup/MarginContainer/RichTextLabel
 var dialog_script: Dictionary = {}
 var questions #for keeping track of the questions answered
 
@@ -561,6 +563,20 @@ func update_text(text: String) -> String:
 
 func _on_text_completed():
 	play_audio('waiting')
+	
+	if record_history:
+		var characterData = get_character(current_event.character)
+		if characterData.has('name'):
+			var parsed_name = characterData['name']
+			if characterData.has('display_name'):
+				if characterData['display_name'] != '':
+					parsed_name = characterData['display_name']
+			var characterColor = characterData.data.get('color', Color.white)
+			historyLabel.append_bbcode(str("[color=",characterColor,"]",parsed_name, "[/color]: "))
+		if current_event.event_id == 'dialogic_001':
+			historyLabel.append_bbcode(str(current_event.text, '\n\n\n'))
+		elif  current_event.event_id == 'dialogic_010':
+			historyLabel.append_bbcode(str(current_event.question, '\n'))
 	
 	finished = true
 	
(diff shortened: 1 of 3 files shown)


zaknafean:
The open issue most relevant is #194

zaknafean:
While the basic history capturing works, there are quite a few design related decisions that are probably necessary. The following are things we could capture.
- [ ] All text
- [ ] Chosen choice
- [ ] All choices
- [ ] Played Sound Effects
- [ ] Played Music
- [ ] Character entrance/leaving
- [ ] Background changes

### Design considerations:
Should this be theme-able?
Should there be a setting that toggles 'history recording'?
Should definition functionality occur in the history?
Should dialog advancement be possible while history is popped up?

### Implementation considerations:
Right now the history timeline is just a RichTextLabel that is appended too. We could instead use something akin to the 'events parts' to allow people to better customize how the history is shown. A custom scene in the **res://dialogic** folder that can be extended by power users. This would almost certainly not be compatible with themes however, but if you wanted something like #194 is showing, you'd almost have to go this route. No theme would make that happen.

addons/dialogic/Nodes/DialogNode.gd
```diff
@@ -45,6 +46,7 @@ var audio_data = {}
 onready var ChoiceButton = load("res://addons/dialogic/Nodes/ChoiceButton.tscn")
 onready var Portrait = load("res://addons/dialogic/Nodes/Portrait.tscn")
 onready var Background = load("res://addons/dialogic/Nodes/Background.tscn")
+onready var historyLabel = $HistoryPopup/MarginContainer/RichTextLabel
```
coppolaemilio, line 49: We usually use PascalCase for node names, so instead of `historyLabel` it should be `HistoryLabel`. Not a big deal, but if you can update it would be great :)

coppolaemilio:
I know this is still not marked to review but when I tried to run an existing timeline I got this error:
image

zaknafean:
Looks like I'm assuming character data. Will make sure to get a check in there for the next  revision. But just to check, what kind of time line event are you using.

coppolaemilio:
No character data indeed, and this is the timeline
image

zaknafean:
Ah, its actually the wait of all things. I didn't expect it to trigger **_on_text_completed** as there is no text involved with it. But it makes sense you'd take advantage of the existing timer there. I'll put a check in there, though I'm working on a more maintainable place for the 'logging' to occur atm. This definitely still falls within proof of concept.

Not sure I'm happy with the logic in **_on_text_completed** , but right now **answer_question** but it does function how I think will be expected there (IE: It doesn't record history until the dialog is actually finished playing or a choice is actually made). Most other places would put the dialog in before it finishes playing, but if you have other thoughts on that let me know!

zaknafean:
Its at a point where feedback is gonna start to get useful, so if you all see something off, speak up!

I moved most of the logic into its own nodes, the important parts are:
HistoryPopup - this controls the timeline and what gets logged. It has 2 important functions
- add_history_row_event(current_event) - Takes event data and parses out important bits to log stuff. Using this for actual text events
- add_history_row_string(stringData, audioDataPath='') - Takes up to two strings and blindly adds the contents to the history. AudioDataPath is optional
With these functions we can add logging against events (and custom events if they call it themselves) like so:

```
if record_history:
	HistoryTimeline.add_history_row_event(current_event)
```
or
```
if record_history:
	HistoryTimeline.add_history_row_string(str(get_character_name(event['character']), ' has arrived.'))
```

If audio data is in the event or a path provided, then the 'play' button on the timeline will be activated for that event.
image

HistoryPopup has its own AudioStream for playback and so stopping playback when you close history is more trivial, but it'll take a bit of work yet to use the proper audio bus.

So quite a bit less proof of concept now.

zaknafean:
Noting suggestions from Emilio in discord here:

- [ ] I would remove that (styling of individual boxes) and only leave the background panel to style it
- [ ] off by default, new tab on the theme settings to enable (settings for history)
- [ ] and in that theme tab for the history show styles similar to other panels (background, etc...) and history button position (preview of history theme)
- [ ] there is also a double open bug: if you have the history open and you click again in the history button you will see how the history closes and opens again automatically (swap button to toggle)

zaknafean:
Choosing a theme for the 'back' panel is good! (done in settings though)
Can also have the dialog boxes retain their styles from set-theme event optionally (again done in settings)

Working on the double click bug and 'shadow' that the popup causes.

zaknafean:
I think its ready for another review as far as direction goes. Highlights so far:
image
We got settings! Enabling, picking a theme, static 'theming' of history dialog boxes versus dynamic theming, and picking where the button goes all there.

image
image

Resize of screen works fairly well, updating generally well. My current todo list is as follows

- [ ] Let HistoryButton inherit theme customization of Choice Buttons
- [ ] Add setting customization of margins of panel or inherif theme customization of DialogBox
- [ ] Make 'Return' button mimic position of 'History' button rather than be fixed

Anything else jump out at anyone?

PR #552, opened 2021-10-01
Audacity label importer

KvaGram:
This is a tool to mainly improve the voice acting support of the text events, but applies to all audio events that feature a start_time and stop_time value.
A developer may now use the label feature in audacity to mark where a line will begin and end. This data may be exported from Audacity to a text file.
This text file may now be imported by this tool.

how to use/test:

1. How to use open any event that has an audio picker.
2. Unfold the audiopicker's advanced settings, find the start and end section.
3. press the new Aud butto
3a. If no label track have been loaded, the file loader will appear.
3b. select a label track export txt file 
3c. press the Aud button again
4. select a timecode.
5. observe that the audiopicker's start and end values change to the selected item
6. press the LOAD NEW LABEL TRACK button to load a different label track (returning to step 3a)
7. Do not let yourself be caught in an infinite loop by step 6

known issues:
- The popup's tool panel is transparent. Tried to fix it with a panel, but it somehow failed.
- The popup appears at the wrong place. Possible that it needs an offset I do not know about.
- Attempting to drag the scrollbar closes the window.


KvaGram:
Alright, new update, as you can see from the commit.
The popup has a small glitch that I don't know how to fix yet, but right now, that's the only issue I can think of.
Small change in the UI. The popup's toolbar is gone.
The popup now is a popup window with a window text replacing the toolbar's text label.

The list of timecodes are now hidden, when no timecode data is loaded.
In its place, is a new button replacing the one from the toolbar, letting the developer load a file.
Left-clicking the Aud button no longer opens the file dialog, that honor befals the new button mentioned above.
Left-clicking the Aud button will now ALWAYS open the popup, regardless if any timecode data is loaded.
Right-clicking the Aud button, on the other hand, will now always open the file dialog, allowing a new file to be loaded.

This change were made with workflow in mind, I hope this works out as well as I imagine it.

KvaGram:
uh, yeah, missed unchecking one of those localization files. Bah, they are annoying.

coppolaemilio:
I will close this one since we discussed about it before, hopefully these kind of extensions will be easier for 2.0

PR #559, opened 2021-10-05
Adding DialogBox Placement Options

Jowan-Spooner:
Added a position selector for the dialog box. It allows for Top/Center/Bottom + Left/Center/Right selection. 

The "Bottom Gap" has been renamed into "Box Padding" and the  old "Box Padding" to "Text Padding".

Works fine for me.


coppolaemilio:
Looks great, but it seems like it doesn't place where it should go unless you resize the screen. You might need to call the alignment when the dialog is added as well.
resize

PR #560, opened 2021-10-06
Add definition type enum

GimmiRuski:


zaknafean (commented):
Logically no change here, but it certainly is more readable overall

coppolaemilio:
I would place this enum in the main Dialogic class instead of creating a new one

GimmiRuski:
> I would place this enum in the main Dialogic class instead of creating a new one

Done.

PR #607, opened 2021-10-31
fix code insertion in faq->Using Signals?

astrale-sharp:
hope its okay if i opened 2 PR's for the same thing :/
I'm editing directly from github that's why , sorry !

make code highlighting and indentation correct


coppolaemilio:
No worries! I think that the syntax wasn't set up properly because this is built in the editor and we don't have the documentation parser set up there. So I can't really merge this change until we patch the documentation viewer

astrale-sharp:
Okay!
with the fix it will look good directly on github too!

PR #618, opened 2021-11-04
Display multiple characters per TextBubble timer timout

dbackstr:
Added a new theme editor variable called "character_step" to represent how many characters should be displayed each time TextBubble.gd calls _on_writing_timer_timeout(), to facilitate faster apparent text display speeds.

### DIFF
--- a/addons/dialogic/Editor/ThemeEditor/ThemeEditor.gd
+++ b/addons/dialogic/Editor/ThemeEditor/ThemeEditor.gd
@@ -5,6 +5,7 @@ var editor_reference
 onready var master_tree = get_node('../MasterTreeContainer/MasterTree')
 onready var settings_editor = get_node('../SettingsEditor')
 var current_theme : String = ''
+var use_advanced_themes : bool = false
 var preview_character_selected : String = 'random'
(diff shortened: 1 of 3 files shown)


Jowan-Spooner:
Hi I like the idea of the changes but it looks to me, as if you also reintroduce a number of old theme settings, that shouldn't be there anymore. You might have to correct that.

dbackstr:
I made these changes off of the distribution version-- 1.3, the one that comes from the asset library-- So any more recent changes, might not be reflected in my version. That might be the cause of reintroducing things that aren't meant to be there. What specifically is there that isn't meant to be there?

Jowan-Spooner:
Yeah that is likely the cause. There are a couple of old theme settings that where removed, you can see it if you look at the changed files on github.

Not sure what the best to solve this is. As there is another PR changing the [speed] command right now, it might be the easiest to redo these changes from a clean branch, once the other one is merged.

dbackstr:
hmm, in that case I'll close the pull request for now. maybe if I have some time I'll get the current master and see if I can re-implement the functionality.

Issue #537, opened 2021-09-27
Allow waiting and changing text speed mid-sentence


maksutheam:
I saw this mentioned in the Dialogic 1.2 highlights devlog, but I didn't see an issue about it here, so I thought I'll write one up just in case.

**Is your feature request related to a problem? Please describe.**
The biggest reason as to why I'm not using Dialogic right now is that I can't pause or change the speed of the text mid-sentence. How long a character takes to say their line, or the delivery of the line in other words, is _really_ important in communicating the meaning of the message when all you have is a textbox. If a character has a hard time saying something, I could use slow text and a lot of pauses to communicate that. If text appears too slow or fast, it can ruin a good joke or make an emotional moment fall flat. 

Here's an example. See how the characters pause their dialogue for half of a second after a comma or a punctuation? And how "Butz" goes from feeling sad to desperate by slowing down and speeding up his text, despite never changing his portrait? And how the pauses between the words at the end makes the reveal more impactful? With the current Dialogic version, I couldn't replicate this.

https://user-images.githubusercontent.com/60304828/134929281-9a6d0220-071d-4831-9a8e-7fdfb31cd004.mp4

**Describe a solution**
For example, if I had a text event that contained 
`[speed=1]What!?[speed=20] What's wrong, Larry?`
the `What!?` would appear at a speed of 1, and the text after that would appear at a speed of 20. You know, left to right.

An alternative to this is right to left. In that case, I would have to write
`What!?[speed=1] What's wrong, Larry?[speed=20]`
But I personally think left-to-right is more intuitive.

**Describe alternatives you've considered**
I could just separate each pause and text speed change into a different line. But then the player would have to manually skip a lot, and it would limit how much text I could show on screen.

Jowan-Spooner:
I agree. This has been asked for many times. Same goes for little pauses mid-sentence.

Just so you know (not as a solution, just a usefull fact), you can use the [nw=1] tag to automatically make the dialog continue after 1 second.

Jowan-Spooner:
Okay this is implemented in the current WIP version (main branch on github). There is now a [pause=X] command to pause at that point for X seconds. Also the [speed=X] command doesn't affect the full event anymore, only what comes after it (so speed can change as often as you want during your text event.

I'll hopefully implement automatic pauses for special symbols soon.

PR #621, opened 2021-11-06
Trackable text commands

KvaGram:
Rewrote TextBubble to track where commands are inserted into text, and triggering commands when `text_label.visible_characters` reach desired indices within the text for the commands.

Currently added support for playing system sounds, emitting the dialogic signal, and changing typingspeed on the fly.
The regex code that used to strip the text of the speed and no wait commands have been unified into just one, joined with the new commands.

Warning: if the previous nw=v feature I worked on is not desired at this time, you may need to cherry pick. It made it into this branch somehow.


Jowan-Spooner:
From what I can tell the implemented functionality so far works great. 

### More arguments?
It could be cool if we would also implement a syntax for giving more then one argument, maybe similar to how the BBCode commands do it (with one word specifying the command and more arguments given with argname="value"). This could make implementation of other events into the text event easier.

### Play command
As already said on the discord, I find the use cases for the [play] command in it's current state quite lacking, and would rather have it play a custom sound I specify with a path, but then again, this would be very easy to do with the signal command which is one of the best additions to dialogic in a long time imo ;).

Jowan-Spooner:
Additionally I would suggest to implement a [pause=0.4] command. Then I would also suggest that we add settings to the settings menu that automatically add this command (at runtime) behind dots, commas and similar symbols. Would be a simple solution to #537

KvaGram:
I want to be the one to do it. So I can make time to work on it today.
So let's see if I got this right, correct me if I'm wrong.
1: remove the play option.
 * its intended use to be replaced with the signal option, pending user implementation
2: Implement support for additional arguments
* The current code makes use of regex to seperate keywords/arguments. I am not confident in finding a way to do this in an elegant way, but I will put in an effort.
3: implement pause feature
 * I am already screwing with the timer with the speed feature, this is barely any more complicated

Jowan-Spooner:
yeah I think you got it although

1. I'm not agains the [play] command, I just don't find it too useful if you cannot play custom sounds, so I would see myself using the [signal] command most of the time
2. This was just thinking ahead and doesn't have to be done for this PR.
3. yeah pause should be easy. Just implement the command and then we can do the auto-insertion in a different PR.

KvaGram:
Ok, here is my thinking right now.
When it comes to future-proofing to allow for commands with multiple arguments, I can't come up with any idea to do this generically.
But since implementations of said commands would be in the same codefile, it does not matter much. Adding additional capturegroups when the need arise should not be much of an issue. Nor is just writing the arguments with comma seperation, or some other seperation character.

With the pause feature already implemented, and that's left is to deal with the play command. And I think I got an idea for that.
We just add another folder in 'dialogic' (the user-folder). Call it 'sounds'. Then play will look for playable audiofiles there.

zaknafean:
> 
> With the pause feature already implemented, and that's left is to deal with the play command. And I think I got an idea for that. We just add another folder in 'dialogic' (the user-folder). Call it 'sounds'. Then play will look for playable audiofiles there.

I'd really want to be able to organize sound assets in a custom way, and don't think forcing generic assets into a static location is going to work for most peoples projects. Maybe a setting for the path instead?

KvaGram:
> > With the pause feature already implemented, and that's left is to deal with the play command. And I think I got an idea for that. We just add another folder in 'dialogic' (the user-folder). Call it 'sounds'. Then play will look for playable audiofiles there.
> 
> I'd really want to be able to organize sound assets in a custom way, and don't think forcing generic assets into a static location is going to work for most peoples projects. Maybe a setting for the path instead?

I see where you are getting at. However, for the intention of this feature, I disagree.
This is intended to be a quick and easy way to play sounds. While I could implement support for pasting in a whole filepath in the argument, the problem is that it would just end up being too long and bulky.
I would rather have `Hello my name is [play=ding] Lars` than `Hello my name is [play=res://assets/mygame/subfolderB/scene56/effects/ding.mp3] Lars`.
I might add support for both, it shouldn't hurt. But I do hope you see the problem I'm trying to work around here?

Jowan-Spooner:
I agree that having the full path in the text editor is really ugly but if I understand @zaknafean correctly he was talking about creating a setting (in the Settings menu) where you could select a folder that would be used for the play command, so in the end it would still mean a short word in the editor

zaknafean:
Pretty much that. Ideally I could point to the root assets folder, then pass the argument something like 'char/john/gasp.ogg' in the event text, which while long, still lets me keep everything in one place, organized and readable.

KvaGram:
I have had a simular idea to that, @Jowan-Spooner . And I agree that would be a nice way of doing things. I don't know how to implement such a setting, but if you can, then please do.
Also: I tried to implement the play command as I described, now. However, I hit a brick wall while testing. I can't make it make a sound. I don't know if I am doing something wrong. Please, give it a go, please tell me what I must have done wrong.

PR #624, opened 2021-11-08
Trying to remove clutter from the Timeline Editor

Jowan-Spooner:


Jowan-Spooner:
Small update: This is likely going to be replaced by an update from emilio, whenever he finds the time to do it.

PR #625, opened 2021-11-10
Recreation of History Functionality

zaknafean:
Cleaned version of PR#591


zaknafean:
Back to about where I was previously, but with all the new merges and cleaner code. 
Summary of the Changes as follows:
Settings.tscn/Settings.gd
image
This and the code to make that work.

History.tscn/History.gd
Bulk of the new code and functionality is encapsulated here

HistoryRow.tscn/HistoryRow.gd
Helper node for History. A new history row is created whenever something is logged. This is a separate node so that advanced users can clone and customize it to their hearts content if they so desire

DialogNode.tscn
Added the History node to this

DialogNode.gd
Added logging statements. Added initial theme loading

Testing:
Fresh install of my branch in a new project. History starts disabled, upon enabling functoins as expected.

Known issue: 
Currently if you use mouse click to advance dialog, the history will pop up AND the dialog will advance. This can likely be fixed with a new History state, but I wanted to get this part done before I muck with DialogNode more.

zaknafean:
Cleaned up the messes and fixed the known bugs. I think we're good for a review here!

Jowan-Spooner:
Suggestions (as already discussed):

- [ ] History close button says "Test" right now
- [ ] If you use a definition as the player display name it doesn't get replaced correctly. My players name is defined by the [PLAYER_NICKNAME] definition and in game it says "John" but in the History it says [PLAYER_NICKNAME].
- [ ] easily customize phrases like "has arrived" (remove the : for these?) 
- [ ] Option to disable the "showing all options" for questions
- [ ] Options for what should be saved and stuff
- [ ] Disabling of default open and close buttons (separately as buttons inside a popup can be difficult to get right)

zaknafean:
Still alive! Got all the discussed upgrades aside from the nickname bug fix, which shouldn't actually be too bad. Will check in after a clean test tomorrow. Preview of new options... its busy unfortunately, but I think any good logging system is always a bit more complex than one would like.
 
image

zaknafean:
And I think I'm good for a review when you get another chance. All discussed options are added.

zaknafean:
Rebased it so its up to date. Doing some more from scratch testing on it

zaknafean:
Testing looks good, and ready for review

PR #632, opened 2021-11-17
Automatic Pauses for Specific Symbols

Jowan-Spooner:
The theme menu has now a section called Autopauses, where you can add a unlimited number of special strings (e.g. ".", ",","?","!"). 
Not well explained in the editor yet, but it actually only cares for single characters, so you can add all of these in one field if you want all of them to have the same pause duration.

A known problem right now is, that adding the "." string breakes all other commands that include a dot (like [pause=0.5] or [play=mysound.wav])

### DIFF
--- a/addons/dialogic/Editor/ThemeEditor/AutoPauseContainer.gd
+++ b/addons/dialogic/Editor/ThemeEditor/AutoPauseContainer.gd
@@ -1,0 +1,14 @@
+tool
+extends HBoxContainer
+
+signal data_changed
+
+func _ready():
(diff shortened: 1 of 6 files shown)


coppolaemilio:
If you add more than one character unexpected stuff happens:
image

Jowan-Spooner:
Unfortunately yeah. Too much time has passed. At this point it's easier to recreate it then to make this work again

coppolaemilio:
okay, leaving this one for 2.0 then

PR #634, opened 2021-11-19
Add customization option to the focusout dim color/opacity

timkrief:


coppolaemilio:
Thanks for the PR, I think this should be on by default and remove the change from the settings menu, as it makes more sense to handle this per theme basis. I'll try to make those edits now :)

PR #649, opened 2021-11-28
Added hovering mode

MerlinSleeps:
Added an additional setting so user can use dialogic without pressing buttons.
When this feature is enabled the user just need to hover over the NextIndicator or the ChoiceButtons for some time to simulate an input.


Jowan-Spooner:
Hi @MerlinSleeps I like the changes a lot so far. 
I was wondering if the actual time it takes should be exposed in the settings as well, so you can change it?
Otherwise I would like it to be mentioned in the hint of the setting (e.g. "Activates choices and advances the dialog after 2 seconds of hovering over them/ the next indicator"). I needed to dig into the code to find out what time it used and I don't want people to have to do that.
What do you think?

MerlinSleeps:
Thanks for the review.
Yeah, already added the option to change the timer in my personel projects. I will add those changes to the PR.
Hints also seems reasonable, should be easy enough.

coppolaemilio:
Hello Merlin Sleeps 👋🏻 

I will close this PR because we are not going to keep improving the 1.4 branch at the moment. If you want to keep working on the 1.X branch there are still some bugs that need fixing but are not high priority; you can check the milestone: https://github.com/coppolaemilio/dialogic/milestone/7

Cheers!

PR #658, opened 2021-11-30
closes #657 build error for DialogSharp.cs

### LINKED ISSUE #657: Mono Build Error: Optional parameters must appear after all required parameters
## The problem

**Describe the bug**
When building a Godot Mono project that uses the Dialogic plugin, a build error occurs.

**To Reproduce**
1. Download and install Godot Mono for Windowx x64, and the documented microsoft SDK on the godot website
2. Create a new Godot mono project and attach a C# script to the root node of the main scene
3. Build the project (should build successfully at this stage)
4. Clone this repo in another folder (used commit [10c14d149e562466b8bb5dc62c8d65d6e7844f3f](https://github.com/coppolaemilio/dialogic/commit/10c14d149e562466b8bb5dc62c8d65d6e7844f3f))
5. Copy the `addons/dialogic` folder from the dialogic repo into the project
6. Create a timeline
7. Add a single text event to the timeline
8. Add `addons/dialogic/Dialog.tscn` to the root node as a child node
9. Select the timeline created earlier in the propery inspector for the Dialog node
10. Build the project

**Expected behavior**
Build error:
```
<folder path ommitted for privacy>\addons\dialogic\Other\DialogicSharp.cs(16,102): Optional parameters must appear after all required parameters
```

## Solutions

**Workaround**

No known workaround, it's a build error.

**Possible fixes**

Short term fix: change add a default value to `String dialogScenePath = ""`. This would be backwards compatible for existing users of the library but I haven't tested if this works.

Long term fix: If `dialogScenePath` can't be optional and default to `""`, reorder the parameters.  This would be backwards incompatible and require support for downstream users of the library.

sesopenko:
When building a Godot Mono 3.4 project using the library a build error
occurs:

`\addons\dialogic\Other\DialogicSharp.cs(16,102): Optional parameters
must appear after all required parameters`

Made set a default parameter for `dialogScenePath` so that ordering
rules for C# are respected and the library builds when used in a
project.

See C# documentation for more details:

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments

### DIFF
--- a/addons/dialogic/Other/DialogicSharp.cs
+++ b/addons/dialogic/Other/DialogicSharp.cs
@@ -13,7 +13,7 @@ public static class DialogicSharp
     return Start<Node>(timeline, default_timeline, DEFAULT_DIALOG_RESOURCE, useCanvasInstead);
   }
 
-  public static T Start<T>(String timeline = "", String default_timeline = "", String dialogScenePath, bool useCanvasInstead = true) where T : class
+  public static T Start<T>(String timeline = "", String default_timeline = "", String dialogScenePath = "", bool useCanvasInstead = true) where T : class
   {
     return (T)_dialogic.Call("start", timeline, default_timeline, dialogScenePath, useCanvasInstead);
   }


coppolaemilio:
Thank you! None of the core Dialogic developers work with C# so it is hard for us to test it properly. Thanks again!

sesopenko:
> Thank you! None of the core Dialogic developers work with C# so it is hard for us to test it properly. Thanks again!

How come C# was introduced to the project?  What's it doing that gdscript isn't already doing?

coppolaemilio:
A contributor wanted to use C# to spawn a Dialogic timeline so they contributed that "wrapper". I'm not entirely sure what their project looked like, but many people come to the discord server and ask for it as well, so I guess that people want to make sure they will be able to use C# somehow 🤷🏻‍♂️

DavidBradbury:
`DialogicSharp.cs:16` needs to be updated to either be:

```C#
  public static T Start<T>(String timeline = "", String default_timeline = "", String dialogScenePath = "res://addons/dialogic/Nodes/DialogNode.tscn", bool useCanvasInstead = true) where T : class
  {
    return (T)_dialogic.Call("start", timeline, default_timeline, dialogScenePath, useCanvasInstead);
  }
```

Or check to see if `dialogScenePath` is set, and if not, call a different `start()` overload. Right now, if you run a new C# project, you get the error:

```
ERROR: Resource file not found: res://.
   At: core/io/resource_loader.cpp:287
```

Because it defaults the resource to empty string. I'd put together a PR but probably won't have time until the weekend.

Issue #343, opened 2021-06-05
Translate Dialogues


PonySchreck:
**Is your feature request related to a problem? Please describe.**
At the moment it doesnt seem easy to create games with multiple languages with this plugin.

**Describe the solution you'd like**
It would be great if the Event Editor would have a Translation Tool. E.g. A Dropdown menu where you can select a language to translate to. And if you select a language the nodes stay the same but you could edit the output messages. Dialogic could then check the TranslationServer and output the right Text.

**Describe alternatives you've considered**
It seems possible with a dictionary and translationfiles but this solution seems too clunky and would make the great usability of the event editor senseless just to support mutliple languages.

**Additional context**
I am working on a Game that i would love to export with German and English text. 
I really love the plugin! Thanks for the great work for the community! <3

Jowan-Spooner:
Hi thanks for your request. I have to add that this is related and possibly a duplicate of #249, although your suggestion is more concrete.

I'm not quite sure what the standard procedure is to translate games and I'm not sure whether it is done in the editor or an external tool for most games, especially bigger ones. I would guess that many people might like to translate their game into english and there native language directly inside the editor, in a way similar to what you proposed. But for all languages that none of the developers speaks, translating in external files/programms seems more logical (for example to have community translations). So there should also be an option to export the text of the game into a table like format, that can then be translated and imported back into the game. 

Both versions should be considered when designing a strong translation tool/workflow for dialogic, in my opinion.

What do you think?

Also nice, I'm german too ;).

zakary93:
Not that I have done it myself, but a standard method that many use is to use (at least on an indie-mid sized studio scale) JSON, XML, or even CSV or something similar in usage to send to freelance translators (which many are familiar with these kinds of file formats) to replace text. They are then inserted in the file directory wherever dialogs go and called when the appropriate language is selected in game settings with the appropriate language suffix (e.g. Dialog1_en.json, Dialog1_jp.json, etc). I agree that having an in-editor translation tool would be helpful, but many times it is just outsourced like I described. Maybe, we could have Dialogic spit out a realistically readable JSON file, then create another dialog which can then read the translated JSON that a translator delivers back as an Event in the editor? (just spitballing here to get better ideas flowing)

scottkunkel:
It should use the standard Godot language support. Which allows the use of csv files.
https://docs.godotengine.org/en/stable/tutorials/i18n/internationalizing_games.html

idchlife:
[...]

PR #676, opened 2021-12-13
Add translation preview

Linkyu:
As per #343:
This adds the possibility to preview the translations when using localisation keys for text events, choices, and glossary entries.

image
image

Main changes/features:
 - A new setting is added to tell Dialogic where the translation CSV file is located
 - This file is converted into a more optimized JSON
 - On load, this JSON is loaded as a dictionary and is used for returning the proper translations.
 - Each text illegible text field now has a duplicate field for the preview; it is readonly.
 - The preview is updated on each field update
 - If the key is not found, it is returned as is -- similar to tr().

I've tried to re-use the existing code as much as possible; hopefully this should fit the current conventions of the project.

### DIFF
--- a/addons/dialogic/Editor/EditorView.gd
+++ b/addons/dialogic/Editor/EditorView.gd
@@ -91,6 +91,10 @@ func _ready():
 		$ToolBar/Version.text = 'Dialogic v' + version_string
 		
 	$MainPanel/MasterTreeContainer/FilterMasterTreeEdit.right_icon = get_icon("Search", "EditorIcons")
+	
+	# Serialize the translations file if translation preview is enabled
+	if DialogicResources.get_settings_config().get_value('dialog', 'translations_preview'):
+		DialogicResources.serialize_translations()
 
 
 func on_master_tree_editor_selected(editor: String):
--- a/addons/dialogic/Editor/Events/Parts/Logic/ChoicePicker.tscn
+++ b/addons/dialogic/Editor/Events/Parts/Logic/ChoicePicker.tscn
@@ -36,7 +36,14 @@ margin_right = 76.0
 margin_bottom = 27.0
 rect_min_size = Vector2( 10, 0 )
 
-[node name="ConditionPicker" parent="." instance=ExtResource( 1 )]
+[node name="Preview" type="LineEdit" parent="HBox"]
 margin_left = 80.0
-margin_right = 202.0
-margin_bottom = 27.0
+margin_right = 138.0
+margin_bottom = 30.0
+editable = false
+expand_to_text_length = true
+
+[node name="ConditionPicker" parent="." instance=ExtResource( 1 )]
+margin_left = 142.0
+margin_right = 564.0
+margin_bottom = 30.0
--- a/addons/dialogic/Editor/Events/Parts/Logic/EventPart_ChoicePicker.gd
+++ b/addons/dialogic/Editor/Events/Parts/Logic/EventPart_ChoicePicker.gd
@@ -5,6 +5,7 @@ extends "res://addons/dialogic/Editor/Events/Parts/EventPart.gd"
 
 ## node references
 onready var input_field = $HBox/ChoiceText
+onready var preview_field = $HBox/Preview
 onready var condition_picker = $ConditionPicker
 
 # used to connect the signals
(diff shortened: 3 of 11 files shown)


AnidemDex:
I'm curious about this implementation, what makes you to handle this with a custom file instead of using the translations files that godot generates when a CSV file is imported?

AnidemDex:
https://github.com/coppolaemilio/dialogic/pull/676/commits/57eda0527ffef511c3c6bc4799c1fa0d115f75d2#diff-616ac1667cf9dfa69fa3c625bf839373add6bbb62d2f7065c82671fb433d5ccfR335

Ok, according to this https://github.com/coppolaemilio/dialogic/issues/343#issuecomment-975893194 it was because you weren't unable to get the TranslationServer on the editor, but dialogic already uses the [TranslationServer](https://github.com/coppolaemilio/dialogic/blob/main/addons/dialogic/Localization/translation_service.gd) to translate itself in the editor. Have you take a look to that file?

Linkyu:
Honestly, the main answer to both questions is "I didn't think about it that way", hah, so those are good points.

That said, I did take a look at TranslationService, but as far as I could tell, Dialogic's implementation of it is very tailored to translating the editor itself by using its own translation files. Now that I'm a bit more familiar with the plugin though, I do agree that it would make sense for DTS to be the one in charge of all translations; I'll see how it can be re-written to take that into account. Maybe with a simple parameter in translate().

Linkyu:
OK so I did manage to rewrite it cleanly to take both cases into account, and in a way that only load a translation file as needed. As I suspected though, it is pretty much a full rewrite:

```gdscript
# Translates a message using the locale set in the settings.
# If no translation is found, the fallback locale is used.
# If still no translation is found, returns the original message.
# Can be used for both editor and project messages
func translate(message: String, is_editor := true) -> String:
	# Get translation file
	var csv_file = ""
	if is_editor:
		csv_file = 'res://addons/dialogic/Localization/dialogic.csv'
	else:
		var config = DialogicResources.get_settings_config()
		csv_file = config.get_value('dialog', 'translations_preview_file', "")
	
	# No need to continue if the CSV file doesn't even exist
	if not Directory.new().file_exists(csv_file):
		push_warning("Attempted to translate message \"" + message + "\" but the translation table " + csv_file + " is missing.")
		return message
	
	# Retrieve the locales used in the settings
	# They are ordered by priority
	var locales := []

	if is_editor:
		var editor_plugin = EditorPlugin.new()
		var editor_settings = editor_plugin.get_editor_interface().get_editor_settings()
		var editor_locale = editor_settings.get('interface/editor/editor_language')
		
		locales = [
			editor_locale,
			'en'
		]
	else:
		locales = [
			ProjectSettings.get_setting("locale/test"),
			ProjectSettings.get_setting("locale/fallback"),
			'en'
		]
	
	var translation_files := []
	
	for locale in locales:
		if locale:
			# Translation files are always named <CSV_basename>.<locale>.translation
			var translation_filename = csv_file.get_basename() + "." + locale + ".translation"
			translation_files.append(translation_filename)
	
	for file in translation_files:
		# Load the translation file
		var translation_file:PHashTranslation = load(file)
		# If it can't be loaded, skip to the next one
		if not translation_file:
			push_warning("Attempted to translate message \"" + message + "\" but the translation file " + file + " could not be found.")
			continue
		
		# Retrieve the translation
		var translation = translation_file.get_message(message)
		if translation:
			return translation
		elif is_editor:
			push_warning("Attempted to translate message \"" + message + "\" but the translation could not be found in the locale " + translation_file.locale + ".")
	
	# If no translation has been found, return the original message
	if is_editor:
		push_warning("Attempted to translate message \"" + message + "\" but the translation could not be found in those locales: " + String(locales) + ".")
		
	return message

AnidemDex:
I didn't tried to make you redo the script 😅

But now that you're there, you're using the .csv file and not the translation resource, is there a reason? Godot uses internally translation resources to its localization, CSV files are just the base to import those translations.

Linkyu:
Huh. I couldn't imagine a scenario where there wouldn't be a csv file, but I guess it's possible to not use one indeed. Aight fair enough, in that case no need to rely on it, I'll just get the file paths directly from ProjectSettings. That makes it simpler.

coppolaemilio:
Hello Linkyu, sorry for not replying on this PR. It looks very good quality but unfortunately we are on a feature freeze in the 1.X branch. If you want to keep working on the 1.X branch there are still some bugs that need fixing but are not high priority; you can check the milestone: https://github.com/coppolaemilio/dialogic/milestone/7

Otherwise feel free to familiarize with our work on the 2.0 branch and feel free to contribute there.

Thank you and sorry again 😅

Issue #231, opened 2021-04-18
Set a different theme for each character


arnaudvergnet:
**Is your feature request related to a problem? Please describe.**
It could be interesting to have a different theme for each character (for example Undertale) to give them more life.

**Describe the solution you'd like**
For each portrait/Expression, allow choosing a theme like we can choose a protrait.

**Describe alternatives you've considered**
Right now we must change the theme each time the character starts talking, which is really time consuming a prone to errors.

Jowan-Spooner:
I want to get back to this, as this feature was requested by EmmaH. 

### What?
Basically you add a theme selector to the character editor and whenever that character says something this theme is loaded, and afterwards the default theme is loaded back. 

### Why?
Now while I don't fully know what EmmaH wanted to use this for, I think this would be a cleaner solution to #457 and #364 then adding these settings individually. Why? Because
A) the character editor's UI is pretty messy already as described in #497. Adding more settings wouldn't help with that?
B) There will surely be some other theming feature that someone would like to set per character, next week. So over time we would just replicate all the theme settings in the character editor.

### The problem
While not totally necessary for this to work, Theme inheritance (#232) would drastically improve the usability of this feature. If it doesn't exist having many very similar themes will get annoying very soon when you want to change one thing for ALL characters.

timkrief:
Theme inheritance would definitely be necessary for this to make sense, but it would be an elegant solution to multiple issues.

PR #680, opened 2021-12-14
Custom Theme per Character 

Jowan-Spooner:
You can now select a theme for each character. 
If the character says something the theme will be loaded. If then a character without a custom theme will says anything, it will go back to the current theme.

An implementation of #231.


zaknafean:
Bug Report:

1. If you delete a theme that's in use by a character, it breaks weirdly. The characters theme field will correctly say 'No custom theme' but your game will crash when it gets to that character until you choose a new theme. That fixes it, but even restarting Godot doesn't otherwise fix it.
2. There doesn't appear to be a way to remove themes once you choose a custom theme

So far thats it! I'll see if I can't dig more bugs out of it.

Jowan-Spooner:
@coppolaemilio if you don't test this soon, I'll just merge it!

Issue #21, opened 2020-12-20
Translate Dialogic UI to multiple languages


coppolaemilio:
Initial work for this one has been done. This will be a tracker now of the things left to translate:

- [ ] Timeline Editor (Event buttons)
- [ ] Character Editor
- [ ] Definition Editor
- [ ] Theme Editor
- [ ] Settings

Jowan-Spooner:
I would like to help. I could translate to german. Is this already possible or just a dream?

coppolaemilio:
@AnidemDex is working on this one, but I'm not sure of the state of it

AnidemDex:
Take a look to https://github.com/coppolaemilio/dialogic/pull/127
The plugin can be translated to any other language, you can [see the CSV files here](https://github.com/AnidemDex/dialogic/tree/translation/addons/dialogic/Translations) and the editor will be translated after reload. Anyway, there's no translations to events, yet

PR #702, opened 2021-12-29
German Translations and Translation Updates

Jowan-Spooner:
Although my translations probably contain many mistakes there is now a german translation!

I've also reorganized the translation file.

I've added all missing translation items and removed old ones. Everything can be translated now, except the event parts. I might do that later or in a follow-up PR. 

The ...

- Theme Editor
- Character Editor
- Value Editor
- Glossary Editor
- Settings
- MasterTree (Right-Click Menu, Settings and Help item)
- Toolbar 
- RemoveConfirmationDialogs

... should now all be fully translatable, except for items that would break functionatlity if translated.
The german translation exists for nearly all items. 

A couple of places have seen UI-Updates, just because I've finally taken the time to do something about them (GlossaryEditor, some settings)

Fix for #21


coppolaemilio:
I see that you also commited the automatic changes to the main scenes (like the hints inside the .tscn files), not sure if we want to merge those

Jowan-Spooner:
Hi I know this is a bit late, but I did keep them in because I didn't want to check each of the scenes and had to change some things for a few of them (I changed the translation key on some TLabels, etc.). Just so you know why I did it.

Issue #694, opened 2021-12-27
Prevent Dialog Node from being removed at the end of the timeline


mechPenSketch:
**Is your feature request related to a problem? Please describe.**
I want my dialog node to stay even after it clear its last event.

**Describe the solution you'd like**
An option to clear the dialog node at the end of the timeline can be added to the Theme, specifically Dialog Text > Behavior.

**Describe alternatives you've considered**
* A new event could be added, or `Wait Seconds` could be toggled whether there should be time to be waited.
* I have previously coded the dialog node to not be removed, while it was under the singleton system (can't exactly remember how). I don't think the new node have the functions that lets me do it.

**Additional context**
This is a stylic choice for games where the dialog system to too well-integrated into the game space.
* I can think of the battle system in pre-gen V Pokémon mainline series as an example. 

I can pull a new request in myself. I just need to check whether this can already be done, or there is a better solution.

Jowan-Spooner:
Afaik there isn't a way to do this yet.

PR #705, opened 2022-01-02
Don't close after last event

mechPenSketch:
This PR adds the option to prevent the dialog box from closing after the last event.

The new check can be found at Theme > Dialog Text > Behavior > Closes After Last Event. If checked, then at the last event, the next indicator won't show up. Pressing the usual button won't close the dialog, either.

~~First test project: [dialogic_sandbox.zip](https://github.com/coppolaemilio/dialogic/files/7798971/dialogic_sandbox.zip)~~
Latest test project: [dialogic_sandbox.zip](https://github.com/coppolaemilio/dialogic/files/7879877/dialogic_sandbox.zip)
Check the option "Don't close after last line" in the theme editor to test for yourself.

My Project to record my progress: https://github.com/mechPenSketch/dialogic/projects/1

Closes https://github.com/coppolaemilio/dialogic/issues/694


mechPenSketch:
First Bug: Clicking next does not go to next event (unless there is another button to do that?).

mechPenSketch:
Should I also update the changelog in this PR since I'm adding a feature?

Jowan-Spooner:
Okay, I've tried this and here are my thoughts:
I think it would be a very good feature to have. It took me some time to get it working though, because I didn't realize the last event would have to be a text event. I would like it, if it just kept the box there in it's current state regardless of what event I end with. 
Also I'm really wondering if this should be a theme option or a setting, but I'm fine with both. What do you think @coppolaemilio ?

mechPenSketch:
^ That would be ideal, though I would need to revise the algorithm to account for all combinations.

mechPenSketch:
Here is a new test project for my new update: [dialogic_sandbox.zip](https://github.com/coppolaemilio/dialogic/files/7863535/dialogic_sandbox.zip)

Jowan-Spooner:
Couldn't you just skip the deleting part? I feel like this checking is a bit over-complicated? 
I thought you would just make a check before deleting the node in _load_next_event(). If the option is checked you don't delete the node and maybe set the nodes state to a `finished_timeline` state or something. Then make sure that if that state is on, you cannot continue or do anything else.

mechPenSketch:
^But by doing so, the next indicator will still appear at the last text event.

Jowan-Spooner:
Hmm. maybe I understood the problem wrong but I thought that you want it to clear the dialog box and have the empty dialog box remain after the timeline ended, not that you wanted to keep the text in there. 

Btw, if you have discord feel free to join emilios discord where we discuss most developement.

mechPenSketch:
I've decided how to handle when the last event is other than Text Event. On the last Text Event, the next indicator still appears, but after pressing it, the subsequent events are played, but the dialog won't close. The previous text is also kept through.

Here is my newest test project: [dialogic_sandbox.zip](https://github.com/coppolaemilio/dialogic/files/7879877/dialogic_sandbox.zip)

Jowan-Spooner:
@mechPenSketch There seems to be a problem because sometimes you use is_last_dialog and is_last_text. Should they be the same? Godot gives me an error because one is not defined.

mechPenSketch:
Ah, that's my bad it's supposed to be consistently `is_last_text`. I must've missed out on updating those property names while edit transferring my edits from the test project to my branch.

Jowan-Spooner:
Looks good to me, will just have to fix the merge conflicts.

PR #707, opened 2022-01-03
add method to clear all variables

bitbrain:
When using Dialogic with multiple save games we need to somehow "reset" Dialogic, as each save game will have its own state that will be populated on savegame load. Currently, it is difficult to reset variables from gdscript so I introduced a new utility method to do exactly that.


coppolaemilio:
Hey, thanks for your first contribution :)

PR #710, opened 2022-01-06
Add Method to Change Timeline

mechPenSketch:
With this PR, you can call `Dialogic.change_timeline(new_timeline)` to make the dialog node change timeline and load its new event on the spot.

To-do next while waiting for approval:
- [ ] Update Documentation
- [ ] Provide test project

### DIFF
--- a/addons/dialogic/Nodes/DialogNode.gd
+++ b/addons/dialogic/Nodes/DialogNode.gd
@@ -706,8 +706,7 @@ func event_handler(event: Dictionary):
 		# TIMELINE EVENTS
 		# Change Timeline event
 		'dialogic_020':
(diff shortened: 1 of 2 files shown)


Jowan-Spooner:
Thanks a lot for the contribution! I like the simplicity of this. Nevertheless there are some things to think about here:

The DialogNode will preserve it's current state. It will NOT change to the default Theme, Characters won't leave, Background Image and Music will stay. This is a very cool thing and allows for stuff that wasn't possible previously. BUT it can also lead to unexpected/unpredictalble behaviour. I think this should be mentioned in the Documentation as it is the main difference to starting a new dialog node.

Also I would like the function to have at least a small explanation in the script file. And please move the documentation up below the start() function.

PR #722, opened 2022-01-12
Update so Mac-hotkeys work

### LINKED ISSUE #720: Mac Hotkeys are not working properly.
Things like `Command + Z` are not working and are still bound to the `Control` keys. There is also no way to remove events using the `Command + delete` hotkey.

Jowan-Spooner:
Should fix #720


coppolaemilio:
Works great! Thanks :)

PR #723, opened 2022-01-13
Create variable if doesn't exist in `set_variable`

LoipesMas:
Does what `TODO` mentioned there. I hope that's the correct way to do this.

### DIFF
--- a/addons/dialogic/Other/DialogicClass.gd
+++ b/addons/dialogic/Other/DialogicClass.gd
@@ -216,9 +216,9 @@ static func set_variable(name: String, value):
 			d['value'] = str(value)
 			exists = true
 	if exists == false:
-		# TODO it would be great to automatically generate that missing variable here so they don't
-		# have to create it from the editor. 
-		print("[Dialogic] Warning! the variable [" + name + "] doesn't exists. Create it from the Dialogic editor.")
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
First of all, thanks for the contribution!
But... I'm not sure I like this change. I know emilio wrote that he would like this, but I feel like this could leed to very strange behaviour. 
I would definitely want to keep a notification in the output, because now you won't be notified if you made an typing mistake. 
Maybe both would be okay, but again, I personally don't know what the use behind this would be.

zaknafean:
I'd definitely want this as an optional setting with a big warning. I can see where it would be handy if you're making lots of arbitrarily named variables in your game, but as a default behavior, it seems... dangerous.

coppolaemilio:
Yeah, sorry if I was confusing on that comment, but I believe it would be nice to create "local" variables to create and use only in the context of that timeline. I don't think that it is a nice default behavior since it can create more problems that it would solve.

LoipesMas:
Now that I think about it, I agree that it is a questionable idea at best. Because in the editor you can't use definitions that are not defined in the editor, so the only place that it would be useful would be for timelines generated at runtime, which I think is pretty niche.

PR #730, opened 2022-01-15
Updated Event Generation

zaknafean:
Updated event generation to use the state system rather than the old Booleans


Jowan-Spooner:
Hmm, I would like the comment about the states to make clear that you are not supposed to use some of them, because your logic will clash with the DialogNode. I think you should only use WAITING/WAITING_INPUT (basically the same behaviour) and READY if your event finished but you want the next action to be performed before continuing. The other are not useful imo, but feel free to explain what you would do with them.

PR #747, opened 2022-01-25
Character Event and Animations with Anima

### LINKED ISSUE #744: Move "mirror portrait" setting to a checkbox inside the body of the event

Jowan-Spooner:
ONE CHARACTER EVENT TO RULE THEM ALL
- Adds the new Character event with three modes (JOIN, LEAVE, UPDATE)
- Adds new CharacterAction and CharacterActionSettings event parts
- Tries to convert all Character Join and Character Leave events into this new one (has id 'dialogic_002')

- adds simple animation support for character leave (but this shall be replaced with anima, as all animations should)

ANIMA
I feel a bit bad, because I don't understand completely how the code I ported works. But it works. for now. There is probably parts in there, that we don't even need. Maybe for 2.0 we can take another look, who knows.

Also added 'change' controls, so a Character Update event doesn't have to update everything, only the things you want.

Also adds a repeat option that repeats the animation the given time. 

Also added a wait for animation to finish option. 

Also added default animation setting. These can be (for join and leave) set in the settings menu.

Should also fix #744


coppolaemilio:
Upon testing an existing project I get all this notifications. I assume it is because of the timeline event scene changes, but just wanted to flag just in case:
image

I'll keep testing since even with the errors everything works as expected

coppolaemilio:
That's all I think. Feel free to either fix these or just merge as is and we can fix/patch later on. I might change a few things (like the order of the effects and a few settings like that) but nothing that would block this one to get merged. Thank you so much, it seems like it was a lot of work!

Jowan-Spooner:
@coppolaemilio You are right, we still need to add the license. I'm not too familiar with the MIT license. Where should we put it? Do need to say that it was modified?

Also I've pushed another fix that should remove all the warnings hopefully.

Issue #746, opened 2022-01-25
Folder paths do not work for get_variable/set_variable


thebardsrc:
Problem: 
Folder paths are not working for the get/set variable functions.

With the following test variable layout, with each variable value set to the folder it is in:
<img width="89" alt="Capture" src="https://user-images.githubusercontent.com/2214793/151053246-9586948f-92b4-48d8-aa13-541d1e78cb8f.PNG">

The tests show that the folder paths are being completely ignored:
<img width="303" alt="Capture2" src="https://user-images.githubusercontent.com/2214793/151053707-7c97e76a-4568-488f-85b7-8261c3ed873c.PNG">

Also, as the same case with my addendum on #739 testing found with the timelines, it's also simply going by the order in the JSON, which means in this example screenshot it's actually pulling the value of Test2/TestValue, as that one was created first, rather than taking the root TestValue since that was created after it, which also can lead to unexpected behavior.

thebardsrc:
From additional testing, I have found two additional problems:

First, it also doesn't get them parsing the text in a text node, either:
<img width="701" alt="Capture3" src="https://user-images.githubusercontent.com/2214793/151062149-b4267765-f394-4479-8429-3aa57c2f4055.PNG">

Second, something weird is happening when calling set_variable to the way Dialogic does logic calls. 

Take the following test timeline:
<img width="404" alt="Capture4" src="https://user-images.githubusercontent.com/2214793/151062486-18f05c82-3adb-4b48-9209-a49262b34b47.PNG">

While the If statement doesn't reflect the subfolders (which is not ideal, but fine since in the background it uses the event_id instead), each one of these If statements have been set separately: /TestValue has been set to 1, /Test2/TestValue has been set to 2, /Test3/sub/TestValue has been set to 3.  when running normally, since all three of them are set, they all execute correctly and show the text event assoicated to them. However, calling Dialogic.set_variable('TestValue', "updated value") in code before creating the dialogue isn't just updating the first found TestValue, as we were expecting from the previous test, but it seems to be updating _all_ of them because then _none_ of them are running. 

And if we print Dialogic.export() from before and after doing anything, we can indeed see that all three of them were changed, which isn't the behavior we want either:
<img width="619" alt="Capture5" src="https://user-images.githubusercontent.com/2214793/151063805-8d607df5-be05-40e0-8227-6690e24cbc23.PNG">

Jowan-Spooner:
Thoughts/Info
[...]

Issue #739, opened 2022-01-21
Test If A Timeline Exists


thebardsrc:
I needed a way to test if a timeline with a specific name existed, for automated validation testing for my project. There is a function that almost does exactly waht I need, but it's internal. So I made up this quick external one:

static func timeline_exists(timeline: String):
	var timeline_file = _get_timeline_file_from_name(timeline)
	if timeline_file:
		return true
	else:
		return false

thebardsrc:
as an addendum to this, the _get_timeline_file_from_name() function seems to fallback for searching the entire tree for a timeline, so it can't be used directly the way I have it for example. this also means that calling Dialogic.start() without an actual path could lead to calling the wrong node since it just goes for the first one it finds in the JSON order, rather than the sorted Dialogic order, which would seeem to be a bug:

Example timelines setup: 
<img width="111" alt="Capture" src="https://user-images.githubusercontent.com/2214793/151050418-3604a424-6792-42a0-b9ab-021fc9d987b2.PNG">

The absolute paths work for the most part, except for checking if one's in the root that one returns True even though there isnt one there. In this setup, the Test3/Sub/Auto timeline was created before Test2/Auto, so a call to Dialogic.start('Auto') ends up creating the subfolder timeline since its the first one found in creation order:

<img width="776" alt="Capture2" src="https://user-images.githubusercontent.com/2214793/151050605-8061794e-dbd0-41cf-b8ee-0ef677151eec.PNG">

if you were to call Dialogic.start('Test2/Auto'), it will call the proper timeline. also trying to call Dialogic.start('Test1/Auto') will give the  "No Timeline Speicified" error as expected, since that one doesn't exist.

Jowan-Spooner:
Using the full path is the recommended way of doing it. Using just the timeline name (which doesn't always work as you pointed out) was only kept in to not break compatibility to older projects.

thebardsrc:
yeah I agree, and it seems just fine on that account other than trying to address any timelines in the root folder

PR #748, opened 2022-01-26
Fixes for addressing Variables and Timelines by folder path

thebardsrc:
This is to add the features for getting/setting and display in a text block for a variable by the folder path name, as in #746. Also, as this was based on the function for doing so in timelines, fixes a bug in the existing timeline folder search code as mentioned in #739 which prevented either that or the new variable search to work on the root of the folder tree

Also included is the command to test if a timeline exists, as requested in #739


Jowan-Spooner (approved):
Looks and works very well!

PR #767, opened 2022-02-06
Timeline and Variable Folder searching bugfix

thebardsrc:
Fixes deeper level folder depth bugs not previously found in PR #748 before it got merged


Jowan-Spooner:
Good catch! Thanks for the fix!

PR #774, opened 2022-02-17
Sort the EventButtons

Jowan-Spooner:
Probably not the best way to do this, but at least it's ordered now.

The EventTemplate has now a sorting_index property. Lower indexes will be further up in the list.


coppolaemilio:
I see you sorting things, you are ready for a google interview 🤣

Issue #704, opened 2022-01-01
New timeline editor style tracker


coppolaemilio:
Making a list of all the things I know I have to finish before releasing this update:

- [x] New high resolution icons
	- [x] Text event
	- [x] Character Join
	- [x] Character Leave
	- [x] Question Event
	- [x] Choice event
	- [x] Close branch
	- [x] if event
	- [x] set value
	- [x] Set Glossary
	- [x] Label Event
	- [x] Go to Event
	- [x] Save
	- [x] Change Timeline
	- [x] Set Background
	- [x] Close Dialog
	- [x] Wait event
	- [x] Set theme
	- [x] Audio Event
	- [x] Background Music
	- [x] Emit Signal
	- [x] Change Scene
	- [x] Call Node
- [x] Add right click menu with the options that previously lived in the three dots menu
- [x] Question events base line not showing after the first indent level
- [x] If event indentation flow lines connections are broken
- [x] Better selection style
- [x] Text and Question text box never wraps for me, if you have long lines
- [x] Events body needs some indentation
	- [x] Audio Event
	- [x] Background Music
	- [x] Call Node
	- [x] Set Definition
	- [x] Audio Pickers on the Text Event
- [x] Add a "Help" option to the right-click-menu because the button got removed
- [x] Move the "Warning" Icon to somewhere else, because it looks pretty strange as it is

Jowan-Spooner:
I've added a couple of points to the list, hope that's okay. 

I would also like to discuss some kind of standard look, placement and sizing of the contents of the events, so that the event parts could get cleaner and more intuitive as well. From what I can see, these are the most common things on events:

## Things on events
- [ ] _Resource Pickers (Character, Timeline, Definition, etc)_
Some have no Box around them right now
![grafik](https://user-images.githubusercontent.com/42868150/148844365-5a2bfe89-831f-4687-be31-d382ba76d9e8.png)
Some have:
![grafik](https://user-images.githubusercontent.com/42868150/148844528-81911b09-d007-4262-979e-62dcd5f68043.png)

- [ ] _Subresource Pickers (Portraits, Labels?)_ 
I would also imaging the Change Timeline event with a "Label" picker to start at a specific label
The Portrait Picker looks identical to the Character Picker which is probably okay, but it is not a resource per se, so maybe a difference could be good? not sure.
[...]

PR #780, opened 2022-02-19
Make TextBox wrap when it get's to big

Jowan-Spooner:
Should help with #704

I also decided to show the line numbers again, because I just like that more.


coppolaemilio:
Nice! Not really sold on the numbers, but I'll merge 👍🏻

Issue #666, opened 2021-12-05
Search portraits and organize them


Bricklou:
**Is your feature request related to a problem? Please describe.**
I'm currently working with a lot of portraits for my characters. This mean I have a **lot** of entries in my portraits list. I'm losing a lot of time just to search one.

**Describe the solution you'd like**
It would be nice to have a search bar to filter entries.

Another features that could be interesting would be able to organize them inside subfolder to have some sort of "state"
(ex: normal/happy, normal/sad in one folder and equiped/happy, equiped/sad, etc...)

coppolaemilio:
Leaving open and also this could be related to: https://github.com/coppolaemilio/dialogic/issues/820

Should portrait be resources? dunno, leaving open to discuss eventually

Bricklou:
Maybe portraits could have some tags or state per character, or sub-categories for each character.

Jowan-Spooner:
I know it's been a long time, but I can pretty safely promise that both a search bar as well as groups will be in dialogic 2.0. Hope we can get it to a beta soon.

PR #781, opened 2022-02-19
Implement a search bar for portraits

Jowan-Spooner:
Although not a full fix, this should help with #666


coppolaemilio:
It looks really good! I never used that many portraits, but hope it helps yeah :)

PR #790, opened 2022-02-21
Add easy way to play the current timeline

Jowan-Spooner:
Adds a button that allows to instanly play the current timeline.

- New Button in the toolbar (when a timeline is opened)
- New "Testing" scene
   - prints the argument if it recieves a dialogic signal
   - ends the game when the timeline ends


Jowan-Spooner:
grafik

PR #804, opened 2022-02-27
Small UI updates

Jowan-Spooner:
Character Editor:
- enabled wrap on the description box
- made character name and display name fiels expand if the text get's to long
- updated the theme picker style to match the resource pickers in the rest of the UI
- added the search icon to the search bar and enabled the clear button

Events:
- Change the default text of the glossary picker to "Select Glossary Item" instead ot "Resource"
- Made sure that if a glossary is deleted it won't result in an error, but the event will reset
- removed the warning logic in the condition picker, because it could be, that the string is the name of a value definition


Jowan-Spooner:
Sorry making this PR from my main was a mistake.

coppolaemilio:
I can't see these:


- enabled wrap on the description box
- made character name and display name fiels expand if the text get's to long
- updated the theme picker style to match the resource pickers in the rest of the UI
- added the search icon to the search bar and enabled the clear button

coppolaemilio:
my bad lol

PR #819, opened 2022-03-05
:memo: add documentation about defining custom portraits

bitbrain:
When I first started using Dialogic I was quite confused by how to use custom portraits. I hope this clarifies this a little.


Jowan-Spooner:
Thank you very much!

PR #821, opened 2022-03-05
History Default theme adjustments

zaknafean:
HistoryRow now use the default theme
HistoryRow themeing can now be disabled
Default Background now matches default theme

### DIFF
--- a/addons/dialogic/Example Assets/History/HistoryRow.gd
+++ b/addons/dialogic/Example Assets/History/HistoryRow.gd
@@ -38,44 +38,56 @@ func add_history(historyString, newAudio=''):
 	AudioButton.focus_mode = FOCUS_NONE
 
 
-# Uses the default theme to make the font and boxes looks a bit nicer
+# Load Theme is called by 
 func load_theme(theme: ConfigFile):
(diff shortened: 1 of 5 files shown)


addons/dialogic/Example Assets/History/HistoryRow.gd
```diff
@@ -38,44 +38,56 @@ func add_history(historyString, newAudio=''):
...
-	
+
 	# Margin
 	var text_margin = theme.get_value('text', 'margin', Vector2(20, 10))
-	TextLabel.set('margin_left', text_margin.x)
-	TextLabel.set('margin_right', text_margin.x * -1)
-	TextLabel.set('margin_top', text_margin.y)
-	TextLabel.set('margin_bottom', text_margin.y * -1)
-	
+	$HBoxContainer.set('margin_left', text_margin.x)
```
bitbrain, line 76: Can we extract the HboxContainer lookup into an onready var? Using the $ operator will do a full node tree scan.

zaknafean: Easy enough!

PR #824, opened 2022-03-08
Replace `call` with `callv` in event 042

AnidemDex:
This makes call event pass each argument in the array instead of passing the whole array.

As an example, consider call event `my_method` with `["1","2","3"]` values to an arbitrary node

```
func my_method(args) <- Current behaviour: call event pass the whole array, so you need to split it manually
```
```
func my_method(arg1, arg2, arg3) <- PR behaviour: each value in the array is passed as an argument
```

### But wait
I consider this can be annoying to people that already uses call event and split arguments manually, so maybe thay should be notified about this change? No idea


coppolaemilio:
I'd be okay breaking compatibility in this one and adding a warning on the changelog. People expect this event to work like how you have it in this PR

bitbrain (approved):

PR #825, opened 2022-03-09
🗃️ improve German translation

bitbrain:
The existing translation for German sounded a bit odd. As a native German speaker, I tried to make it a bit more natural sounding.


bitbrain:
Do I have to manually generate the translation files? Or how does that work?

coppolaemilio:
The editor generates those, but now that you remind me of this I think I need to add tons of new labels to the translation file. It is a bit hard to keep up with it 😅

bitbrain:
@coppolaemilio I am happy to wait for those and provide the German translations as well!

coppolaemilio:
I'll merge now to avoid conflicts later 👍🏻 
Thanks!

Issue #761, opened 2022-02-03
Enable More Positional Control of Dialog Box


AnomalyAces:
I would like to be able to set an offset on the entire dialog box, similar to the way we can set an offset to the next indicator

Thanks

thebardsrc:
from discussion on the Discord, it seems this might be a bug with any of the positions in themes that are based on Center (the four cardinal centers, and middle), where it doesn't take into account the Box Margin in the positional.  Top and Bottom Center are ignoring the X, Left and Right Center are ignoring the Y, and Center is ignoring both

AnomalyAces:
For more context, It also seems there is a soft cap of 100 for the choice button offsets as well but you can choose any negative number you want. So the current workaround is to position them in the bottom right and use negative numbers to position them. It also may make more sense to position the buttons relative to the dialog box if possible

In a similar vain if you position the dialog with without using one of the 5 "Center" positions you can use the margin to position those as a workaround

zaknafean:
@lamsorsen PR is merged into main and handles this stuff very nicely. Much easier to do cool stuff now

PR #826, opened 2022-03-09
Enable More Positional Control of Dialog Box

lamsorsen:
Resolves this issue:
https://github.com/coppolaemilio/dialogic/issues/761

In action:
dialogic_margins

I think the margin and padding settings had some weirdness in the first place, so it's a little hard for me to say if I did a good job preserving functionality. The *box margin* is the margin from the textbox to the edge of the screen and the *text margin* is the margin between the textbox and the actual text within.


bitbrain:
Nice change! Will this be backwards compatible? What is the default behaviour?

lamsorsen:
@bitbrain It's backwards compatible. Default it does exactly what the old behavior did. Testing it on the default theme that comes with dialogic, it gives exactly the same result. It uses new internal variables, but it checks the old variables in case those were defined in an older version.

zaknafean:
It may not be the 'prettiest' setting, but you can't argue with how effective it is. Rebased, merged as we continue towards 1.4.2! Thanks for the contribution!

Issue #779, opened 2022-02-18
Using a custom Portrait scene means the portrait is misplaced


Jowan-Spooner:
## The problem

**Describe the bug**
Still a bit unsure how to work with the custom scenes but it seems that once I joined my character with the custom scene, all the other portraits aren't visible anymore.

It should recalculate the position imo.

zaknafean:
This should be fixd with #829 now being merged

PR #829, opened 2022-03-13
Re-Center Portraits Option 

thebardsrc:
Adds a toggleable option to enable re-centering the portrait each time it is updated by the dialogue timelines, which is enabled by default. This fixes a number of various issues that come from the old behavior:

- Centering of portraits was done only on a join or update on whichever portrait was set, so if using portraits of different heights that are supposed to be flush against the bottom of the screen, there would be a gap, and they would not be centered if they had different widths
- Fixes inconsistency with custom scenes as portraits. as Nodes don't have dimensions, Dialogic discards the position on a Join/Update, leaving the custom scene to define it's positions manually. But since it was only happening on Join/Update, changing to a custom scene portrait a through a text event would keep the initial Join/Update position settings, requiring a _different_ setup in the custom scene

Since this is significant change in behavior, this is made to be a toggle. I have set it as enabled by default, though, as new projects should be using this by default for consistent behavior expectations.


Jowan-Spooner:
Note: Mention this "compatibilty-breaking" change in the changelog and say it needs to be disabled for old behaviour.

Jowan-Spooner:
The problem is that I have very little experience with custom portraits so it's hard for me to test and compare this.
Maybe @bitbrain could you take a look at this?

Jowan-Spooner:
This should theoretically fix #779 if I understand it correctly.

thebardsrc:
it should, yes

Issue #841, opened 2022-03-19
Dialogic 1.4.1 doesn't work on Android


Lutra360:
## The problem
The game crash.

**To Reproduce**
1. Export to android
2. Start the game

**System (please complete the following information):**
 - OS: Android
 - Godot Version: 3.4.2
 - Dialogic Version: 1.4.1

Jowan-Spooner:
Cannot recreate. Only exported to my phone using the built in testing tho. 
Please provide more information, like Error messages, how exactly you exported, etc. Thanks.

hicchanwan:
On android, version 1.4.1 starts normally, but when you click on the screen, no action occurs. I don't understand what the problem is

MecoCL:
I can confirm, it keeps happening to me, I export the game successfully then I play it and when I load the Dialogic scene, it crashes without reason!

zaknafean:
Got hit with possibly this bug today! Here's the output I got. It doesn't sound exactly like whats being reported here, but my HTML5 and Windows builds to work fine.

![image](https://user-images.githubusercontent.com/7741797/165779855-7fb222d4-9091-4288-9897-406d8869d558.png)

zaknafean:
After digging around, this appears to be some sort of 'namespacing' issue specific to Dialogic and Android.

If I change the name of the ```res://dialogic``` to anything else, I export fine (though everything else breaks). Going to confirm my results by making the export work.

zaknafean:
Ok so after a long day I figured out what I THINK is the root cause of all this nonsense, unfortunatley the original error for this issue is sparse on details and it doesn't QUITE match @MecoCL issue either, but overall it may still be the fix.

Issue is in expott_plugin.gd, which is supposed to remove the need to manually add .json and .cfg to the export templates. At the moment it doesn't do that 100% of the time, and it can cause files to double up in an Android export zip, which breaks the build pretty badly.

I'm able to avoid the issue by adding 
```
# This fixes the android export 'multiple file in zip' bug
	for feature in features:
		if feature == "Android":
			return
```
To the top of the _export_begins function. features is by default at most 6 or 7 strings, so this shouldn't add much if any time to the export. I'll build up a PR for 1.4.2 to handle this.

PR #874, opened 2022-04-29
Exclude android builds from json/cfg fill

zaknafean:
Fix for the android export issue I was experiencing


zaknafean:
Fix for #841

Issue #848, opened 2022-03-25
Clicking on android don't work


hicchanwan:
Hi!
I ported my game with Dialogic to my android device. I created an instant dialog appearance in the scene. Choose event works well, but text events don't respond. In theory, by clicking on the screen, the following text should appear, but instead - nothing. Maybe this problem has already been discussed here? If you can tell me how to solve it, I will be grateful. 
Thanks

Android: 12
Godot: 3.4.3 stable

hicchanwan:
I solved this problem by adding a condition in the DialogNode file.tscn at the moment where the _input() function works. Where *if not Engine.is_editor_hints() and event.is_action_pressed(Dialogic.get_action_button()):* understands only mouse clicking on PC, but when importing to Android, mouse or click emulation does not work at all. I added a second condition for touchscreens *if not Engine.is_editor_hints() and InputEventScreenTouch:*. Everything is working well now. I hope this will help you fix similar problems

Jowan-Spooner:
Hi. I can confirm that the dialog doesn't respond to touch input as one would expect and because I think it would be a good thing to suport out of the box I will reopen this issue.

thebardsrc:
on my Android tablet, i found that its because the touchscreen isn't Device 0, which is what we have the default for mouse clicks set to. I had to change mine to All Devices Left Button insead of Device 0 Left Button

![image](https://user-images.githubusercontent.com/2214793/160294133-048d3de8-30da-4a2e-ac78-ee9089919672.png)

(this projects from before dialogic_default_action was created but that's where it would need to be updated)

zaknafean:
Adding this line to DialogNode

```if not Engine.is_editor_hint() and (event.is_action_pressed(Dialogic.get_action_button()) or event is InputEventScreenTouch):```
![image](https://user-images.githubusercontent.com/7741797/165870167-2e6c1abe-c7d9-4748-be3c-a5aec562fd9b.png)

Also just makes dialogic advance with any touch. Some solution is definitely needed, and we have alot of options.

PR #876, opened 2022-04-29
A basic touch to advance implementation

zaknafean:
Added a touch button overlay that covers the whole screen by default, but is resizable if you just want part of the screen to be touch sensitive.


zaknafean:
This PR adds some fun touch functionality for mobile devices that is currenlty lacking.

We start with a **TouchScreenButton** that is just a rectangle shape
image

In _ready() of DialogNode we set it to the Dialogic action button
image

and we change the currently not functioning 'Tap Screen to continue' with a new 'Tap area covers full screen'
image

If thats checked, the button extents will resize themselves automatically to match the reference rect size. Otherwise it'll leave the button alone. This will allow people to add 'touch dialog box directly to advance' or 'touch the blinker to advance' functionality to their mobile games if they so desire.

Regression test wise this doesn't interfere with regular mouse or keyboard controls.

Would love another set of eyes on this however, anyone affected by #848 in particular

coppolaemilio:
@zaknafean can you solve the conflicts on this one? otherwise I could give it a go

PR #887, opened 2022-05-02
Custom events android fix

zaknafean:


zaknafean:
Disabling the translation of the EventBlock due to it breaking with custom events on non-windows builds. This will be re-evaulated once Dialogic 2.0 is live.

PR #888, opened 2022-05-02
Translation preview, editing and browsing support.

Wyxaldir:
### Usage
To start, just click on the translation edit button next to text events (options not supported at this time)
image

You should then see a popup with four buttons, these are the meat of the functionality provided.
image

By clicking "Save at current location" changes made to a translation will be saved at the same location it loaded it from. It is greyed out in this case since we couldn't find the location (there is not key).

Click Save at.  .  . to save the key/value pair to a new or existing file. This will replace the old key if it is found or append it to the file if not. It is greyed out because there is no key to save.

Clicking browse keys will popup another window from which the user can browse the keys from any file within the selected folders (according to the new settings in the settings page)
image
image

Selecting a key will overwrite the content of the text box with the key.

Once a key is selected and if it can be loaded it appears under the text box like so
image

### The Glossary
This also supports editing for the glossary, as seen below. The buttons behave like they do above.
image

### Extra settings
This commit features two new settings: "in-editor translation editing" and "additional translations location". Disabling the first setting disables all of the new features from this commit, and the second is a folder from which we recursively search through to find extra translations for use with the new features.
image

### The problem

> Unfortunately, this method isn't without flaw, and I don't know how to fix it. When changing the .csv file in-editor the .translation files do not get regenerated. This means the UI becomes inaccurate until the next project refresh. While there are some workarounds I've implemented, it's not ideal. I've tried deleting the file and scanning, deleting the .import file and scanning, but nothing short of re-launching the project seems to work. If this can be fixed I'd consider this complete and pull-worthy (once I add support for options).

This has been fixed in the latest commit.


addons/dialogic/Editor/Events/Parts/Text/EditTranslationButton.gd
```diff
@@ -0,0 +1,63 @@
...
+signal value_saved(value)
+signal key_saved(value)
+
+export(bool) var handle_translation = false
+
+export(NodePath) var key_text
+export(NodePath) var translated_text
+
+
+func _on_EditTranslationButton_pressed() -> void:
```
AnidemDex, line 13: [`Button` has a `_pressed` virtual method](https://docs.godotengine.org/en/stable/classes/class_basebutton.html#class-basebutton-method-pressed) to know when it was pressed. Is called right before the node emmits the `pressed` signal.

addons/dialogic/Editor/Events/Parts/Text/TranslationKeyBrowserPopup.gd
```diff
@@ -0,0 +1,110 @@
...
+	emit_signal("key_selected", key)
+	hide()
+
+
+func _on_key_hovered(var content : String):
+	$HBoxContainer/KeyContents.text = content
+
+
+func _get_locale_index(var csv_line : PoolStringArray) -> int:
+	var editor_plugin = EditorPlugin.new()
```
AnidemDex, line 74: Editor plugin node is generated but never removed, this may cause leaked instances at exit and have a negative impact in editor

addons/dialogic/Editor/Events/Parts/Text/TranslationKeyBrowserPopup.gd
```diff
@@ -0,0 +1,110 @@
...
+	while file_name != "":
+		if dir.current_is_dir():
+			result.append_array(_get_translation_files(base_folder + "/" + file_name))
+			file_name = dir.get_next()
+			continue
+		if file_name.ends_with(".csv"):
+			result.append(base_folder + "/" + file_name)
+		file_name = dir.get_next()
+	
+	return result
```
AnidemDex, line 104: Are you sure you want to use all `*.csv` files in `base_folder`? Some users makes .csv files for storage that are not going to be used as traslation, or makes multiple `.csv` files for test translations.

Maybe you can use the translations added by the user in the project settings instead (`ProjectSettings.get_setting("locale/translations")`)? I don't know

AnidemDex (commented):
I'm not totally sure about the translation service file and the way it handles translations. I'm not against it, I'm just a bit worried about its impact in a long run

addons/dialogic/Editor/Events/Parts/Text/TranslationKeyBrowserPopup.gd
Wyxaldir: As far as I know since we're not adding the editor_plugin to the scene tree it shouldn't cause issues.

addons/dialogic/Editor/Events/Parts/Text/EditTranslationButton.gd
Wyxaldir: Ah, I didn't know about the _pressed method. Is there a benefit to using it instead of the signal in this case?

addons/dialogic/Editor/Events/Parts/Text/TranslationKeyBrowserPopup.gd
Wyxaldir: I understand this concern, at the very least I should include checks to make sure that the .csv's contain translation data, something along the lines of confirming that the top row only has valid locales maybe?

addons/dialogic/Editor/Events/Parts/Text/TranslationKeyBrowserPopup.gd
AnidemDex: `EditorPlugin` is `Node` type. It'll be an orphan node. In fact, the current status of the plugin already generates a lot of orphan noes and leaked resources (I haven't fount what commit introduced that issue)

addons/dialogic/Editor/Events/Parts/Text/EditTranslationButton.gd
AnidemDex: In this scenario is technically the same, so there's nothing to worry about. But if (in a hypothetical scenario) the scene file is modified and the signal is disconnected by accident, this function will never be called

addons/dialogic/Editor/Events/Parts/Text/TranslationKeyBrowserPopup.gd
Wyxaldir: I had no idea, that's really good to know, thanks! I'll go through and make sure to call queue_free() wherever needed.

addons/dialogic/Editor/Events/Parts/Text/TranslationKeyBrowserPopup.gd
Wyxaldir: We now check if the file is a valid .csv (see newest commit) so I'll mark this as resolved.

Wyxaldir:
Any chance this getting pulled in? If not what should be changed for that to happen?

coppolaemilio:
Hello Wyxaldir, sorry for not replying in such a long time!

Unfortunately this won't be merged since the 1.X branch is going to stay as is and we will continue working on the 2.0 branch. Feel free to get familiar with the new branch and see if you can contribute in that one instead since it is our top priority. If you want to keep working on the 1.X branch there are still some bugs that need fixing but are not high priority; you can check the milestone: https://github.com/coppolaemilio/dialogic/milestone/7

Thanks!

PR #890, opened 2022-05-03
Hotkey-fix

zaknafean:
None choice now exists and WORKS for hotkeys


zaknafean:
This will fix the Choice hotkeys being unselectable error

Issue #899, opened 2022-05-17
Name label disappears after question


pjalsGit:
## The problem

**Describe the bug**
For some reason, when I make a question to a character with a custom theme, the name label of the next character disappears.

**To Reproduce**
1. Make a question
2. Put a text event in the choice branch
3. Play timeline
4. See that name label disappears

**Expected behavior**
For the name label not to disappear.

**Screenshots**

https://user-images.githubusercontent.com/25231103/168756056-e5835b87-894f-4b98-99ea-0e9cf5fc0819.mp4

(yea i had some fun making the bug report)

Niko's custom theme has "Hide name labels" off.
Player's custom theme has "Hide name labels" on.

**System (please complete the following information):**
 - OS: Fedora Linux 35
 - Godot Version: 3.3.3.stable, tested also with 3.4.4.stable
 - Dialogic Version: [2efa1a0](https://github.com/coppolaemilio/dialogic/commit/2efa1a092143c10bff344539b904e84328f8cdc1)

## Solutions

**Workaround**

Disable name labels and act like it didn't happen.

**Possible fixes**

Not sure.

Yotopihi-Dev:
It's fading cause you didnt add a character join event for the player **_lol_**

zaknafean:
My hunch was right. It was the same root issue as #893. This is now fixed in the main branch.

Issue #893, opened 2022-05-13
Character name color 


coppolaemilio:
From Lucas Molina on Twitter:

Dialogic 1.4.1.

After a character that has a custom theme speaks, the color of the name of the next character that speaks changes to white for one line of dialogue, then changes back to the actual color it is supposed to be.

I'm sending you a gif of the bug. The name 'Waitress' should never appear as white, yet it does. The bug is reproducible 100% of the time for me.

https://user-images.githubusercontent.com/2206700/168375373-c298f30f-4e97-40bb-9631-16bca133c93c.mp4

zaknafean:
Can confirm this one, hits me all the time.

Yotopihi-Dev:
Not sure if this may be the issue, but in the theme editor theres an option for the character's color to be its own original version, so if the character has its own specific theme and is switching around in the timeline it may be overlapping, just an idea though.

Yotopihi-Dev:
Also, there's another thing. I saw somebodies character text change colors because the dialog called the name
![Screenshot_2022-05-15_150305](https://user-images.githubusercontent.com/101610280/168541356-95f236eb-c7fb-4697-8929-92b5c7385df4.png)
![Screenshot_2022-05-15_150257](https://user-images.githubusercontent.com/101610280/168541368-2dd6660a-a7f0-4cc1-bd5d-2a830819ad53.png)

zaknafean:
This is fixed in main as of #902

AD1337:
Thanks for fixing this @zaknafean!

PR #902, opened 2022-05-22
Fixed bug where name was updated before theme was loaded

zaknafean:
Moved where update_name occurs.


zaknafean:
This fixes #899 and #893 as they were the same problem in the end.

PR #905, opened 2022-05-26
Remove clip content to prevent outlining cut.

OmegaBowser:
If your font use an outline, each beginning row will cut this outline.
Checking "clip_content" to false fix this "bug".


zaknafean:
Confirmed, its a good change. Merging it in

PR #925, opened 2022-06-13
2.0 Character editor and (simple) Portrait management

Jowan-Spooner:
# Commit 1:
This commit adds a Wait event with a float input for seconds.

## Notes:
- Adds a float mode to the Number input
- Fixed a bug when the last event was an End Branch event.

# Commit 2
## Character editor
This re-implements the character editor from 1.x, but with a few changes. 
Scale, offset and mirror are now per-portrait, although I added a new `Main scale` setting that affects all portraits. The "real-size-preview" is a bit more advanced. It has an origin point now and actually displays the portrait offset. It's the new default view. I believe for everything except pixel-art games, this is the view you wan't to use.

There have been a couple of few changes to the timeline editor and toolbar to make them behave when the character editor is opened.

# Commit 3
## Portrait display nodes
There are now 2 more Dialogic Display Nodes:

- PortraitPosition scene is a Position2D node that only has a portrait_index (any integer  0 <= x <= 100) 
- PortraitHolder scene is a CanvasLayer that the portraits will be added to as children.

There should always only be at most one Portrait Holder (although more will not break) and there can be unlimited PortraitPositions, although they should ideally have different indexes. The DefaultDialogNode has 5 positions set up with the indexes 1 to 5

## DialogicGameHandler functions
- update_portrait() -> updates or adds a portrait
- remove_portrait() -> removes a character
- is_character_joined() -> true if the character has a visible portrait
- get_current_portrait_info_of_character() -> returns all the stored info on the portrait of the character (e.g portrait name, position, the actual node, etc.)

These functions mainly use the `current_portraits` array to communicate with each other/store stuff.

## Event updates
The character event and text event have been updated to use the methods listed above to manipulate the portraits. 

- Character Event has a position spin box. This is not optimal, especiall as writing a number in the text format is really obnoxious. My suggestion are somewhat named positions, either in the settings (e.g {1:'left', 2:'center_left', 3:'center', 4:'center_right', 5:'right'}) or as strings actually entered on the PortraitPosition nodes (then it would be hard to suggest them tho).

## Example Assets
I've added CC0 portraits of characters to test this. If this is to much, we can delete them again. They are the same that I used in the official Dialogic Test Project. 

## Missing from last commit :(
The character resource actually has a couple of needed updates for the last commit that I forgot to stage. These include new variables for nicknames, description, scale and theme.

# Images:
grafik
grafik
grafik
grafik
grafik


addons/dialogic/Editor/CharacterEditor.gd
```diff
@@ -1,7 +1,227 @@
...
+	find_parent('EditorView').edit_character(resource)
+
+
+func load_character(resource: DialogicCharacter) -> void:
+	if not resource:
+		return
+	current_character = resource
+	toolbar.get_node('CurrentResource').text = resource.resource_path
+	toolbar.set_character_mode()
+	get_node('%NameLineEdit').text = resource.name
```
coppolaemilio, line 27: we can use `$'%nodename'` instead of get_node() in this cases. I'll probably replace them after mergin

addons/dialogic/Resources/Events/event_wait.gd
```diff
@@ -0,0 +1,78 @@
+tool
+extends DialogicEvent
+
+
+# DEFINE ALL PROPERTIES OF THE EVENT
+var Time :float = 1.0
+
+func _execute() -> void:
+	var timer = Timer.new()
```
coppolaemilio, line 9: We should probably use the scene tree timer here instead of creating one https://docs.godotengine.org/en/stable/classes/class_scenetreetimer.html

PR #969, opened 2022-06-29
2.0 Allow to go back after a jump

Jowan-Spooner:
Basically this allows you to set a Jump event as a "checkpoint" and then at any time use another jump event to go to the last checkpoint (this could go multiple levels deep). 

This allows you to go to a different timeline and then return from where you came from.

The UI is honestly quite bad right now, but it seems to work well.

*Outdated images:*
grafik
grafik

### DIFF
--- a/addons/dialogic/Events/Jump/event.gd
+++ b/addons/dialogic/Events/Jump/event.gd
@@ -10,31 +10,26 @@ var Label : String = ""
 var Return:int = ReturnTypes.None
 
 func _execute() -> void:
-#	print("ret: ", Return)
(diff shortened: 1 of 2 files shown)


coppolaemilio:
I would update the language to call them checkpoints. So: "Checkpoint" and "Return to last checkpoint"

coppolaemilio:
Leaving the screenshot with the mockup here

Screen Shot 2022-07-07 at 21 17 10

coppolaemilio:
Unfortunately it is outdated.

Issue #915, opened 2022-06-03
Theme setting fails to load if canvas layer is changed


zaknafean:
## The problem

**Describe the bug**
![image](https://user-images.githubusercontent.com/7741797/171764452-b04c7aa5-9eb0-4f5c-a2fe-2214fb1a3c21.png)
If Canvas layer setting is changed, but Default Theme is not, then an error is thrown here:

![image](https://user-images.githubusercontent.com/7741797/171764593-66003fd2-0bc2-4467-83b6-35b452d05122.png)

**System (please complete the following information):**
 - OS: Windows 10
 - Godot Version: 3.4.4
 - Dialogic Version: 1.4.2

## Solutions

**Workaround**
Quickest work around is to just change the theme then change it back so an entry is written in the settings.cfg

**Possible fixes**
The fix is to add 
```if settings.has_section_key('theme', 'default'):```
to line 148 or 149 of DialogNode.gd

Don't have time to fix this now, so just writing this all for now. This may be the root cause of some issues other people are seeing.

zaknafean:
This and its dupes should be fixed with #978

PR #978, opened 2022-07-05
1.4.3 - Fix for export error due to settings theme failure

zaknafean:
This fix changes the logic for loading a theme file to give and use a fallback even if nothing exists in the settings file. Previously this value COULD be a null given certain circumstances during an export.


zaknafean:
This should fix #915 and the many duplicated related to it.

coppolaemilio:
thank you so much!

Issue #904, opened 2022-05-25
timeline_end signal returns timeline filename instead of timeline name


IsThisTheFuture:
## The problem

**Describe the bug**
I have multiple dialogs in a scene and after each dialog ends I use an 'after_dialog' function that gets triggered by the 'timeline_end' signal to check for the timelines name that was just played and do stuff accordingly.
When I start a dialog timeline and connect the "timeline_end" signal as shown below, the signal returns the timelines filename (e.g. 'timeline-1653476470.json') as a parameter instead of the actual name I have given it in the Dialogic Editor (in this case: 'shopkeeper-convo'). 
This makes the code unreadable since I have no idea what conversation 'timeline-1653476470.json' actually is

**To Reproduce**

```gdscript
func _on_dialog_area_body_entered(body):
	var new_dialog = Dialogic.start("shopkeeper-convo") 
	new_dialog.connect("timeline_end", self, "after_dialog")
	add_child(new_dialog)

func after_dialog(timeline_name):
	if timeline_name == "shopkeeper-convo":
		# Does nothing, because timeline_name is the filename
	elif timeline_name == "timeline-1653476470.json":
		# This is how I have to do it right now
```

**Expected behavior**
I would expect Dialogic to return the timeline name (e.g. "shopkeeper-convo") via the 'timeline_end' signal instead of the filename.

**System (please complete the following information):**
 - OS: Linux (Arch Linux, kernel 5.17.9)
 - Godot Version: 3.4.4, also 3.5RC2
 - Dialogic Version: 1.4.1

Is this expected behavior? It just seems so unintuitive to use the filename instead of the conversations name, which the user would actually be able to interpret.
However, if this works as intended, how should I go about it instead? Should I set the 'dialogic_signal' in the timeline editor and add the timeline name as a parameter?

zaknafean:
Fixed with #979. Timeline end and start both push the human readable timeline path/name that Started the dialog. If you change timelines it won't return the name of the new one.

PR #979, opened 2022-07-05
1.4.2 - Fix for #904

zaknafean:
We track the human readable timeline_name and return it during emitted signals


zaknafean:
Please note this is KIND of a breaking merge, as rather anyone who was doing comparisons with the randomly generated ID, will now have to do it with the human readable name/path-name

PR #984, opened 2022-07-07
2.0 Variables improvement (+ Settingeditor improvements)

Jowan-Spooner:
# Dialogic variables
You can now create and manage dialogic variables from the settings. They can be grouped in a folder structure.

Variables are managed entirely by the Variables Subsystem.

You can reference them in texts like {MyVariable} or {MyFolder.MySubfolder.MyVariable}.

You can set them with the Set Variable event.

You can get and set them by writing Dialogic.VAR.MyVariable = "Hello" or Dialogic.VAR.MyFolder.MySubgolder.MyVariable += 10

# Variable editor
The variable editor in the settings supports drag and drop for folders and  variables.

# Subsystems can have settings editors
In the `get_required_subsystems()` method of events you can now add a third element to each subsystem specifying the settings editor scene (a path). This scene is added as a tab to the settings editor.

The variables editor is added as such a subsystem editor for example.

pros:
- makes the settings editor cleaner
- makes adding custom settings for "dialogic plugins" easier


coppolaemilio:
I would avoid using the terms folders for this, since the term is never used this way. Something like `category` or `section` should be a more fitting term. You wouldn't call the inspector sections _folders_, so this should be similar to that.

PR #986, opened 2022-07-08
1.4.x - Added NoSkip Event and Auto Advance Mode

zaknafean:
Added a no Skip event. This blocks default action input from working until an unchecked no skip event is reached. This is most useful for combining Dialogic with cut scenes and the [nw] command in order to avoid messed up timings.

Only advancement of dialog is skipped. Choices must still be made and will still accept mouse clicks/taps

image

### DIFF
--- a/addons/dialogic/Editor/Events/NoSkipEvent.tscn
+++ b/addons/dialogic/Editor/Events/NoSkipEvent.tscn
@@ -1,0 +1,82 @@
+[gd_scene load_steps=7 format=2]
+
+[ext_resource path="res://addons/dialogic/Editor/Events/Parts/NoSkip/NoSkipPart.tscn" type="PackedScene" id=2]
+[ext_resource path="res://addons/dialogic/Images/Event Icons/Main Icons/close-dialog.svg" type="Texture" id=3]
+[ext_resource path="res://addons/dialogic/Editor/Events/Templates/EventTemplate.tscn" type="PackedScene" id=4]
+
+[sub_resource type="Image" id=4]
+data = {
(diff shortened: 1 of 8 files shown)


coppolaemilio:
Do you need to manually add the "[nw]" keyword when this is enabled?

zaknafean:
Yes, it doesn't auto, auto advance text for you. Do you think it should?

coppolaemilio:
It should at least say that you need to do that. I'm not too worried about it but I'm sure it will cause new comers some confusion on "why is the text stuck".

zaknafean:
Ok, making a couple changes to support that and get like 90% of the way to 'auto advance mode'. Maybe I'll go all the way.

zaknafean:
- No Skip event now automatically auto advances your dialog, as if each text event has a **[nw]** tag.
- You can change the default skip time for all text events by adjusting the wait time in the no skip as shown below. If you set it to '5' it would be the equivalent of giving every text event the **[nw=5]** tag
image

- Auto Advance mode now exists. It exists purely as an api function. Turn on Auto Advance mode like this:
`Dialogic.auto_advance_on(true, 4)`
Where **Argument 1** is true (auto advance on), or false (auto advance off) and **Argument 2** is the default seconds to wait after all text is displayed
- Auto Advance mode is automatically cancelled if any 'valid' input is received from the 'default action input' of dialogic. This includes during choice questions if applicable

1. No Skip and Auto advance share the same logic except where noted above
2. If your text event contains a **[nw]** tag, the tag takes precedence over any other wait settings
3. If vocals are enabled, No Skip and Auto advance wait until the vocal time remaining is completed to auto advance. This is the equivalent of the **[nw=v]** tag functionally speaking
4. Questions must still be answered manually


Design wise the No Skip event mode is meant to be used in cut scenes or other situations where timing is critical but you still want things to be displayed with Dialogic.

Auto Advance mode is when you want to allow people to enjoy your story without interactions or hands on the device at all.

KvaGram:
Tested the new feature with a short dialog. Made a video.
https://youtu.be/wWUgHqG1mfk
Timeline:
[timeline-1657463632.json.zip](https://github.com/coppolaemilio/dialogic/files/9079346/timeline-1657463632.json.zip)

Frogsblood:
Is there a way to use NoSkip without auto advancing the dialog? It would be very useful for making sure that key dialog isn't missed by accident.

zaknafean:
I'm not 100% sure how you could do that. If you aren't auto advancing, then how are you moving the dialog forward if you can't skip it either? What do you have in mind?

Frogsblood:
Dialogic lets you skip dialog while its typing out. I want it so you can only move the the next dialog after its finished.
Maybe I'm mistaken on how this works?

PR #995, opened 2022-07-10
1.4.x Translation Service Speed Fixes

thebardsrc:
**Problem**
The translation service for the plugin, DTS, is slow, and significantly slows down the loading of a project, as well as the loading time of a timeline. 

**Analysis and Cause**
The reason for it comes down to the way that DTS was designed. It's a drop-in plugin for translations for editor plugins for Godot, so it was written to be as easy to use as possible. But that also causes a slowdown. To be able to be slot-in, it was designed as a static class. By design, Godot doesn't have any static variables, and so it was written to load the translations from file every time it is called.

The problem is that's a _lot_ of loading. Particularly with how Godot uses the plugin. While not everything is in the translations currently, _every_ UI item of Dialogic is created with DTS use in mind, so everything tries to load a translation. And there are so many elements in Dialogic. Just on loading a game with the plugin, without even clicking on the Dialogic plugin to do anything, Dialogic makes **275** requests to DTS just to start Godot itself. And every time you click on something in Dialogic it loads all the translations for what's on that page. And whenever you load a timeline, it also loads it for every*(with caveat that I will touch on later) single item in the timeline as well. All of these loads really slow down everything, from initial project load, to editing timelines - especially longer ones. 

For the most part, I don't think anyone really noticed this before now, and just accepted the slowness. I only noticed this myself because I had verbose debugging turned on for fixing something unrelate to Dialogic, and saw in my debug log all the large numbers of loading the translation files over and over.

**Workaround**
As a workaround that was suggested on the Discord, one way to solve this when the translation service isn't actually needed would be to make it just return the text without fetching a translation. I was originally going to just submit a patch for this, with a toggle in the Settings, but that also meant it could only be usable with a speed improvement in English only. But I decided instead to approach it a different way instead.

**Properly Fix**
So how do we solve the problem of a static class not being able to keep a static variable to keep the translations anymore? Well, by making the class non-static anymore. Which meant it needs an instance. 

The root of the Dialogic editor plugin is the `EditorView`. Everything is a child of that. So that's where our new class variable, `dialogicTranslator`, exists. Some of the components already were using it for other purposes, using a variable standard of `editor_reference` which the `_ready()` function of some of them already defined as the EditorView through a find parent. So with this change, that's how every call to DTS is now changed. Instead of `DTS.translate()`, now they are all changed to `editor_reference.dialogicTranslator.translate()` calls. This PR has all of the existing ones converted, but any new ones added 1.4.x will have to be changed to the new method too, as DTS can no longer be called directly as it is no longer a static class.

The proof is in the pudding. This is a load time comparison with my largest timeline, which has over 400 elements, between the old static version, and the new non-static version. It also shows that the new method is the same performance as just the bypass workaround, so that isn't even needed anymore:
unknown

I wasn't sure how to get millisecond load times for starting a Godot project, but with the previously mentioned 275 calls to DTS just to start a project, it would seem to have a pretty similar speedup.

**Complication**
There is one caveat to this, and that's a different issue that DTS caused: exports. Namely, the problem in #887 had disabled DTS on the Timeline editor, because for some weird reason it was having problem on exports of projects with custom events on certain platforms. And it doesn't make sense at all, because this shouldn't do that, it's only really used by the editor. Since this also turns that back on, I have no idea what this will do in this case. Maybe the change from a static class may fix it by itself, or may bring that one back. It will need other testing from the people that were experiencing that issue, and if it does then some samples of the custom events and the setups they're using to try and find the real cause of those issues, rather than just that workaround from before. 

**Moving Forward**
2.0 hasn't implemented DTS yet at this stage. With the fundamental change to using resources, it may not end up being used to the same extent in there. But for any use of it, I would suggest implementing it the same way as my change here, for the same reason. Loading the translations only once on project load, instead of on every single call for a translation, significantly improves the speed.


zaknafean:
I was able to create #887 at will, so will regression test this later tonight

thebardsrc:
Since it really shouldn't be calling DTS on custom events to begin with, since they're not gonna be in it anyway, I've added a check now to just keep it from being called on them altogether

zaknafean:
Gets my thumbs up. I haven't been able to break it, and the exports look great.

PR #1015, opened 2022-07-14
Version 2.0 Audio sounds over text

M7mdKady14:
This is my first time making a pull request so excuse me if I did something wrong

what is available:

1. a new display node that can make sounds the text gets revealed
2. it can choose one sound from a random number of sounds you supply
3. you can add a random variance to the pitch and volume of the sounds
4. you can choose what number of characters to show before making the sounds (making sounds every character, every other character ,etc.)
5. You can choose what characters not to trigger the sounds (for example the "space" or the '"ull stop  point" so someting like this won't make any sound except for the first three characters "hmm . . . . . . . . ")
6. all of it is customizable from the godot editor

things that can be improved:

1. the namings sucks, I have no sense in naming anything
2. a better way to get the dialog richtext node
3.  I don't know how events work so I didn't do an event, but if someone can, you can make an event changing the current audio sound node (don't know what to call it) so that it produces different sounds


Jowan-Spooner:
First of all, thank you very much for your work!

However here are a few thoughts of mine:
The connection between the DialogicDisplay_TypeSounds and DialogicDisplay_DialogText is not good currently, as it doesn't allow for multiple Text nodes which are crucial for multiple scenes. I would suggest that you have to add the TypeSound as the child of the DialogText node (then we can use get_parent()).

Also one important thing that is missing is how to overwrite that audio for certain characters. This brings other issues though, so I would merge this and add those other things later.

Also I will have to move the TypeSound class into the Dialogic/Events/Text folder as it is very specific to that event.

Going to test now!

Jowan-Spooner:
Alright, going to merge this now! Thanks very much.

Jowan-Spooner:
Also if you are on discord feel invited to come over into #dialogic_dev or #Dialogic 2.0 on emilios server!

M7mdKady14:
> Also one important thing that is missing is how to overwrite that audio for certain characters. This brings other issues though, so I would merge this and add those other things later.

My idea wasn't for only one TypeSound to be present. It was to add multiple of them with different names like John_sad, Mark_angry or things like this and then add the event to change which one to use according to names.

or better yet, integrate it into the character file so you can assign a character multiple voice tones and then choose which one to use in the text event

Issue #962, opened 2022-06-28
2.X Reimplement SFX


Jowan-Spooner:
- [ ] Choice hover sound
- [ ] Choice click sound
- [ ] Typing/Speaking sound
- [ ] Text finished revealing sound

- [ ] Possibly sounds for other events (see #384)

Jowan-Spooner:
In theory these sounds could be implemented on the display nodes if choice and text, but you wouldn't want to set the choice hover sound on each button. Also the typing sounds could depend on the character.

PR #1017, opened 2022-07-14
2.0 add the sound that plays when text is revealed

M7mdKady14:
Go to the child of the dialog richtext and add the sound you want in "end_sound"


M7mdKady14:
#962 
sound after text is done
so is the button ones and added more for focus, unhover and released

addons/dialogic/Events/Choice/Display_ChoiceButton.gd
```diff
@@ -4,6 +4,21 @@ class_name DialogicDisplay_ChoiceButton, 'icon.png'
...
+
+onready var sound_node = get_parent().get_node("DialogicDisplay_ButtonSound")
+
 func _ready():
+	connect('pressed', sound_node, '_on_pressed', [sound_pressed])
+	connect('button_up', sound_node, '_on_released', [sound_released])
+	connect('mouse_entered', sound_node, '_on_hover', [sound_hover])
+	connect('mouse_exited', sound_node, '_on_unhover', [sound_unhover])
+	connect('focus_entered', sound_node, '_on_focus', [sound_focus])
+	connect('focus_exited', sound_node, '_on_unfocus', [sound_unfocus])
```
Jowan-Spooner, line 22: Maybe we could do this in the ChoiceSoundScript? Like
```
for child in get_parent().get_children():
    if child is DialogicDisplay_ChoiceButton:
        child.connect('pressed', self, '_on_pressed', child.sound_pressed)
        #...
```
Feels like it's cleaner if it's in the ButtonSound script, but it's not very important.
We should definitely use the class names to connect these nodes tho, because fixed names are not a good idea imo.

addons/dialogic/Events/Choice/SoundButton.gd
```diff
@@ -0,0 +1,60 @@
+extends AudioStreamPlayer
+
+class_name DialogicDisplay_ButtonSound
```
Jowan-Spooner, line 3: This file should be named Display_ButtonSound or even Display_ChoiceSound (maybe the class name should contain Choice as well)

addons/dialogic/Events/Text/DialogicDisplay_TypeSounds.gd
```diff
@@ -78,9 +82,16 @@ func _on_continued_revealing_text(new_character) -> void:
 	sound_finished = false
 
 
-func _on_Sound_finished():
+func _on_Sound_finished() -> void:
 	sound_finished = true
```
Jowan-Spooner, line 86: Is this actually used anywhere? Like is anything connected to this?

addons/dialogic/Events/Choice/SoundButton.gd
M7mdKady14: why remove the dialogic at the start?

PR #1032, opened 2022-07-23
2.0 Set Variable Improvements

### LINKED ISSUE #1030: 2.x Improve Set Variable event
The set variable event should have a random number option and multiple operators (+=,-=, *=, /=, ==).

Jowan-Spooner:
# Set Variable
- now has a operator selection (allows "to be", "to itself plus", "to itself minus", "to itself multiplied by" and "to itself divided by")
- now has suggestions for built-in dialogic variables (both on the variable and value picker)
- now has options to us a random int and set max and min values for it

The event is now also saved in a special format like this:
```
VAR PlayerName = Jonny
VAR MyAutoload.health += 20
VAR Player.Stats.Strength -= [random="True" min="10" max="20] # if you set a value AND random, the value will be ignored
```

# Complex Picker
I renamed the ResourcePicker to ComplexPicker as it is not used for resources only, but for pretty much anything you want. I've also made small improvements to other input fields:

- hide empty labels on OptionSelector
- make OptionSelector Popup Panel match other menu panels
- make default size for SingleLineText smaller
- hide empty labels on SingleLineText
- add condition to header label

- should fix #1030 and #1033


coppolaemilio:
If no variables exists it should show a button to open the settings menu or the documentation page
Screen Shot 2022-07-23 at 11 04 10

Jowan-Spooner:
Same with characters tho. Would probably do that on a seperate Issue and PR if that's okay.

coppolaemilio:
yeah, a separate PR works for me

PR #1035, opened 2022-07-23
2.0 Editor Improvements

Jowan-Spooner:
# ComplexPicker
Has been changed once more.
The suggestion function now expects more complex dictionaries. This allows to give more information, like tooltip, icon, etc.

This has been implemented on many events. Improved pickers can be found on:
- Text
- Character
- Jump
- Variable

# Expand Event
Events can be expanded and collapsed with the tool icon.
Most events will hide their body by default (all except Text currently).
Double-clicking the event also expands/collapses the event.

# Last Resources Picker
The toolbar now has a ComplexPicker for selecting a recently used resource. Also changed the startup load behaviour to use that list and thus allow loading characters as well.

# Unsaved Changes Warning
... when changing resource and having unsaved changes.

### DIFF
--- a/addons/dialogic/Editor/CharacterEditor/CharacterEditor.gd
+++ b/addons/dialogic/Editor/CharacterEditor/CharacterEditor.gd
@@ -27,8 +27,7 @@ func load_character(resource: DialogicCharacter) -> void:
 	if not resource:
 		return
 	current_character = resource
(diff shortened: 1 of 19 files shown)


coppolaemilio:
If the text is too long it adds horizontal scroll:
image

PR #1041, opened 2022-07-27
Implement voice support in Dialogic 2.0

KvaGram:
currently a draft.
Unlike in Dialogic 1.0, voice support is not integrated into the text event. Instead it is a separate event that the Text event can detect and pull data from*. 
Another change from 1.0, is that only the audio region is required desperately for each line. The audio file and settings are shared. If a user requires a line to be different, for whatever reason, they can split the text event, and add a new voice event.
This choice was made to reduce the clutter of most often identical settings (personal experience).

_* ~~Exactly how this is done is still on the todo-list.~~ This is done in a simular way to the Choice event. Except it is placed above the text event, in contrast to Choice events which are placed below._


KvaGram:
Well. The voice support is ready for review... except now I need to go ahead with starting to port this to Godot 4.
This will be done in a separate branch, for a separate pull request.
Leave it to you on whatever to merge or close this. I am not closing it myself.

Issue #1029, opened 2022-07-21
Improve Wait Event


Jowan-Spooner:
The wait event in 1.x had a hide and skippable option. 
- [x] hide
- [ ] skippable

Issue #956, opened 2022-06-28
Add settings to character events


Jowan-Spooner:
Connected to #948
Main missing ones are:
- [x] Animation
- [x] Animation length
- [x] Z-index
- [x] Wait for animation
- [x] Mirror
- [x] Allow 'All Characters' in Leave mode

Jowan-Spooner:
- Most of these are covered in #998

PR #1055, opened 2022-07-30
Dialogic 1.x to 2 Converter

thebardsrc:
**Dialogic 1.x to 2.0 Converter**

I'm going to make this as a Draft PR, so as I am working on it I can collect all the various issues and complications I run into, for where the Dialogic 2.0 implementation has shortcomings and is missing features from 1.x


thebardsrc:
Before I start on the timeline converter, I already see that there's going to be a problem with Characters in the timelines:

With how they are currently, they do not include folder path, and every one has to be unique. This creates a situation in which if 1.x had characters that were the same name in different folders will have collisions now, and may not use the expected character. In this example, it is using the "sub/Dex" character rahter than the root "Dex"

image

As the Timeline editor is also currently lacking the character colors, as 1.x had, you also can't tell at a glance what one is being used, either. 

image


Other Godot 4 conversion note on this issue: on 3.5 you could drag a character resource into the timeline character slot to select it directly, but in 4.0 as of this writing you cannot

Jowan-Spooner:
> Other Godot 4 conversion note on this issue: on 3.5 you could drag a character resource into the timeline character slot to select it directly, but in 4.0 as of this writing you cannot

Will make a fix for that!

thebardsrc:
Note to self for later: since variables are directly stored in the project settings now, it actually doesn't make sense for the file picker to allow a path outside of res:// for the other converted files to be created

thebardsrc:
so I'm not exactly sure that I actually _can_ update the Project settings from running scene. ive not really looked into it, but that seems a not-good design to allow a running game to change the project like that? so for now I'm gonna export the variables - and eventually other theme and global settings that port over - into another file, that it can be imported in later with a separate import function

thebardsrc:
in order to add it to the Settings editor panel, the only way I could figure out was registering an "Event" for it. which also meant I had to move it into the Events folder.  if there's another way to do it without that, please let me know, I would prefer it in its own separate folder so it's easier to find to remove when finished, like it initially was

thebardsrc:
1.x things not yet implemented in 2.0
- a bunch of the Character Update options 
- conditions on choices
- the entire If function altogether
- variable set value doesnt seem to really work at all, but is certianly missing random
- wait evnts in 1.x you could set whether to hide the dialog or not during them
- change scene 
- call node only has a path currently

thebardsrc:
thanks to @Jowan-Spooner , turns out instead issue is there's a horizontal scroll for me in Godot4 and i didnt even see the options were there on all of those things to see more things in the timeline! so probably all of that previous list can be ignored

Jowan-Spooner:
Not all. 
- There are a couple of character update things missing indeed (#956)
- change scene isn't a thing and I doubt it will be. What you can do however is to implement it as a call node event
- wait events have no settings (#1029)

PR #1067, opened 2022-08-01
Dialogic 1.x to 2 converter

thebardsrc:
Reopen of #1055 to continue work, because apparently merging the upstream branch force closed the original PR


thebardsrc:
for converting variable names, I'm not actually sure where else they could be used outside of Text events other than Conditionals and Choices... I know Character Name, Display Name, Description, I don't know if Nicknames can, anything else?

thebardsrc:
everything that's been implemented so far in 2.0 to this point is in now, ready to merge!

thebardsrc:
I don't like that prefix idea at all as an alternative to being able to properly use folders, that's just a mess. but i needed something for the interim until we can work that out better

thebardsrc:
all 1.x timeline parameters now implemented, with several of them going unused currently on some scripts where their functionality hasn't been used yet. But now all timelines and characters can be converted with all of their options, and so the converter itself is done

PR #1068, opened 2022-08-01
Make serial audio region width-greedy

KvaGram:
this makes it compete with the audiobus field in Voice event, and saves a little bit of space in the UI.


Jowan-Spooner:
I will merge that even though it's only a temporary fix, as some way of moving it to a new line needs to be implemented for other events too.

PR #1076, opened 2022-08-03
2.0 Settings Window Fix

thebardsrc:
Sets the main popup size to match the size of the panel that the General tab was designed for. Also switches the order of the commands for the show_settings() command, which makes it actually change the size of the popup properly


coppolaemilio:
Thanks!

PR #1088, opened 2022-08-04
2.0 Script Export

thebardsrc:
Just this file was throwing a script error because it happened to be still in 3.x export format and not 4.0


coppolaemilio:
Thanks! Not sure how we missed this one!

PR #1090, opened 2022-08-05
2.0 Various Speedup Improvemnets

thebardsrc:
I'll keep documentation about each one in here as I go along, but you can merge it in the morning and I'll open a new branch for more later

### DIFF
--- a/addons/dialogic/Resources/TimelineResourceLoader.gd
+++ b/addons/dialogic/Resources/TimelineResourceLoader.gd
@@ -66,6 +66,12 @@ func _load(path: String, original_path: String, use_sub_threads: bool, cache_mod
 		var event_content = line_stripped
 		var event = DialogicUtil.get_event_by_string(event_content).new()
 		
+		# While in editor, we want to let this process all ahead to build the editor, but not at runtime
+		if Engine.is_editor_hint():
+			event.set_meta("delayed_process", false)
+		else:
+			event.set_meta("delayed_process", true)
+		
 		# add the following lines until the event says it's full there is an empty line or the indent changes
 		while !event.is_string_full_event(event_content):
 			idx += 1
@@ -82,7 +88,11 @@ func _load(path: String, original_path: String, use_sub_threads: bool, cache_mod
 			event_content += "\n"+following_line_stripped
 		
 		event_content = event_content.replace("\n"+indent, "\n")
-		event._load_from_string(event_content)
+		if event.get_meta("delayed_process") == false:
+			event._load_from_string(event_content)
+		else:
+			#hold it for later if we're not processing it right now
+			event.set_meta("event_content", event_content)
 		events.append(event)
 		prev_was_opener = event.can_contain_events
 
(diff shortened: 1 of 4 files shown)


thebardsrc:
first notable fix: as Dialogic is initializing and loading the subsystems, and it gets to the Text subsystem, it was loading every character at that step to build the character names and a regex to replace them. There's a check *later* to see if you're actually using, but not at the beginning. 

I'm not using that myself on this project. This change cut my game's initialization time (loading jsut a blank scene with a button and pressing that button then loads at timeline) from ~3000 ms, to ~1500 ms,

thebardsrc:
main slowdowns is in the actual timeline loader. each line of the timeline takes ~80 ms to go through each pass of the "while idx < len(lines)-1:" loop, which adds up quickly, with the timeline we're basing this on having 110 lines, adds up to a little over 8 seconds of load time, front loaded at the beginning

characters are loaded as it reads the timeline, if that line has a character in it. character loading is only adding about 10 ms, so _that_ is happening pretty quickly.  

rest of the routines break down like this, based on time outputs for each further loop:
image
Line 62 -> Line 69: ~5 ms
Line 69 -> Line 73: ~25 ms
LIne 73 -> Line 88: ~5 ms 
Line 88 -> Line 62 of next loop: ~45 ms

the actual DialogicUtil.get_event_by_string() loader is taking some of the time, but its not the biggest part of the time, that's those last four lines. They break down as follows:

event_content.replace(): 2ms 
event._load_from_string(): all the rest of this time 
events.append(): 3ms
prev_was_opener = event.can_contain_events(): 3ms

it's late now and I need to sleep and my ability to comprehend now is waning, so I can't go further into it.  but looking at  DialogicUtil.get_event_by_string() it's constantly reloading files every single loop, which really should be loaded only once and referenced from there (since all of DialogicUtil is a static class file, for reference this was the same problem DTS had in 1.x to cause it's extensive slowdown that I had to rewrite it to be not a static class for). and the actual event processing in event._load_from_string() should be held off to runtime instead, running that with each new line as its being processed rather than runnig that all at the beginning first

thebardsrc:
this could probly overall be rewritten to be better, but this deferral takes care of = a big chunk of the speed problem.

addons/dialogic/Resources/TimelineResourceLoader.gd
```diff
@@ -82,7 +88,19 @@ func _load(path: String, original_path: String, use_sub_threads: bool, cache_mod
...
 		event_content = event_content.replace("\n"+indent, "\n")
-		event._load_from_string(event_content)
+		
+
+		
+		if event.get_meta("delayed_process") == false:
+			event._load_from_string(event_content)
+		else:
+			# a few types have exceptions with how they're currently written
+			if (event['event_name'] == "Label") || (event['event_name'] == "Choice") || (event['event_name'] == "EndBranch"):
```
Jowan-Spooner, line 98: No need to check for the EndBranch event it doesn't load anything anyways.

coppolaemilio:
Do you want me to merge this one?

thebardsrc:
you can merge it as is, i'll make a new branch and PR for any more changes from here

PR #1096, opened 2022-08-06
More runtime speed improvements

### LINKED ISSUE #1091: 2.0 Slow Timeline Load At Runtime
findings of the slowdowns and where I left off on the first speedup PR:

main slowdowns is in the actual timeline loader. each line of the timeline takes ~80 ms to go through each pass of the "while idx < len(lines)-1:" loop, which adds up quickly, with the timeline we're basing this on having 110 lines, adds up to a little over 8 seconds of load time, front loaded at the beginning

characters are loaded as it reads the timeline, if that line has a character in it. character loading is only adding about 10 ms, so _that_ is happening pretty quickly.  

rest of the routines break down like this, based on time outputs for each further loop:
image
Line 62 -> Line 69: ~5 ms
Line 69 -> Line 73: ~25 ms
LIne 73 -> Line 88: ~5 ms 
Line 88 -> Line 62 of next loop: ~45 ms

the actual DialogicUtil.get_event_by_string() loader is taking some of the time, but its not the biggest part of the time, that's those last four lines. They break down as follows:

event_content.replace(): 2ms 
event._load_from_string(): all the rest of this time 
events.append(): 3ms
prev_was_opener = event.can_contain_events(): 3ms

it's late now and I need to sleep and my ability to comprehend now is waning, so I can't go further into it.  but looking at  DialogicUtil.get_event_by_string() it's constantly reloading files every single loop, which really should be loaded only once and referenced from there (since all of DialogicUtil is a static class file, for reference this was the same problem DTS had in 1.x to cause it's extensive slowdown that I had to rewrite it to be not a static class for). and the actual event processing in event._load_from_string() should be held off to runtime instead, running that with each new line as its being processed rather than runnig that all at the beginning first

### LINKED ISSUE #1093: 2.0 Slow Load At Project Open
This is very different - but connected to - #1091. 

These two settings are saved in the Project Settings, to be able to reload the last opened timeline: 
image

Because of those in my project settings, and because loading timelines recursively crawls _all_ of the timelines it touches, just restarting this evening because of a crash with one of my timelines from my VN project, which are all interconnected to each other with a loop back to the very beginning at the end, meant that it had to touch _every_ timeline in my project. Godot editor will not allow anything else to finish loading _until_ that is done, so I had to wait about 10 minutes or so for it to finish that process, because Godot would not populate the rest of the "res://" tree until that was completed, so I could not do anything else with it

thebardsrc:
Finishes the creation of the threads started on #1090, so now it's a continuous running queue that Dialogic can continue feeding into. Ideally now, the only time there should be any slowdown for handling of events during runtime will be if the timeline jumps to a new timeline that wasn't processed by the queue yet.  Overall fixes #1091 

Also completely removes the recursive crawl of timelines as it causes so many issues. This required changing the main `DialogicEvent.load_from_string_to_store()` function to no longer load .dtl files, permanently. This changes `DialogicJumpEvent `to override the setter on the `Timeline `so it stores the path into a new variable, which it then uses for the new `load_timeline()` function when finally ready to load the actual `DialogicTimeline `into the actual `Timeline `variable.  Should fix #1093 as well, but I don't want to merge this in yet until I can test to make sure it doesn't have side effects in the editor, which I haven't checked yet, but I need to leave for a while so can't check until later tonight. Beyond this, the Editor will need something completely different, separate arrays to iterate through or something, for keeping the timeline and jump suggestions


thebardsrc:
there's also a weird rare problem building the new event on trying to run it if it hadn't been processed yet, where it would fail at loading one of the event_scripts. I couldn't recreate it, but I want to figure that out too before I merge it 
image

thebardsrc:
alright with the change, Timeline Editor just shows all the jumps as "this timeline"
image

thebardsrc:
that commit fixes the timeline editor for now so it's usable, although doesn't look good. 

pushing this now because the recursive crawl of timelines was a critical bug that locks up both game and Godot editor

PR #1099, opened 2022-08-07
2.0 Additional Background Options 

thebardsrc:
Adds the background crossfade option back from 1.x. It's skipped on loading custom scenes as a background, since they should be let to do their own thing, but maybe sending them a signal with the fade time if they want to use it would be a good idea.

Also I will add the hide dialog option back as well, but not in yet as of this comment. Skippable was also a property on them as well, ive not looked at how much of the skipping has been implemented yet tho


thebardsrc:
wait hide dialog wasn't on Background initially, it's on a wait. Skippable wasn't before either, but it should maybe be implicitly skippable and skip the wait time

PR #1105, opened 2022-08-09
Fix Event Lines if Timeline ends with Question

Jowan-Spooner:
- also add filter argument to suggestions method of the Toolbar File Picker

### DIFF
--- a/addons/dialogic/Editor/Common/Toolbar.gd
+++ b/addons/dialogic/Editor/Common/Toolbar.gd
@@ -63,7 +63,7 @@ func _on_AddCharacter_pressed() -> void:
 	)
 
 
-func suggest_resources() -> Dictionary:
+func suggest_resources(filter) -> Dictionary:
 	var suggestions = {}
 	for i in DialogicUtil.get_project_setting('dialogic/editor/last_resources', []):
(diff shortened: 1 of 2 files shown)


addons/dialogic/Editor/TimelineEditor/TimelineArea.gd
```diff
@@ -106,16 +106,16 @@ func _draw():
 						if $Timeline.get_child(sub_idx).current_indent_level == event.current_indent_level:
 							end_node = $Timeline.get_child(sub_idx-1)
 							break
-				
+				var rect_size = Vector2()
```
coppolaemilio, line 109: This can be statically typed

addons/dialogic/Editor/Common/Toolbar.gd
```diff
@@ -63,7 +63,7 @@ func _on_AddCharacter_pressed() -> void:
 	)
 
 
-func suggest_resources() -> Dictionary:
+func suggest_resources(filter) -> Dictionary:
```
coppolaemilio, line 66: Does the filter have a particular type? It's okay if it is dynamic, just want to make sure

PR #1107, opened 2022-08-09
2.0 Resource Loader Further Speedups

thebardsrc:
Sometimes - not all the time - when starting Godot 4 current alpha, it seems to lose it's resource cache, and needs to reload everything. When this happens the Dialogic resource loader blocks Godot's UI from finishing indexing the rest of the project, and it sits there with an empty directory until it's done. 

Initially, before other speed improvement I had put in before, this was taking a very long time, upwards of 10 minutes for my VN project - which consists of 33 character files, 50 timeline files, and about 5000 events - as the baseline. My pverious changes brought it down to 2 minutes. This brings it down further, to under 20 seconds.  

This is implemented by just not actually doing the event processing at all, using the same changes that were happening on runtime as well. This defers almost all the event processing until actually needed, which in Editor is only when using the timeline editor, so it loads it as part of the step for preparing the timeline editor object


thebardsrc:
should be good but leaving it in draft for the moment, because I have to leave and want to double-check that new events in the editor get created as ready (I believe they do)

PR #1111, opened 2022-08-14
Add glossary

Jowan-Spooner:
# Glossary resource
There is a new resource called DialogicGlossary.
It has two properties: 
- entries (Dictionary)
- enabled (bool)

# Glossary editor
There is a new settings editor for glossaries that allows to create new glossaries and edit their entries.

# Glossary subsystem
There is a new glossary subsystem. The most important part is the parse_glossary method that is used by the Text event.

# Glossary event
There is a glossary event, but it is currently empty. 

Other:
- moves the example scenes into an folder in the ExampleAssets.


coppolaemilio:
It would be nice that when we hover on settings like this ones we show a tooltip with how to edit that via code. I can create a ticket so that we can track those kind of improvements but wanted to leave here the comment as a reference:

Screen Shot 2022-08-14 at 19 48 54

coppolaemilio:
The entries editor should be disabled if there isn't any glossary file.

Screen Shot 2022-08-14 at 19 58 45

coppolaemilio:
Getting errors when I try to create an entry
image

coppolaemilio:
Pressing tab to populate the field works until you want to go out of the Text editor and it doesn't change to the Extra input.
Same trying to jump from the Extra to Enabled. After that the focus continues working normally.
Screen Shot 2022-08-14 at 20 04 22

coppolaemilio:
Having to press `enter` to "save" the changes to the entry name is a bit weird, specially if rename the entry and close the settings dialog. It should save on text modified.
Screen Shot 2022-08-14 at 20 11 32

coppolaemilio:
`Enabled` on an entry should be **on** by default

coppolaemilio:
You should be able to print variable values by using the usual `{ variable }` syntax inside the glossary entry

coppolaemilio:
If you create two entries with the same name, after closing the settings dialog, they first one will get overwritten by the second's entry. Not sure if there is an easy way to identify if the entry already exists but I would try to either warn you about it, or add `_1` on the new one. It is one of those annoying things to fix, but it would suck to lose some data because of unknowingly naming two entries the same.

Jowan-Spooner:
True. Another problem is that you can have two entries named the same in different glossaries. While this won't erase either of them, they won't both work in game as it expects the names of all entries to be different.

Jowan-Spooner:
Made fixes:
- Hide Entry Editor when no entry or glossary selected
- Update name instantly
- Add {variable} support to entry display

Issue #1121, opened 2022-08-15
Portraits are added twice


Jowan-Spooner:
@thebardsrc When using the Join event the portraits are added twice, not sure why. 
Also they are not added to the PortraitHolder anymore, but to the positions themselves. Maybe that's intended, not sure.

thebardsrc:
> When using the Join event the portraits are added twice, not sure why.

Hm I didn't see that myself in testing, but i'll check it out

> Also they are not added to the PortraitHolder anymore, but to the positions themselves. Maybe that's intended, not sure.

Custom portrait scenes were being added to the position themselves, I copied  how that that method with the new PlainPortrait scene.

thebardsrc:
ohhh ok. so I dug through more, because you really had me confused with that PortraitHolder statement because I didn't remember touching anything about that... It got removed from the code entirely during one of the initial Godot 4 conversion changes

which, honestly, is most of the change i wanted to make to begin with, I _want_ them in a node like PortraitHolder for the character sorting and other things. so, perfect, it's already there, I just have to put them back on it

PR #1124, opened 2022-08-16
Portrait Position Overhaul

### LINKED ISSUE #1120: Character Offset broken
@thebardsrc I think this came from your changes. I'm getting an error in the Portraits subsystem change_portrait() method:
```
	$Portrait.position.x = character.portraits[portrait]['offset']['x'] * character.portraits[portrait].scale *character.scale
	$Portrait.position.y = character.portraits[portrait]['offset']['y'] * character.portraits[portrait].scale *character.scale
```
It says there is no offset. I think you should use `.get('offset', Vector2())`. Also the offset is stored as a vector not as two values if I'm not mistaken

Edit: Other values are also assumed to be in portrait dictionary when they are not: scale, mirror

Edit: Also this line seems to multiply the scale twice, not sure why:
grafik

thebardsrc:
Putting in draft because I need some more space then tiny airplane table to make a test scene to check it, but letting others review. Also need to take a look at the RPG Mode functions to make sure they are current, which I haven't had the chance to try yet, as well as modify saving/loading for this change.  Also it requires interaction before continuing with the new event, I didn't have the chance to figure out how to make it immediately continue yet.

Re-attaches the portraits to the PortraitHolder CanvasLayer, which was originally intended but lost in the conversion from Godot 3.5 WIP to Godot 4. With those reattached, adds back in Z-Indexing on the portraits.

Additionally, adds a new event type,  `DialogicPositionEvent`. This event type allows for manipulating the positions themselves. With this, positions can be moved in a timeline, and new positions can be added. Position changes can either be made by relative or absolute position, and can either be immediate or a transition over time. Also it allows resetting all the positions back to their defaults, either from the original scene loaded, or from the initial addition of a position at runtime.

Additionally, adds a linear transition to the character Update as well, for smoothly transitioning a character to a new position. This should run in parallel to an animation, so they both happen as one action.

Other tweaks are creating a faster lookup dictionary for the positions, to simplify changing them, and storing the actual current position on the character node as a metatag for ease of lookup, as well as a new type for the Editor parts to allow larger numbers for the position updates. Also fixes #1120 and #1121.


Jowan-Spooner:
Haven't tested anything yet, but I guess the positions should also be saved and loaded. I would be okay if we leave that for later, but then we need to make an issue about it.

thebardsrc:
alright everything seems to be working in tests

Jowan-Spooner (approved):

Issue #1010, opened 2022-07-13
Input Event


Jowan-Spooner:
An event that opens a input field with some text above it maybe and a variable that will be set once the input is entered (or accepted with a button).

This is a basic VN feature, so I think supporting it by default would be a cool thing to do.

PR #1144, opened 2022-08-24
Add TextInput event

Jowan-Spooner:
- fix for #1010

# Text Input event
Options:
- Variable (the variable that the input will be stored to)
- Text (will by default be displayed above the line edit)
- Default
- Placeholder
- Allow Empty

*Is missing an icon*

# Text input display node
A control node that has three exported node paths to a text label, an input line edit and a confirmation button.

A node of this type along the needed child nodes was added to the DefaultDialogNode scene.


addons/dialogic/Events/TextInput/DialogicDisplay_TextInput.gd
```diff
@@ -0,0 +1,45 @@
...
+
+func set_allow_empty(boolean:bool) -> void:
+	allow_empty = boolean
+
+func _on_input_text_changed(text:String) -> void:
+	if confirmation_button.is_empty():
+		return
+	get_node(confirmation_button).disabled = !allow_empty and text.is_empty()
+
+func _on_confirmation_button_pressed():
```
coppolaemilio, line 42:
```suggestion
func _on_confirmation_button_pressed() -> void:
```

coppolaemilio:
Okay, after testing it I think there should be a way to customize what the input says. 
I know this makes things a bit more complicated because of localization but I'm sure people will want to customize the "Please enter some text" prompt.

Issue #1156, opened 2022-08-25
[2.0] Text Editor mode stripping the names from Text events


thebardsrc:
Text Editor is stripping the names of who's speaking from text events. Specifically that, if the timeline's opened in Visual Editor its fine until Text Editor gets to it, then they're lost

Example:
![image](https://user-images.githubusercontent.com/2214793/186545713-d8ed195c-2bd0-461f-b934-ea467c8cf428.png)

Switching to Text Editor:
![image](https://user-images.githubusercontent.com/2214793/186545767-2a8e78ea-9af1-4787-8214-0ad7f8c01dcf.png)

Then back in Visual Editor again, its permanently lost:
![image](https://user-images.githubusercontent.com/2214793/186545809-c906e131-28ea-4acc-a092-a9232ea77698.png)

thebardsrc:
ok i think this is a side effect of the change in #1154. trying to Cut an event from the visual timeline editor also puts the event on the clipboard without the name as well

PR #1157, opened 2022-08-25
Fix character name getting lost

Jowan-Spooner:
- hopefully a fix for #1156


thebardsrc (approved):
confirmed it fixes the bug

PR #1161, opened 2022-08-25
2.0 Runtime Character Lookup Table

### LINKED ISSUE #1058: Allow multiple characters with same name
This is important for saving events (mainly character and text).

Also the picker should be improved so that they can be easily distinguished.

thebardsrc:
Optimization for runtime, preloads all the Characters when the Autoload begins, so Text and Character events pull from the lookup table instead of going through repeated loops of searching res:// for the Characters and loading them ad-hoc. Character Objects are pretty small and load quickly, so having them all loaded from the beginning is fine. 

This is also preliminary change to an implementation to resolve #1058, but that will require further more extensive changes to both runtime and editor, so that will be a followup PR.


coppolaemilio:
That's really smart. Looking forward to this one!

addons/dialogic/Other/DialogicGameHandler.gd
```diff
@@ -36,6 +37,13 @@ func _ready() -> void:
 	if Engine.is_editor_hint() == false:
 		deferred_loader = Thread.new()
 		
+		# Runtime will also build the character_directory dictionary. Editor will have to handle it a different way
+		var characters = DialogicUtil.list_resources_of_type(".dch")
+		
+		for character in characters:
+			var charfile = load(character)
```
coppolaemilio, line 44: type? 👀

addons/dialogic/Other/DialogicGameHandler.gd
```diff
@@ -4,6 +4,7 @@ enum states {IDLE, SHOWING_TEXT, ANIMATING, AWAITING_CHOICE, WAITING}
 
 var current_timeline = null
 var current_timeline_events = []
+var character_directory = {}
```
coppolaemilio, line 7:
```suggestion
var character_directory:Dictionary = {}
```

coppolaemilio (commented):
Won't leave comments everywhere, but can we add types? 😅

addons/dialogic/Other/DialogicGameHandler.gd
thebardsrc:
```suggestion
			var charfile: DialogicCharacter= load(character)
```

PR #1162, opened 2022-08-26
Big character editor/custom portraits updates

Jowan-Spooner:
# Character editor updates
- made sections have clear styles
- moved main portrait settings to the general tab
- remade portrait list as a TREE, might allow future grouping/folder-system. Tree items have a duplicate and remove button. Also the default portrait shows a star.
- portrait list items only have a name. The image and scene can be set from the portrait settings section
- Portrait settings have been put into different sections
- by default the scene is the DefaultPortrait scene.
- if it's not, then the Image section is hidden
- new option "ignore scale" added -> only intended for custom scenes (hidden otherwise).
 
## Preview of custom scenes
- both normal and custom scenes work the same way now, by instancing them into the preview. If they are in tool mode (which the defaults are) this comes with a small risk of someone doing damage to the editor, but it's unlikely if we give good examples.
- if a custom scene is selected it will be instanced and placed at the pivot point with the correct offset and scale.
- If the main script of that scene is a tool script, it will attempt to call '_update_portrait' on that script (as in game). This allows to use the same scene on many portrait and (possibly) even characters. But that is unlikely useful.
- to enable the Full View mode custom portraits can implement the _get_covered_rect method to specify the region that should be visible.

# Custom portrait scenes
- custom portrait scenes are expected to have a root Node2D node. The position and scale of this node will be altered by dialogic. All the children can be placed and scaled however wanted
you can implement these methods:
- _update_portrait(character, portrait)
- _set_mirror(mirror)
- _set_extra_data(string)
- _get_covered_rect() # for editor
- _should_do_portrait_update(character, portrait) # defaults to true. If it returns true, means that when a change to another portrait with the same scene is attempted it will just keep this one and call `_update_portrait` instead of creating a new scene.

# Example portraits
There have been multiple tests. Not sure if we will keep them for release, but I thought it would be useful to have some shared tests.

### CustomPortrait_AnimatedSprite 
Currently only has one animation but in theory could have more animations. Tries to play an animation of the name of the portrait. Used on character Emilio.

### CustomPortrait_Face_Atlas
Has 5 faces and one accessoir.
Changes faces based on the portrait name.
Can also add and remove a little alien accessoir with extra_data. Used on character Jowan.

See these in action in timeline Chapter 1


thebardsrc:
the examples are nice in there, if the art assets come from somewhere else we need to make sure the license file reflects that with whatever license and necessary attibution requirements they fall under

Jowan-Spooner:
> the examples are nice in there, if the art assets come from somewhere else we need to make sure the license file reflects that with whatever license and necessary attibution requirements they fall under

True. I usually only uses cc0 asstets from oga for testing.

coppolaemilio:
Opened the branch, tried running the Chapter1.dtl example and I get this error:

image

coppolaemilio:
I get this error when opening the editor:

>  res://addons/dialogic/Events/Character/Subsystem_Portraits.gd:0 - Parse Error: Could not find base class "DialogicSubsystem".

coppolaemilio:
I made it so that the error doesn't stop the game from running and shows an error.

PR #1163, opened 2022-08-26
[2.0] Further timeline speedups

thebardsrc:
Implements further speedups to the timeline resource, across the board.

To summarize the changes, at the core, the crux of the problem with the Timeline resource is it seems resources aren't *actually* meant to be used in this fashion. Building a Resource from a bunch of modular parts is inherently slow, because has to keep reloading all of the modules that it uses, at the step of `DialogicUtil.get_event_by_string()`. Then when it actually parses the event and fills out the data in it, that further reloads the module and adds even more load time, which I had previously cut down on by adding a thread for deferred processing at runtime. 

There seems to be no way to implement a cache for this directly in the Resource loader, because the Resource loader checks the resource script file as a whole as the very *first* thing, and marks it wholesale unusable if any thing is missing. For example, at runtime I added a cache to the event scripts, and tried having the Timeline loader use it, but the resource loader loads before the Autoload for DialogicGameHandler, and even trying to add it in a conditional to check if it's loaded yet, it was still throwing an error that it was missing and it marks that resource type unusable for the game's runtime.

The only solution to this is by caching elsewhere, and processing it as a secondary step. Stripping the Timeline Loader itself down to just building a barebones event list of just strings, and then making the events themselves after the cache can finally build. In runtime, a cache on the  DialogicGameHandler that builds the script cache while it's loading the subsystems. In editor will be a bit more complicated, since it will have to monitor for changes in the filesystem and rebuild as well, but it will probably have to be attached to the EditorView.  then all the normal event creation will need to check the cache instead of actually iterating through the files calling `DialogicUtil.get_event_by_string()` . For saving, that seems to be taking a long time as well, and I haven't looked at it but I suspect it's doing the same multiple loads thing, and will  have to switch to the cache as well.


thebardsrc:
Commits as of this writing: 
Implements it on the runtime for loading only. Changes `TimelineUtil.text_to_events()` to simply load the .dtl file, strip out the blank lines, and build an array with just the text values for each line. As all the work processing it is deferred until later, total load time for just the resource loading at runtime comes down to ~10 ms. 

Initial load time of test timeline of 220 events before changes: around 9000 ms 

Load time after this change: **30** ms
image

I was... not expecting anywhere *near* this much of an improvement. I was hoping at *best* to get it down to 1/4 of the original load time, to go from 9000 to 1500 ms

And having it process all the events at this stage as well only adds about 10 more ms:
image

The deferred loader thread is simply not needed any longer. It's completely redundant now.

thebardsrc:
summary of where i have to leave off for the moment: print's are littered everywhere at the moment for performance checking, but saving for Text timelines is implemented, and both Text and Visual timeline loads work. Character matching isnt working in Visual timeline yet, nor is saving

however, there is the following... strange issue, whcih I thought at first started with #1129, but is even still getting it trying to undo those changes to get rid of TimelineUtil. I suspect is an issue with the current Godot version, ultimately. but on project start the following errors will occur, and will repeatedly show up over and over:

'ERROR: Script does not inherit a CustomResourceLoader: res://addons/dialogic/Resources/TimelineResourceLoader.gd.
   at: (core/io/resource_loader.cpp:979)
ERROR: Script does not inherit a CustomResourceSaver: res://addons/dialogic/Resources/TimelineResourceSaver.gd.
   at: (core/io/resource_saver.cpp:211)'

due to this, Saving never completes, and trying to open a new timeline from the file picker gets this error: 
'editor/editor_node.cpp:1201 - Condition "!res.is_valid()" is true. Returning: ERR_CANT_OPEN'

**HOWEVER**

For some bizzare reason, *touching* those two files fixes this for the session, until you close and reopen Godot again. So opening up the TimelineResourceLoader and TimelineResourceSaver files, adding a space in there somewhere, and then saving, will let it work fine

thebardsrc:
visual timelines should save and load properly now as well. need some more thorough testing

thebardsrc:
the timleine resource errors are gone for me now. at the moment the indents aren't being handled correctly and that's being lost, once I got that resolved eveyrhting should be good

thebardsrc:
didnt have much time to work on it today, but fixed the tab calculation, so now saving and loading properly for the Visual timeline editor.  the tabs are being lost, however, in the switch between Text and Visual timelines, so something different is happening with those that I still need to figure out

thebardsrc:
alright! now saving works properly from both timelines by themselves, and they both can switch modes properly from one to the other without issue. Only one final problem left now: how the "Test Timeline" button works is now screwing it up overriding the file before it runs. once that's done I'll document all the final changes

thebardsrc:
Finally! All of them are good! 

There are a few extra steps on the saving on each editor that are probably not necessary, but I was having a lot of trouble getting the data to switch back and forth between the Visual and Text editors to work properly without them. But overall, there are the following changes:

- Adds an event cache, attached to the EditorNode in editor, and attached to the DialogicGameHandler at runtime. This immensely speeds up the actual loading and creation of events, due to not constantly reiterating through the script files and reloading them over and over as it was before. in editor, it should probably rebuild that if a new one is detected (if someoe adds custom events, for example), but doesn't currently.
- Adds a character cache of all the current characters in the project. this also is attached to EditorNode in editor, and DialogicGameHandler at runtime. at runtime, when creating the Character or Text events, it will search for them in the character cache and use it from there, speeding up load time.  due to an editor-only error trying to find the EditorNode in the scene tree from in the actual DialogicEvent creation to pull the character cache, a reference is passed to them at node creation as a meta property - this is a temporary handling, and will be fixed with other revamps to the charatcter suggestor.  as with the event cache, it should rebuild in-editor if a Character is saved, which I didn't because the Character editor was being rebuilt at the same time and im not sure if its 100% yet. 
- TimelineUtil was removed, and it's two functions were moved elsewhere, for reasons following
- Timeline Loader is stripped down to barebones, and only initially generates an array of the strings of the events, instead of DialogicEvents themselves. This is the only way, unfortunately, to allow the loader to use the event cache
- Timeline Saver is stripped down as well. Due to the translation functions being in the saver as well, a DialogicEvents timeline *is* needed here, but it only takes the text from the `event_node_as_text` property from each DialogicEvent
- To facilitate the above two changes, a property was added to DialogicTimeline, `_events_processed`. This is set on load as a flag to denote that the timeline has not been converted into Events yet, and is set to true once it has been handled. Additionally, the same flag is used for saving to denote whether the events have been converted back into text again and it's ready to save, and the saver will reject it if the timeline is passed to it with it set to true
- adds a further protection as well to saving, and will not save if the timeline is 0 events. this should get rid of other causes of  losing data because of a problem with the data sent to the saver. due to this change, users will need to delete the timelines if they want to blank them out
- EditorView.process_timeline(): this is where one of the two TimelineUtil functions went. It goes through the array of text events from the loader - or from an editor - and creates Event nodes from it and populates it, and sets `_events_processed` once its completed. Additionally, this also prepares a DialogicTimeline for saving, by populating the `event_node_as_text` with the current event text
- Dialogic.process_timeline(): corresponding version of the processor for runtime. this will automatically be called by both start_timeline() and preload_timeline(), and so only particually need to be used separately if creating a timeline in code at runtime from strings, rather than builidng it from DialogicEvent nodes from the event cache
- TextEditor now loads the data from events from `event_node_as_text` only, discarding the rest. When saving, it builds a new array from the editor of just text node, as if it were the same as loading a Timeline (since the text mode is identical to .dtl file), and sends it back to EditorView.process_timeline() before saving. It 
- VisualEditor - which has now had it's files renamed to be that instead of TimelineEditor - has the most speedups. it loads and stores AddEventButton.tscn to instance on each use when populating the event buttons, instead of loading it, whihc speeds it up a bit and also cuts out a lot of the 'invalid UUID' warnings that was being generated by each load of it before. It also builds the buttons based on the event cache, cutting out further file searches and loads.    when saving, like TextEditor, it builds a new array of text, created from each event node's `_store_as_string()`, and sends it back to EditorView.process_timeline() before saving. a bit redundant to send it back first on this one, instead of just storing the '_store_as_string()' back in each node and sending it on to the saver, but its for safety, and also fixed some issue with the data not exchanging correctly when switching back and forth between TextEditor and VisualEditor. it then switches the timeline array back to that same text array after saving, for that same reason. 
- Both timeline editors are now properly and completely hooked up to the `set_resource_unsaved()` signal in the toolbar, so there's visual notification if any changes have been made. Both timelines will now only save if there has been changes, to save on unnecessary file saving. This unfortunately doesn't apply to the not-current timeline, so other timelines still open in Godot's memory will not reflect this in the recent file suggestions
- Due to the above issue with it not applying to other open timelines, a meta property is set on the timelines as well, `timeline_not_saved`. This is set as true if a change is made, and set as false once it is saved. As (due to the recent suggestion selector in the toolbar) saving the project itself will try saving _all_ of the open timelines, not just the current one on the timeline editor, the Timeline Saver checks for this meta property on the DialogicTimeline before doing anything else, so the non-current timelines don't try saving as well, possibly throwing errors and blanking the file because they aren't  garaunteed to have been prepared for save in wherever Godot is holding them open
- made the current editor mode save properly, so it wouldnt keep reverting back to visual editor on project restart (this seemed to start with a different recent change, not sure which, it used to remember it before)
- due to the massive speedup having the event cache brings, gets rid of the old processing thread that was previously used for optimization entirely. this means Dialogic.process_timeline() will parse every event when run, although it still isn't loading nested timelines due to mainly memory constraints. function can be added to this later if needed in the preloader, I still need to test on my RPG for Dialogic's response time now compared to 1.x where there was a noticable stutter on every timeline start.
- adds some queue_free() in a few places as _exit_timeline() functions, to try and minimize the leaked resources at project close. doesn't get rid of them all, though.
- also includes adjustments to the 1.x Converter for the changes from other recent PR to the Character file format, which I needed to include to rebuild my timelines as I was testing because they kept being corrupted by in-progress work and i was running out and it would not run due to those changes. also includes an icon in the folder, so Godot doesnt show a missing file error
- adds a fix for DialogicUtil.get_event_scripts() so it doesn't generate errors if there are empty folders in the Event paths without event.gd files 

And I think that's all of them

Issue #1177, opened 2022-08-28
2.0 Recent Files Chooser Problem


thebardsrc:
Dropdown for the recent files in the Chooser shows only the file name. If there's more than one file by that filename in the directory, it chooses one at random

thebardsrc:
Also assigning this one to me since it's the same picker as I have to change to make the Character chooser support file paths

PR #1207, opened 2022-09-02
Character and Timeline Partial Paths Support

### LINKED ISSUE #1170: Support for generic write-in characters directly in the text event
From Discord: "Request support for just generic write-in characters directly in the text event.
For when a character only have a temporary role, and don't need a character file. It gets messy when a character who only appears with two lines in one scene needs to be defined as a character."

should be real simple to implement, with simply surrounding it in special character that can't be part of a filename. Something like this:

`*Temporary*: Hi, I'm just a guest here`

Mostly just depends on which special character works best with the regex without rewriting it (if possible)

### LINKED ISSUE #1189: Being able to scape the `:` to avoid it being recognized as a character name
If I want to write a message that just says something like:

> Hey! How are you doing? :)

It will only print: 

> )

### LINKED ISSUE #1191: ComplexPicker removes duplicate suggestions
When creating suggestions for the complex picker they are put into a dictionary. This means that duplicate entries just override each other. 

Noticed this while having two timelines with the same name in different folders and trying to find both in the "recently used" list. Only one showed up.

This probably needs fixing mostly in the method that lists these to add some indication. Maybe similar to what @thebardsrc is doing in for #1058 (showing more parts of the path for duplicates).

### LINKED ISSUE #1205: 2.0 RPG Mode Behavior Issues
From Discord:
"First of. There is no position 0 in the example layout. I had to set it up myself in my copy of it.

Secund, if a portrait is not set for a line with a new character, no portrait will be shown.
Expected: "Don't change", when in RPG mode, should default to either the default portrait or the last used one for that character."

thebardsrc:
Makes changes to allow for partial paths in Character names, to allow for multiple characters with the same file name. Implements this with the following changes:

- Updates the previously made character directory from #1161 with a generated lookup table, ensuring every character has a unique path to it, and if there are multiples with the same name it will prefix the next higher folder on it
- Changes the actual Character and Text events to use this new lookup table. As a fallback, for example if a manually created timeline has more of the character path then necessary, they will iterate through trough the entire directory and choose the first one that matches the path. It will also allow for spaces in the name, and automatically put the name in quotes in the timeline if it is needed 
- Makes some changes to ComplexPicker to accomodate this new directory. due to its design, a number of things had to be changed directly on it to allow it to work without file loading errors, as well as still allow drag and drop to work
- Removes pretty_name() from being used for Characters, due to with path support, and other changes, it allows for spaces in names and so forth. pretty_name() will cause ambiguity, as you then can't tell between "AliceBob.dch" and "Alice Bob.dch" as it would try and put a space between the names
-  makes the Character Editor rebuild the character directory when it saves a character

Other changes in this update:
- Attaches a light instance of DialogicGameHandler to the EditorView. This is for allowing Editor to get to DGH's character and event directories, removing the separate duplicate functions for building the directories and the process_timeline() function, for simpler maintenace. Due to the order in which the buttons are built for the editor, since they use the event cache, EditorView still needs to separately generate the event cache on its own 
- Additionally builds a Timeline directory, in the same way as above. 
- With the Timeline directory in place, reintroduces ComplexPicker to the Jump nodes for picking the timeline, and also updates Jump to use the same lookup table, with a fallback to the file name. This fixes #1191 and #1177
- reworks the Text and Character event regex, thanks to @Jowan-Spooner rewriting, to make it less error-prone, allowing for the quotes for spaces in names, and only tries to match the colon on the first word.  as caveat, an actual _intentional_ colon in a Text event without a character will need an escape to it, which this will generate automatically for Visual Editor. This fixes #1189
- at runtime only, if both the two lookups for the character in Character and Text fail to find it, it is treated as a guest character, and added to the directory as an empty Character. This fixes #1170
- readjusts behaviors for RPG mode (which doesn't need to use the character directory, since preloading all of the characters makes all resources present so they are there without needing to load). Changes behavior to remember the last used portrait, and if one was not set then to use the default portrait, when the Text event does not specify a specific character. This fixes #1205
- changes the RPG mode to place on Position 1, rather than Position 0. this is due to consistency with how positions are used in Visual Novel mode. this also allows running timelines out of the box in RPG mode with only that toggle, because the default timeline does not have a position 0, and it is additional steps to change the test scene, and a new user won't necessarily know where to find it for RPG  mode and wonder why it just crashes. NOTE: I was only able to change one of the two example RPG mode timelines, I did not change /ExampleScenes/RPG_BigCharacterDialogNode.tscn  because that one actually has a missing resource and I do not know what it was supposed to be nor could I find it trying to explore the history


addons/dialogic/Events/Character/PortraitSettings.tscn
```diff
@@ -59,140 +59,140 @@ popup/item_1/id = 1
...
 [node name="Info" type="Label" parent="General"]
 offset_top = 75.0
-offset_right = 522.0
-offset_bottom = 257.0
+offset_right = 1112.0
+offset_bottom = 175.0
 text = "Visual Novel: Multiple characters can show up in different positions at the same time. They have to be joined and leaved manually.
 
 RPG: Only the character that speaks is visible (always in the same location). Does not have to be joined/left manually. Uses only position_idx 0."
 label_settings = ExtResource("2_47ms3")
```
Jowan-Spooner, line 67: Could you update the label on the settings page? About using idx 1 now.

addons/dialogic/Events/Character/Subsystem_Portraits.gd
```diff
@@ -245,6 +245,12 @@ func get_joined_characters() -> Array:
 
 func update_rpg_portrait_mode(character:DialogicCharacter = null, portrait:String = "") -> void:
 	if DialogicUtil.get_project_setting('dialogic/portrait_mode', 0) == DialogicCharacterEvent.PortraitModes.RPG:
+		if portrait == "":
+			if character.has_meta('rpg_last_portrait'):
+				portrait = character.get_meta('rpg_last_portrait')
```
Jowan-Spooner, line 250: In the last portrait something we should rather store in the dialogic.cuurent_state_info dict? Because we might need it after saving and the metadata is nit persistent I think (and also global).

addons/dialogic/Events/Jump/event.gd
```diff
@@ -9,12 +9,17 @@ var LabelName : String = ""
 var _timeline_file: String = ""
 var _timeline_loaded: bool = false
 
+var _timeline_directory: Dictionary = {}
```
Jowan-Spooner, line 12: Is this actually used? Could'nt see it used, but stupid me decided to make this review on a phone... So could be my bad.

Jowan-Spooner (commented):
Things I noticed while looking through the code. Will test asap

coppolaemilio:
I see that Jowan already started the review, but the most obvious issue I see is with the variable types, I want to enforce static typing everywhere and I see that most of the new lines in this PR don't do it. Not a big deal since the supposed performance improvement is 20% in some cases, but I think it is better for us internally as well. I can totally merge this and later on continue working in transforming all to static, but it is easier when you are writing something new since you know what is the expected type. I'll keep testing it out :)

coppolaemilio:
I added the:
`if not current_timeline_events[event_index].is_connected('event_finished', handle_next_event):`
conditional because if you spammed the "next" key you would get errors in the output saying that "the signal is already connected".

addons/dialogic/Events/Character/PortraitSettings.tscn
thebardsrc: updated

addons/dialogic/Events/Character/Subsystem_Portraits.gd
thebardsrc: ahh yep, didnt think of that one

thebardsrc: changed now

addons/dialogic/Events/Jump/event.gd
thebardsrc: ah youre right, i had started with the same setup as in Character and Text, but Jump uses the shortcode loader so couldnt initialize the same way, and its overall much simpler

thebardsrc:
> I see that Jowan already started the review, but the most obvious issue I see is with the variable types, I want to enforce static typing everywhere and I see that most of the new lines in this PR don't do it. Not a big deal since the supposed performance improvement is 20% in some cases, but I think it is better for us internally as well. I can totally merge this and later on continue working in transforming all to static, but it is easier when you are writing something new since you know what is the expected type. I'll keep testing it out :)

i thought I had been good about statically typing all my new code -_-

PR #1218, opened 2022-09-11
Speaking_Character Signal

thebardsrc:
Adds a signal `speaking_character` to the Text subsystem. Sends the current display name of a Text event's character to the signal, or "(Nobody)" if there's not any character on that event. 

Purpose is to allow a game to listen for that event and do something with it, if they choose. For example, using multiple text boxes for different characters in a 2D/3D game that are positioned with that character, without any manipulation of themes in the timeline.

### DIFF
--- a/addons/dialogic/Events/Text/Subsystem_Text.gd
+++ b/addons/dialogic/Events/Text/Subsystem_Text.gd
@@ -5,6 +5,7 @@ var character_colors := {}
 var color_regex := RegEx.new()
 
 signal text_finished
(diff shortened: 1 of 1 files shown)


coppolaemilio:
Not super sold on the (Nobody) option, but I'll approve and we can rename in the future 😅

thebardsrc:
could be something else, doesnt really matter. but it's a way to let the non-character text box have a string to match too. Parenthesis are disallowed in character names, so it will be garaunteed to not be a real character that way

Issue #1100, opened 2022-08-07
2.x 'Already Read Text' feature


exelia-antonov:
I'm gonna put this one on 2.x for now, but I don't think it would be that difficult to implement for just 2.0.

One feature that's common in Visual Novels, is keeping track of text you've already read. Usually this is at minimum used for a setting to skip already read text, but sometimes this is also used to like change the colors of the text too if you've already read it as well. Other types of games too, where you have characters giving large lines of dialogue, you may want this as well to give the player an optional skip for text they have already read.

This would be it seems pretty simple to implement. It would just need adding additional property of `already_read` to the Text node, then if the feature's enabled in project settings, it would save that to the memory as part of the save data, then when the timeline is loaded the DialogicTimelineFormatLoader just checks the save for the already read lines from that timeline and populates the property

Proggle:
It would be useful to expose the already_read property for use in the timeline itself.

RPGs commonly have two-text NPCs, where the first time you talk to them you get their full story, and the second time you talk (and every time after then) you get a shorter piece of text instead.

Rather than needing to store that text in two timelines, or define a Value for each repeated timeline, it'd be a lot easier if you could just branch on already_read.

PR #1219, opened 2022-09-11
History Subsystem

thebardsrc:
History subsystem. Built to be flexible so it can be used in various different ways. Has two different types of histories: a full history stack of all events, and a separate history of just all read text options. Full history is meant for multiple uses including data rollback, read text is more focused for VN/RPG type use, where it typically allows for styling the text a different color or allowing for a toggle to only skip already-read text (see #1100). before saving it into the full state, it will strip the actual Event object from the timeline

I copied it's files from the Glossary event, so some of them are currently still Glossary files. I'm using this draft to document things I need to implement:

Core necessary things:
- [x] actual project settings to enable the system as a whole, whether the two separate histories are used individually, and the actual length of the full history (sometimes you don't want 100% of it and only allow a player to go back only so far back)
- [x] saving a few of the key things needed for a rollback that aren't directly in the history stack (choices made, previous variable values when they are changed, etc). 
- [x] options to save expanded data from a few of the specific event types to the save file directly, if wanted. also a function to be able to reload the timelines separately and grab those events to repopulate it after loading a save, if you're not saving expanded data 
- [x] an event to deliberately clear the history in a timeline (see for example DDLC's history manipulation)
- [x] an additional type on the Text event for the already_read, and a quick check if it's in the dictionary on handle_event() to set that variable. ideally , beyond that any implementaion of different behavior with that set should be on a Theme rather than directly in Dialogic, probably

A few other things will be tangentially made possible with this directly or with changes needed for making this enabled, so I'll link them as I go if I find them

The above is all the core functionality of the History subsystem itself. Then after I'll make an example History view, probably mimicing the way that Dialogic 1.x History looks for starters. But as it's own subsystem of just data, people will be able to use it and customize the actual visuals of the history however they want.


thebardsrc:
note to self: no explicit choice data needs to be saved, as the choices are built from the Text event before it, and it only calls handle_event() on whichever choice is selected

Jowan-Spooner:
I'm wondering if events should be able to implement a _execute_reverse() or _rollback() method where they get acess to some previous data? But I haven't looked into how you want to implement the rollback functionality. 

Also is the history stack saved? Can it be loaded back in from different slots?

thebardsrc:
history stack is put in the state, yeah, in a stripped down form (which timeline it is, which line it is, that event type it is) at minimum. there will be options to save more data than that if they want it, instead of having to reload the timelines and grab the full events again, which could be useful in certain different uses of the timeline. in memory before that, itll also have the full event object

for actual rollback function, my thought had been just an actual event reversing functionality would just play the reverse timeline in order, one step at a time. there's a few spots where it would need to like reference back to the original timeline to rebuild while doing that, such as choices so it can rebuild the choice menu

thebardsrc:
Niche case note for idk how to fix it right now, whether it really needs to be, and can be addressed later: the bbcode for putting in a random word in a text block, theres not really a way to save that for the history

thebardsrc:
example History display created, and added to the Default scene. it's only on the Default theme, not the Special theme

oh that one wasn't in my checklists, gotta finish those still

thebardsrc:
noting here: since I had been planning on updating the Visual Novel template (since I'm using it for one of my projects myself), I'm going to have that as the example implementation for the usage of the `text_read_history`, as well as actually using the full_history for a rollback function 

also, in line with more DDLC shenanigans, gonna make a custom event type that can put fake events in the History but don't show up elswhere. I'll attach it onto here as a zip, probly, until the new Dialogic website can be set up with the custom events gallery

thebardsrc:
decided to go with a signal in the History subsystem for the already read text event, `history_text_already_read` rather than going with another type added to Text itself. cleaner this way, instead of having to add another property on the Text event itself and having to update a bunch of functions with it. the History subsystem is handled before the event itself is, so the signal can make the display scene ready ahead of time to apply to whatever changes you want to use

thebardsrc:
ok there was some more options i wanted to add to history for more options for the full history saving, but I genuinely dont rememger them right now and cant find the notes I made about them. so I'm gonna just open this one up as ready now, and add those later when i fianlly remmber the other situations

coppolaemilio:
image

I'm getting these errors when opening the regular test project for the first and second time using this branch, no idea why, just posting the screenshot here

coppolaemilio:
That fixes the two first ones

coppolaemilio:
That should be all the errors, I'll fix the merge issues now.

coppolaemilio:
@thebardsrc I'm getting this error when trying to run a timeline with the first event being a text event with no character set:

image

PR #1224, opened 2022-09-14
Text sped 0 special case

KvaGram:
Problem:
When working on a project that requires a text speed of 0, voice events, autocontinue, and making use of pause effect in the text events, many text events began with a large pause.

Debug process:

The problem was soon identified. In function reveal_text, the text-box's timer is set to speed.
However, with the speed value set to 0,  godot's timer logic breaks down. The wait time was not set, and in events where this problem occurred, it was observed it used the wait time set by the previous event's last pause effect.

Setting the wait time directly caused an error. A timer's wait time cannot be 0 or less (engine hardcoded).
Bypassing the timer in reveal_text with a direct call to continue_reveal appeared to fix the problem. 
However it caused a new problem.

Text events without a pause effect halted and did not exit to the next event.

This was at first believed to be a separate bug in Dialogic. In a way, maybe it was.

After lengthy line-by-line debugging, it was found that bypassing the timer caused some signal connections to not complete before the text finished revealing.

Complete fix:
Add a special case in reveal_text and continue_reveal checking for a speed of 0.
In Reveal text, force the timer to cycle for 0.01 seconds, allowing the signal connections to be set, while also not running the wait time from earlier pause effects.
In continue_reveal, bypass the timer and run recursively if speed is 0.


KvaGram:
updated the description with a full report of this adventure.

thebardsrc:
I can confirm that this fixes one bug I had in one of my converted timelines, where my normal project text speed is 0, a choice event began with text of a speed of 2 (which was originally there for a 1.x bug and I never removed since this is a timeline for debugging and launching different other branches), and then every event after that would have a wait included before they occured

Jowan-Spooner (approved):

PR #1228, opened 2022-09-16
Change timers process callback to physics

Xwdit:
Changed all timer process callbacks to physical frames so that they are framerate independent. This prevents dialogue from causing problems with certain game features that require precise timing. (e.g. replay system that can accelerate/decelerate)

This is based on a recent PR to SceneTreeTimer: https://github.com/godotengine/godot/pull/63026

### DIFF
--- a/addons/dialogic/Editor/Settings/Settings_General.gd
+++ b/addons/dialogic/Editor/Settings/Settings_General.gd
@@ -24,6 +24,7 @@ func refresh():
 	%CustomEventsFolderButton.icon = get_theme_icon("Folder", "EditorIcons")
 	%TestingSceneButton.icon = get_theme_icon("Folder", "EditorIcons")
 	%TestingSceneLabel.text = DialogicUtil.get_project_setting('dialogic/editor/test_dialog_scene', 'res://addons/dialogic/Other/DefaultDialogNode.tscn')
+	%PhysicsTimerButton.button_pressed = DialogicUtil.is_physics_timer()
 	
 	# Color Palett
 	color_palette = DialogicUtil.get_color_palette()
@@ -64,4 +65,8 @@ func _on_TestingSceneButton_pressed():
 func custom_testing_scene_selected(path:String):
 	%TestingSceneLabel.text = path
 	ProjectSettings.set_setting('dialogic/editor/test_dialog_scene', path)
+	ProjectSettings.save()
+
+func _on_physics_timer_button_toggled(button_pressed:bool):
+	ProjectSettings.set_setting('dialogic/timer/process_in_physics', button_pressed)
 	ProjectSettings.save()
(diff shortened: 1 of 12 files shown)


Jowan-Spooner (approved):
Thanks for the PR. Looks clean and logical to me.

coppolaemilio:
I think it makes sense to offer this as a feature, but you don't always want the timers to be based on frames, some times you need to sync with time, so an option should be added to select which one you want. 

Do you think you can make that change to the PR? 
It would require a new setting and check that setting for the timers to be set either by frame or time.

Xwdit:
> I think it makes sense to offer this as a feature, but you don't always want the timers to be based on frames, some times you need to sync with time, so an option should be added to select which one you want.
> 
> Do you think you can make that change to the PR? It would require a new setting and check that setting for the timers to be set either by frame or time.

Even if the timer process callback is changed to physics frame, it still be sync with time in seconds (based on delta time of physics frame), So it shouldn't cause any problems to regular users.  

But I still added the corresponding setting for some special cases, or for someone doesn't like timing in physical frames. (Idle process callback is the default value)  

We may also be able to instructing users to enable physical frame timing in the documentation to solve issues caused by erratic idle timing?

addons/dialogic/Editor/Settings/Settings_General.gd
```diff
@@ -65,3 +66,7 @@ func custom_testing_scene_selected(path:String):
 	%TestingSceneLabel.text = path
 	ProjectSettings.set_setting('dialogic/editor/test_dialog_scene', path)
 	ProjectSettings.save()
+
+func _on_physics_timer_button_toggled(button_pressed:bool):
```
Jowan-Spooner, line 70: `-> void:`

addons/dialogic/Other/DialogicUtil.gd
```diff
@@ -164,3 +164,11 @@ static func get_color_palette(default:bool = false) -> Dictionary:
 static func get_color(value:String) -> Color:
 	var colors = get_color_palette()
 	return colors[value]
+
+
+static func is_physics_timer()->bool:
+	return get_project_setting('dialogic/timer/process_in_physics', false)
+	
+
+static func update_timer_process_callback(timer:Timer):
```
Jowan-Spooner, line 173: ` -> void:`

addons/dialogic/Editor/VisualEditor/VisualEditor.gd
```diff
@@ -175,7 +175,7 @@ func load_batch(data):
 
 func _on_batch_loaded():
 	if _batches.size() > 0:
-		await get_tree().create_timer(0.01).timeout
+		await get_tree().create_timer(0.01, true, DialogicUtil.is_physics_timer()).timeout
```
Jowan-Spooner, line 178: Might not be a problem to do this here, but as this is in the editor, not in game, I'm not sure if it's necessary/good practice.

coppolaemilio: Yeah, I think I will revert this one since it is in-editor

PR #1247, opened 2022-10-09
Fixing adding timeline events and creating new timeline

GreenCloversGames:


coppolaemilio:
Thanks! I'll review it now :)

addons/dialogic/Resources/TimelineResourceSaver.gd
```diff
@@ -75,6 +75,7 @@ func _save(resource: Resource, path: String = '', flags: int = 0) -> int:
 			printerr(path + ": Timeline was not in ready state for saving! Timeline was not saved!")
 			return ERR_INVALID_DATA
 	else:
+		var file := FileAccess.open(path, FileAccess.WRITE)
```
coppolaemilio, line 78: Why are you opening this file here?

coppolaemilio (commented):
I have one question about the change in the `TimelineResourceSaver.gd`

addons/dialogic/Resources/TimelineResourceSaver.gd
GreenCloversGames: When you make a new timeline, it calls the save procedure, but since the new timeline doesn't have the "timeline_not_saved" metadata, it won't try to save it. This causes issues later on, as you don't save the file in the first place, it doesn't exist and you can't save to it.

coppolaemilio: Makes sense! Thank you :)

PR #1250, opened 2022-10-17
Fix for allowing character resources to be saved.

schiem:
There were two issues with character saving:
  * There was a check to see if the file exists first, resulting in new files never being created
  * The files were opened in READ mode instead of WRITE


Jowan-Spooner (approved):
Have not tested, but code looks good to me!

Jowan-Spooner:
Thanks!

PR #1253, opened 2022-10-17
Make sure resource get's passed instead of script

### LINKED ISSUE #1183: Adding events per drag and drop doesn't work
Probably something blocking the mouse.

Jowan-Spooner:
Should fix #1183


coppolaemilio:
That works! There is one bug still to fix but I'll create a separate issue. Thank you! it was such an easy fix I question my ability to code 💩

coppolaemilio:
The issue: https://github.com/coppolaemilio/dialogic/issues/1257

PR #1255, opened 2022-10-19
Timeline Saving Fixes (Again)

thebardsrc:
Timelines started blanking out again, so this is to fix that issue. Also found a few other issues with saving as well, as well as loading, so I'll get them fixed in here.

It appears that part of the problem, and the solution in this initial commit, might actually be a Godot bug with the Enum's not acting properly. I'll have to play with it more to try and figure that out. So that fix might be temporary.

### DIFF
--- a/addons/dialogic/Resources/TimelineResourceSaver.gd
+++ b/addons/dialogic/Resources/TimelineResourceSaver.gd
@@ -29,6 +29,7 @@ func _save(resource: Resource, path: String = '', flags: int = 0) -> int:
 			print('[Dialogic] Saving timeline...')
 			if FileAccess.file_exists(path):
 				var file := FileAccess.open(path, FileAccess.WRITE)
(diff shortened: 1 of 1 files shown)


thebardsrc:
Ok, so the reason the #1247 added the line was so new files could be saved on timeline creation, because it skips creating them when starting it because timeline save is called when starting a new timeline too (for some reason idk Godot resources thing). And it was doing that because there was an if statement to make sure the file existed before saving to it.

But that's pretty much left over from an earlier fix for the issue of timeline being erased. With all the other checks in place, there's no reason it should get that far and *not* want to save to a file, so there's no reason to check if it already exists anymore. So I've redone it to remove that if statement. Instead it adds one final check before opening the file, so if for some reason the actual string itself that's being built with the text timeline, thats supposed to be written to the file, is completely empty, it won't open the file and clear its contents. 

Gonna check the other file writing routines to make sure they're all safe too then the PR will be ready

PR #1262, opened 2022-10-22
Many UI changes for events

Jowan-Spooner:
Themeing
- most UI elements now have the godot4 LineEdit styleboxes for normal and focus
- I've reduced the line length for end Branch events, we can discuss this later. I wanted all the events to be shorter so this one had to be shortened too

Custom Number field
- smaller, good style
- saves not only on enter
- should allow implementing string editing in number fields later
-> better Vector field for Position event

Complex Picker
- now allows to navigate suggestions with up and down keys
- now has a focus style 
- better aligns icon
- better supports focus enter and exit
- suggestion function is now passed as a callable not an array (before: [object, method_name])
    -> this has resulted in a change in many events, some settings, the character editor as well as the toolbar (everything that uses ComplexPicker)
- suggestion function now should not filter the results anymore, this is done by the complex picker itself. The input is still passed, so it can be used in special cases (e.g. Value field in Conditions).

Misc
- some print statements where removed

Event field changes
- many texts in front of fields have been removed/changed. Usually the header should read like a sentence or contain very little, while the body contains settings more like a list.
- some fields where moved from header to content or vice versa. Only basic information should be in the header for a clean timeline.
- some events are now folded by default


coppolaemilio:
Wouldn't it be better to have the x and the y to live outside of the editable field?
image

coppolaemilio:
I really like all the changes 👏

PR #1267, opened 2022-10-24
Timeline Jump Stack

thebardsrc:
Adds a stack for timeline jumping, and option on the Jump event to return to the timeline after completing the timeline jumped to. Once reaching the end of a timeline, it will check the stack, and take the top one off if there's still more on the stack, and only call end_timeline() once the stack is empty

For example, take the following three timelines:
image
image
image

Running the timeline from sub1.dtl will then have the timleine events play in this order:

base 1

base 2

sub 1

sub 2

sub sub 1

sub sub 2

sub sub 3

sub sub 4

sub 3

sub 4

base 3

base 4

### DIFF
--- a/addons/dialogic/Events/Jump/event.gd
+++ b/addons/dialogic/Events/Jump/event.gd
@@ -6,10 +6,13 @@ class_name DialogicJumpEvent
 var Timeline :DialogicTimeline = null :
 	set = _set_timeline
 var LabelName : String = ""
+var Return: bool = false
 var _timeline_file: String = ""
 var _timeline_loaded: bool = false
 
 func _execute() -> void:
+	if Return:
+		dialogic.push_to_jump_stack()
 	if Timeline and Timeline != dialogic.current_timeline:
 		#print("---------------switching timelines----------------")
 		dialogic.start_timeline(Timeline, LabelName)
@@ -51,6 +54,7 @@ func get_shortcode_parameters() -> Dictionary:
 		#param_name : property_name
 		"timeline"	: "_timeline_file",
 		"label"		: "LabelName",
+		"return"	: "Return"
 	}
 
 func load_timeline() -> void:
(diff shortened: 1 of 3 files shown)


addons/dialogic/Events/Jump/event.gd
```diff
@@ -6,10 +6,13 @@ class_name DialogicJumpEvent
 var Timeline :DialogicTimeline = null :
 	set = _set_timeline
 var LabelName : String = ""
+var Return: bool = false
```
coppolaemilio, line 9: I think we should rename this variable to `can_return` or something like that, otherwise it is too close to the regular `return`

addons/dialogic/Other/DialogicGameHandler.gd
```diff
@@ -132,6 +137,18 @@ func jump_to_label(label:String) -> void:
...
 	current_event_idx = idx
 
+func push_to_jump_stack() -> void:
+	var current_point:Dictionary = {}
+	current_point['timeline'] = current_timeline 
+	current_point['index'] = current_event_idx
+	timeline_jump_stack.push_front(current_point)
+
+func pop_from_jump_stack() -> void:
+	var stack_point = timeline_jump_stack.pop_front()
```
coppolaemilio, line 147: I assume this should be `:Array` as well?

addons/dialogic/Events/Jump/event.gd
thebardsrc: Yeah good point on renaming it

addons/dialogic/Other/DialogicGameHandler.gd
thebardsrc: Ah no whoops that should actually be Dictionary, because it's the dictionary at [0] of the array

thebardsrc:
fix for #1270 by just removing the `duplicate()` from  `get_event()`. nothing using this function has any reason to actually duplicate the event

thebardsrc:
back out of draft now with the bug fix included

Issue #1239, opened 2022-09-25
Adding new timelines is unbelievably slow at scale


Tummyache:
## The problem

**Describe the bug**
Adding or removing timelines gets excessively slow after a large number of timelines (over 250) have been added.  Currently it's taking over 10 seconds to add or rename a timeline.  When adding a new timeline, Godot completely freezes and my CPU spikes from <1% to over 10%.

**To Reproduce**
1. Create 5 Folders
2. Create 10 Folders inside of each of those folders
3. Create 5 Timelines in each of the subfolders

**Expected behavior**
Adding a new timeline should be nearly instant, at least under a second.

**System (please complete the following information):**
 - OS: Windows
 - Godot Version: 3.4.2
 - Dialogic Version: 1.4

**Workaround**
I've just been cramming more dialog into each timeline and then using condition branches to separate the content.

**Possible fixes**
The problem appears to be tied to the plugin restructuring the entire project every time a file is added/renamed.  Oddly enough, deleting a timeline isn't slow at all.

zaknafean:
VERY big issue in 1.4.x as I'm finding myself in my current project. It's mostly due to how lookups are handled during creation. I think @thebardsrc had some good insight into the problem and fixes for the problem.

Tummyache:
I think in the meantime I'm just going to make a quick little separate application to create new empty timeline files and sync them with the folder_structure file.  Then disable timeline creation/renaming in the plugin code so I don't accidentally trigger the lookups.  That way I can at least close Godot, create all the files I need, and open Godot back up to generate the lookups or whatever is causing it.  One big load when I open the project is a lot better than a big load every time I accidentally double click a timeline.

Fixing it in the editor would be preferable, I just don't have the time to go digging around in the source.  So if there's any insight on where the problem is approximately, I could look at trying to help out.

Tummyache:
Cool, I'll keep an eye out and just roll with my hacky workaround until then, no rush.

zaknafean:
Honestly, if you get the work around going post it here. My team
mates would love to use it too.

Tummyache:
@zaknafean just finished it, haven't had a ton of time to test it.

[DialogStopgap](https://github.com/Tummyache/DialogStopgap)

[...]

PR #1271, opened 2022-10-27
1.x Speedups 

exelia-antonov:
Various fixes I've had in mind for improving Dialogic 1.x speed, which I had initally put aside when the Dialogic 2 development started but they're good for long term maintenance for people that for one reason or another can't move off of Godot 3


thebardsrc:
First improvement: 

- Dynamic cache for the built-in events. will load them only as used, then keep them in memory and pull them from the cache as needed. doesn't cache the custom events, just because they could changed much more commonly then the Dialogic core events

thebardsrc:
starting to address #1239, doing some initial observations as I try and step through from that. I've added timestamps to `build_resource_folder()` in the MasterTree.gd, since i gotta work backwards through all the steps to figure out whats happening where before I can do the full improved internal timeline/variable address dictionary changes

Umm

image

creating one single timeline folder *probably* shouldn't  be calling `build_resource_folder()` a total of **22** times...

`

thebardsrc:
wait... omg... it's calling that for *every* folder... the count of calls goes up for each new folder created. no wonder it keeps getting slower and slower!

thebardsrc:
Further observations tonight: every `update_folder_structure()` first loads all the firles, loads the folder strucutre JSON, does the recursive building, and then saves the JSON 

image

with the comments syaing 'when files got deleted and on program start', its called once in `_ready()` on the main editor, once in `_on_RemoveConfirmation_confirmed()`... then once in each of the build functions: `build_timelines()`, `build_characters()`, `build_definitions()`, `build_themes()`

the read in first before saving seems a bit redundant to me, so might readjust that. really should only _genuinely_ happen in the editor on startup, and all the other functions be write-only.

thebardsrc:
note to myself since I need to leave: the folder properties metadata isn't included in how this is built at the moment. needs to be added for Editor to use both for folder colors, as well as for preserving empty folders. can then be stripped out later for the runtime dictionaries because they don't matter there. also will need to add some additional sorting properties for editor, for when its exploded back into the folders.json organization

might be possible to add finer drag and drop sorting into the folder list with what I have in mind, though, so you can actually rearrange timelines in the same folder

thebardsrc:
ok FINALLY, lol. was really busy and couldnt focus a lot of time, but also recursion suuuuuuucks lol

image (4)

now that the framework for both reading in to new file structure, then writing back again to folder_structure.json, I can clean it up and start hooking all the rest of the parts of 1.x into it

Tummyache:
I just made a maze generator with path finding, so I feel that recursion hate.  Can't wait for the fix, really appreciate the effort!

thebardsrc:
ok hmm trying to put the rest of what's in DialogicClass.prepare() (which is where I was doing it only temporarily for step testing) into DialogicResources.get_resource_folder_flat_structure() makes the whole thing crash with cyclic errors, which seems to be a static typing issue with GDScript in 3.x, gotta think through that a bit differently then

thebardsrc:
reminder to myself to check what happens here when you actually have no name in the timeline file 

image

thebardsrc:
todo:
make new files go down to the first item in the folder, so it doesn't appear above subfolders (it rebuilds on project reload to be in that position)

thebardsrc:
reopened the bug #1278 just so i can do a verify check to make sure that works propelry with these changes

thebardsrc:
need to be able to reverify colors in Characters if somehting changes, becuse otherwise it'll revert until full reload

thebardsrc:
reminder to myself that I will need to add some checks and documentation because this introduces the minorest of breaking changes that forward slashes will be disallowed in folder and file names and need to be stripped on names and renames - if they aren't already? might be alrady idk I'll need to check

thebardsrc:
todo: make sure loading saves works

thebardsrc:
current things needing to be addressed before this can be merged:

- [ ] new definitions/glossaries are not being added to subfolders, but to root
- [ ] correcting where new files of each category go if the new timeline/character/definition/glossary/theme button is pressed when that's not the current highlighted tree
- [ ] the warning about forward slashes in names of folders and files, and the stripping of them for the breaking change 

current things that would be nice too, but I'm not marking them as blocking:

- [ ] the dropdown menus in editor for timeline events to use the new tree instead of older recursive tree
- [ ] trying to restore jumping to the right position in the tree on new item creation, and automatically going into rename mode
- [ ] restoring the actual event parts cache, since that was the beginning of this patch but something weird with instancing in Godot 3 makes it preserve values on reinstancing (I'm using the exact same dang method in Godot 4 for the Dialogic 2 event cache, and it doesnt do the same thing there??)

thebardsrc:
TODO: since I found a much simpler way of doing the dropdown menus, gotta do it for the other pickers too

goblinJoel:
Not sure how actively this is being worked on right now, or if I'll be upgrading to 2.x anyway before I finish development, but:

I noticed that 1.4.5 is loading every character file, apparently from disk, every time any character is looked up. (`DialogicUtil.get_character_list()`, which is called by `get_character(id), get_sorted_character_list(), and get_characters_dict()`. Does this PR already address that?

PR #1275, opened 2022-11-01
Translation updates

### LINKED ISSUE #1063: Alternative text for disabled choices
- ported from #316

It would be great to be able to add an alternative text for when a choices condition isn't met and it's disabled.

### LINKED ISSUE #1171: Allow translating multiple properties per event
Currently only one property can be translated per event. This is not good.
Instead I suggest to have the ability to store as many properties as wanted in the translation sheets, possibly like this:

```
| 2923472947_text 		|  Hi, I'm Jowan.        
| 1317932873_choice		|  YES.
| 1317932873_choice_alt		|  YES (to little confidence).
| 412312313_promt 		|  Please enter a name:
| 412312313_default 		|  Emilio
```
*Note the same id on multiple of these*

This could be very useful for localizing events with more then one text property.
E.g.:
- Input Event
- possibly Choice (if an alternative text is implemented)
- other custom events

### LINKED ISSUE #1264: Fix Translation System
- [ ] First of all it should work again at all. I think it's somewhat disabled right now.
- [ ] Multiple strings should be translatable per event. This will need some working.

Jowan-Spooner:
Hooray! Lot's of changes!

# Translation Settings Page Update
- page layout was simplified
- there is a lot of helping text again.

# New translation id/methods
- translation id's are now counted up hex numbers (New method DialogicUtil.get_next_translation_id())
- events can now register multiple properties for translation
- in the csv files keys are now a combination of event name, translation id and property name (e.g. `Text/3a/text`, `Choice/2f1/disabled_text`)
- translation settings have been moved into the dialogic/translation subcategory in the project settings (e.g. `dialogic/translation/enabled`)

## Translation methods
The event class now has the following methods for translation:
***To be overwritten in (translatable) event scripts:***
    -  _get_translatable_properties() -> Array
    - _get_property_orginial_translation(property_name) -> String

***To be used in (translatable) event scripts:***
    - get_property_transated(property_name) -> String

***Other translation methods (only used internally):***
     - add_translation_id() -> String
     - remove_translation_id() -> void
     - can_be_translated() -> bool
     - get_property_translation_key(property) -> String

***Misc:***
      - update_text_version() -> void ## I've added this method as I had trouble with saving because saving uses the event_node_as_text property that doesn't auto-update when changing properties of events. I hope @thebrd or I can make another saving simplification now that translation has been moved out of the TimelineSaver, so maybe this might be changed later.

Text event and Choice event/Subsystem have been updated to work with the new translation methods.

# AutoTranslation updates removed
Translation files are not auto-updated whenever timelines are saved now. This makes the whole translation and saving process a lot simpler. I'm pretty sure just using the buttons is enough.

# Misc changes
*These are not really related to translations, so I hope it's okay that they are in here*

    - timeline._to_string() changed to show the path instead of the instance id as it is much more insightful
    - complex picker now also searches the tooltip of each item
    - file picker now allows to set the mode to FILE_MODE_OPEN_DIR. In that case it will not only show the file name oc
    - Added DisabledText option to the Choice event. This text will be shown when the condition is wrong and else mode is "disable". This needed some change in the syntax of the Choices system. It is now:
`- Choice Text [if Condition] [else="disable" alt_text="DisabledText"]`
     The last part is basically a regular shortcode. I've done this more to test multiple translatable properties on one event (so disabled text is translatable as well).

# New workflow:
*The text below is also present in the translation settings tab. I've pasted it here for anyone wanting to understand how to work with translations/ the menu now.*

**Setting up translations:**
```
1. Enable translations
2. Set default locale (the language you used) and file mode. For per-project-mode also select a folder. In per-timeline mode files will be put next to timelines (and need to stay there!).
3. Hit "Update CSV files". This will create new csv files containing all translatable strings from your timelnes as well as one for the character names.
4. Wait until reimport is finnished. Then click "Collect translations"
```

**How to work with translations:**
```
Editing:
You can edit the csv files in most table editors (like google sheets or libre office calc). You can add new languages by creating a new column with a valid locale in the first cell. After doing this you will have to hit "Collect translations" again.

Updating:
If you make changes to your timelines after having created translations, just hit "Update CSV files" again, it will not remove changes already made to the file.

Testing:
You can test in different languages by editing the test_locale setting in project settings (Advanced Settings>Internationalization>Locale).
```

- Should fix #1264 
- Should fix #1171 
- Should fix #1063


coppolaemilio:
- [ ] Settings don't save. After changing something like `Per Timeline` to `Per Project`, or  selecting a translation folder, after closing and opening it again things revert to the default state

PR #1281, opened 2022-11-04
Add sidebar to editor

Jowan-Spooner:
This sidebar show all the latest resources and allows to open them. It uses the lookup from character- and timeline_directory so similar named files will show more to reveal which one they are.


coppolaemilio:
I can't seem to remove this extra size on this side of the handle:
image

coppolaemilio:
The flow is so much nicer now!
I noticed that the toolbar still has the resource picker. I think it should be removed now that we have the sidebar, but I know it is integrated with the "is saved" system. Not sure how simple it would be to remove it.

PR #1325, opened 2022-11-10
Move voiceover play activation to dialog effects

KvaGram:
Currently a draft.
General idea is to disconnect the voice events from text events, while also giving users more control of voiceovers.


KvaGram:
Voice event is now independent from text event. Voiceover data is saved in the voice subsystem when set, and can be used by any text event.

Text events must now use a text event [voice=(index)] to play a voiceover, where the index is an audio region specified in the Voice event.
The index counts from 0.

KvaGram:
Why this is still a draft:
The new voice effect has not yet been integrated into documentation and the timeline editor's support features.

KvaGram:
It is now possible to import audio regions from audacity label format.
Simply open the label track export from audacity in a text editor, and copy the text to your clipboard.
in Dialogic, open your Voice event, hit Load from Clipboard.

Issue #1338, opened 2022-12-02
History Row does not resolve BBCode [br]


eaganlai:
As shown below.

![image](https://user-images.githubusercontent.com/74284983/205314491-a29af649-daf5-449b-8355-f58df4e1d324.png)

The [br] is displayed correctly in the dialogue box.

![image](https://user-images.githubusercontent.com/74284983/205314128-a738ba82-f87f-42f1-a1c9-054648aaa4d9.png)

eaganlai:
Can be fixed by editing History.gd Line 224:
`		newHistoryRow.add_history(str(characterPrefix, eventData.text), audioData)`
to
`		newHistoryRow.add_history(str(characterPrefix, eventData.text.replace('[br]', '\n')), audioData)`

zaknafean:
Good catch. I always forget we manually do the [br] in the display parser and forgot to do that in the history parder.

zaknafean:
This is merged with #1339 Thanks for the contribution!

PR #1339, opened 2022-12-02
Fix History.gd [br] resolve.

eaganlai:
For issue https://github.com/coppolaemilio/dialogic/issues/1338


zaknafean (approved):
This is necessary since the [br] is handled in the display parser of text bubble, and was left out of the history parser.

coppolaemilio:
Thank you both!

PR #1341, opened 2022-12-04
Fixed Speed tag resolve in HistoryRow

eaganlai:
Fixed Speed tag resolve in HistoryRow with Regex. Also refactored [BR} replacement for better readability.


eaganlai:
https://github.com/coppolaemilio/dialogic/issues/1342

Issue #1331, opened 2022-11-14
SCRIPT ERROR: Invalid call. Nonexistent function 'is_tool' in base 'Nil'


bitbrain:
MRP: [dialogic-errors.zip](https://github.com/coppolaemilio/dialogic/files/10007718/dialogic-errors.zip)

```
SCRIPT ERROR: Invalid call. Nonexistent function 'is_tool' in base 'Nil'.
          at: update_preview (res://addons/dialogic/Editor/CharacterEditor/CharacterEditor.gd:296)
```

Steps to reproduce:

1. Download MRP
2. Move Dialogic into addons folder
3. do this:
![dwarf3](https://user-images.githubusercontent.com/822035/201787068-34e14563-0ca3-4165-878e-194a5106cdae.gif)

coppolaemilio:
do this 👁️ 👄 👁️

Jowan-Spooner:
Sounds like you are using a custom portrait scene that is not in tool mode. Shouldn't produce an error, but I think if you add @tool at the top (of the portrait scene script) it should fix it.

bitbrain:
@Jowan-Spooner I was able to fix it by doing so but I'd still consider this a bug - Dialogic should handle such case.

Jowan-Spooner:
Absolutely! just wanted to give some short-term help.

JamesMWestbrook:
https://github.com/coppolaemilio/dialogic/commit/2ef9b59315f10bcdcaa36ef3814988d752c48d34
updated both lines to first check that there is a valid instance of a script before checking for if a script is **is_tool**, error no longer occurs 
`if is_instance_valid(scene.get_script()) and scene.script.is_tool():`
`if  is_instance_valid(scene.get_script()) and scene.script.is_tool() and scene.has_method('_get_covered_rect'):`

Still very much new to contributing to public repos so wanted input before I make the pull request

Jowan-Spooner:
If it solves the error, certainly make a PR if you can! Thank you very much for your effort already.

Jowan-Spooner:
Should be fixed by #1354

PR #1354, opened 2022-12-12
Check for script before checking is_tool

JamesMWestbrook:
To address SCRIPT ERROR: Invalid call. Nonexistent function 'is_tool' in base 'Nil'. when using a custom scene that doesn't have a script as a portrait
to address issue https://github.com/coppolaemilio/dialogic/issues/1331


Jowan-Spooner:
Thanks!

Issue #1174, opened 2022-08-28
Add a button to edit the current character selected from character events.


coppolaemilio:
When you are on a timeline and you want to quickly jump to edit a character you have selected in a Character event, there is no easy way to go there, you have to search for the file in the filesystem tab and open. There should be an easier way to start editing it from the timeline editor.

thebardsrc:
I'll get this along with my changes to implement #1058, since I gotta make changes to the picker for that anyway

Issue #1169, opened 2022-08-26
Improve Character Event "Leave All" option UI


Jowan-Spooner:
Currently to make all characters leave, you have to toggle a hard to find toggle and not select a character. ALL shoukd be an option on the dropdown.

thebardsrc:
IIRC, "Leave -All-" is the actual text command for it. It sets the boolean of `_leave_all` on parsing the script, the boolean's not meant to be used directly. But I also couldn't get '-All-' to appear in the Charcater picker

thebardsrc:
Adding the -All- is easy enough now I could do it, but I'm not gonna close this out yet because I still want to do some rework of ComplexPicker before release, so I'll wait to add that in along with that

Issue #1149, opened 2022-08-24
Add Shortcut from Variable event to VariableEditor


Jowan-Spooner:
This means
- adding some methods for opening specific tabs of the settings editor
- adding buttons to events and connect them to methods

Issue #1114, opened 2022-08-14
Glossary Editor: Add warning for items with same name


Jowan-Spooner:
If you create two entries with the same name, after closing the settings dialog, they first one will get overwritten by the second's entry. Not sure if there is an easy way to identify if the entry already exists but I would try to either warn you about it, or add `_1` on the new one. It is one of those annoying things to fix, but it would suck to lose some data because of unknowingly naming two entries the same.

_Originally posted by @coppolaemilio in https://github.com/coppolaemilio/dialogic/issues/1111#issuecomment-1214426154_

PR #1355, opened 2022-12-14
Multiple small improvements & Autocompletion for text editor

Jowan-Spooner:
# Autocompletion
# Auto-completion for text editor
Adds autocompletion to the text editor.
The following things are suggested:
- shortcode identifiers
- shortcode parameters
- if shortcode parameter has a suggesions function setup, those will be suggested as values
- Character names
- Character Portraits
- Text Effects
- Join, Leave, Update, VAR, if, elif, else
- Variable Names

## Paramter Value suggestions
For many events some value suggestions have been added:
- true and false for boolean values
- audio bus's for bus
- timelines for jump event
- enum values for History and Position event

# Smaller fixes for:
- #1174
- #1169
- #1149
- #1114
- #1069
- Variable Subsystem incorrectly named Variables instead of VAR (broke Set variable event and other variable systems)
- make Condition (if, elif, else) End UI buttons work again
- Implement Buttons on events (Character Event and Set Variable Event)
- Next Indicator was placed wrong


coppolaemilio:
I'm so excited by this one, it is more than I would have imagined to have in Dialogic 2 🤩 

Thank you Jowan!

PR #1399, opened 2023-01-10
subsystem_portraits: Don't return null in functions that return void

vmedea:
This causes errors in current godot master.


coppolaemilio:
Thank you!

PR #1400, opened 2023-01-12
Fix a few editor startup errors

vmedea:
Fix a few SCRIPT ERRORs at startup:

- Don't return value from create_drag_and_drop_event, which returns void
- Fix Category enum value MAIN -> Main
- Update set_drag_forwarding and associated signatures: There were some changes to set_drag_forwarding in Godot 4:
  - `Control.set_drag_forwarding` takes three callables instead of an object
  - `_can_drop_data_fw`, `_drop_data_fw` don't have a `from_control` argument anymore

### DIFF
--- a/addons/dialogic/Editor/Events/Fields/FilePicker.gd
+++ b/addons/dialogic/Editor/Events/Fields/FilePicker.gd
@@ -31,7 +31,7 @@ func _ready():
 	%ClearButton.icon = get_theme_icon("Reload", "EditorIcons")
 	%OpenButton.button_down.connect(_on_OpenButton_pressed)
 	%ClearButton.button_up.connect(clear_path)
-	%Field.set_drag_forwarding(self)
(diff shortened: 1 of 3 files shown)


coppolaemilio:
Thank you so much :)

coppolaemilio:
Sorry, I had to revert this changes, I thought I tested them before but I wasn't in the correct branch. With your changes I get these kind of errors:
`  res://addons/dialogic/Editor/Events/Fields/FilePicker.gd:34 - Invalid call to function 'set_drag_forwarding' in base 'LineEdit'. Expected 1 arguments`

vmedea:
Oops, sorry! I wonder when that changed, the ["latest"](https://docs.godotengine.org/en/latest/classes/class_control.html)  documentation does show the function with three arguments, didn't expect it to just change inbetween like that (I'm using godot master because I'm testing some things). Will try to check next time. In any case, you'll likely need the change at some point.

PR #1401, opened 2023-01-14
Visual timeline loading speedups

Jowan-Spooner:
WIP pr, that tries to speedup the loading of visual timelines. 

While still far from optimal, this PR makes it so the body of events that have their body hidden by default is only build, once that body is opened. 

Because the adding of nodes to the scene tree is the most intensive action during the loading this already speeds it up, especially for events with many body settings (like character, position and choice). More could possibly be removed by using draw commands (e.g. for icons) instead of nodes, especially on the Complex Picker (which is used in the condition picker for example).


Jowan-Spooner:
Would appreciate if someone with a real project could test this and let me know if the experience has improved for them, when opening long timelines.

PR #1422, opened 2023-02-14
Fix default layout path

QuasiStellar:
In https://github.com/coppolaemilio/dialogic/pull/1421 `DialogicDefaultLayout` (formerly `DialogicDefaultScene`) was moved and renamed, however paths pointing to its location were not updated, because GDScript is a very good language and refactoring code written in it is always a pleasure.

There are some occasional formatting fixes to dilute the atmosphere of inevitable hardcoded doom.


Jowan-Spooner:
Thanks for this PR. 
However, we can now use DialogicUtil.get_default_layout() to get this path. The methods in settings general seem like they are not needed anymore.

QuasiStellar:
Updated and tested with the same usecase.

Jowan-Spooner:
Thanks!

PR #1437, opened 2023-02-21
Remove audio region

KvaGram:
Also removes placeholder framework for multible text lines in text event.

Kinda sad to see work I have done being removed, but this is how it must be.
And as such, I want to do it myself.


Jowan-Spooner:
Thanks! I hope you can use some of this work on your extension!

PR #1461, opened 2023-03-08
Simplify and fix History system 

Jowan-Spooner:
This update fixes all the previous issues with the history system such as the button not hiding, crashes on character leaves. 

It's also simplifiying the way the history system works.
There is now three types of history the history system can save:
- Simple history
    - events can log information here, that can be used to display a backlog
- Full Event history
    - stores a copy of all visited events in an array
- Already Read history
    - remembers what text events you've already read and emmits signals to use that info

Splitting these systems makes the code much cleaner and the individual goals easier to maintain. 

The full event system might currently not be very useful. It wasn't used for anyting really before, so I just kept a very bare version of it. It can be expanded in the future, if it's useful for something.

Neither of the histories are currently saved.


addons/dialogic/Events/Choice/event_choice.gd
```diff
@@ -27,6 +27,12 @@ func _execute() -> void:
 	# This event is basically a placeholder only used to indicate a position. 
 	# It is never really reached. Instead the Subsystem_Choices queries the events 
 	#   to find the choices that belong to the question.  
+	if dialogic.has_subsystem('History'):
```
thebardsrc, line 30: so the comment above says this one isn't actually real, its just a placeholder? so either that needs to be changed, or this code here doens't actually do aything

addons/dialogic/Events/History/subsystem_history.gd
```diff
@@ -3,89 +3,85 @@ extends DialogicSubsystem
 ## Subsystem that manages history storing.
 
 
-signal history_text_already_read()
+## Simple history that stores limited information
+## Used for the hitstory display
```
thebardsrc, line 7: minor documentation typo

addons/dialogic/Events/History/subsystem_history.gd
```diff
@@ -3,89 +3,85 @@ extends DialogicSubsystem
...
+## Full event history (can be used for undo)
+var full_event_history_enabled := false
+var full_event_history_length := 50
+var full_event_history_content := []
+signal full_event_history_changed
+
+## Read text history 
+## Stores which text events and choices have already been visited
+var already_read_history_enabled := false
+var already_read_history_content := []
```
thebardsrc, line 21: the already read history, specifically, should be a Dictionary, not an Array. it's done by unique key lookup, and it will be much slower the longer the array, whereas in a Dictionary it remains quick as it does lookups by hash instead of iterating the array. as this is meant to have every text event in an entire game eventulally, at 100% completion, this will very quickly get longer and longer 

a quick performance example, of just raw code doing a lookup on an array and a dictionary of 50000 items, filled with a string + number. code for creation and search functions included:

image
image


image

image

Jowan-Spooner: Changed to a dict.

PR #1492, opened 2023-03-30
Changed Glossary Category

mechPenSketch:
This PR sets `event_category` in `DialogicGlossaryEvent` to `Other` because the previous value, `AudioVisual`, is not found in `DialogicEvent.Category`.

I originally presumed `Other` to be the category, but it could be `Audio` instead. Let me know the intended category, and I'll make a new PR branch.


Jowan-Spooner:
Other is correct. I made this change too already for my local version but didn't push yet ;) Thanks!

Jowan-Spooner:
The glossary event isn't even enabled right now, as it does nothing, so this really is just an annoyig error message.

PR #1495, opened 2023-03-31
Dialogic 1.x Converter minor fixes

metanoder:
Converter has minor errors that keep it from running.  This will at least get the converter to run again. 

*Caveat
if you had variables as names in Dialogic 1.x,
there's a conflict with the auto color: subsystem_text.gd 

### func color_names(text:String) -> String:
   ...
   **for result in color_regex.search_all(text):**
      ...

Workaround if you want to keep variables as names...
1. go to : Project Settings...
2. turn on : Advanced Settings...
3. scroll down to : Dialogic/Text
4. turn off : Autocolor Names
5. lastly, back in character profile:  use {} instead of [] in name

Else: currently, you have to convert your variable names into manually typed static names.


metanoder:
Further testing triggered the debugger.  
Looks like the timeline conversion script needs to be debugged.  

Moving this to "draft" until it's sorted out.

thebardsrc (approved):
looks good to me!

PR #1505, opened 2023-04-09
Make Sub Directories if Not Already Existed

mechPenSketch:
This prevent "directory" from being set to null upon calling "DirAccess.make_dir_absolute(dir)".

Also, make "refresh()" to be called on itself rather than its parent.


coppolaemilio:
Thank you :)

PR #1506, opened 2023-04-09
add null check to text update_name_label

oddfacade:
There are a few places where we do a `update_name_label(null)`, which makes `character.resource_path` throw an invalid index error. This commit fixes this bug; also a fixes a related bug in the same function where it is possible to access
`dialogic.current_state_info['character']` before it is defined.

### DIFF
--- a/addons/dialogic/Events/Text/subsystem_text.gd
+++ b/addons/dialogic/Events/Text/subsystem_text.gd
@@ -87,11 +87,12 @@ func _on_dialog_text_finished():
 
 
 func update_name_label(character:DialogicCharacter) -> void:
-	if character.resource_path == dialogic.current_state_info['character']:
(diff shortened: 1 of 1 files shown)


addons/dialogic/Events/Text/subsystem_text.gd
```diff
@@ -87,11 +87,12 @@ func _on_dialog_text_finished():
 
 
 func update_name_label(character:DialogicCharacter) -> void:
-	if character.resource_path == dialogic.current_state_info['character']:
+	var character_path = character.resource_path if character else null
```
coppolaemilio, line 90: can you split this into several lines?

coppolaemilio:
Merging for now since it fixes the issue :)

PR #1509, opened 2023-04-10
Check if Keys are Avaiable

mechPenSketch:
Otherwise, the value would have been taken null and the code won't progress further.

This is a fix to the timeline conversion.


coppolaemilio:
Thank you!

PR #1525, opened 2023-04-23
Make sure autoadvance setting works

Jowan-Spooner:
There was a confusion with the autoadvance setting being named autocontinue in some places and thus not properly working. 

It's now ALWAYS named autoadvance.


ronsiv8:
Thanks for the quick response!

coppolaemilio:
Thanks!

PR #1531, opened 2023-04-24
Add modules manager and first default override ability

Jowan-Spooner:
There is a new settings tab that allows managing all the modules and extensions. It has a fully functional tree list of all the modules and their parts.

There is a editor for the event defaults. If changed these will be stored in a event default override dictionary. From there they are automatically set in each event before the actual event is loaded. Shortcode events also take these defaults into account for saving (won't add a value if it's the custom default).

I had to make sure these custom defaults are also applied to new events in the visual editor (as they are not acutally "loaded" there if new).

You can also specify default values for the text event (character and portrait settings). 

For this to be fully functional, I had to add some syntax flavors. 

With no default character and portrait set:
`This is some text` will be said by noone
`John: Some text` will be said by John with his default portrait

With a default character set:
`This is some text` will be said by the default character
`_: This is some text` will be said by noone

If a default portrait is set for the event, it will have a higher rank then the default portrait of the character. To revert to the latter, you can write:
`John(): Some text` which will be said with Johns default portrait even if a default portrait is set for the text event.

Similar things are true for the character event. Some parameters cannot be edited for this one, as they clash with the already existing animation settings in the portrait settings tab.

# Other
- fixed capitalization in Sound and Music event

### DIFF
--- a/addons/dialogic/Editor/Settings/settings_editor.tscn
+++ b/addons/dialogic/Editor/Settings/settings_editor.tscn
@@ -1,8 +1,9 @@
-[gd_scene load_steps=4 format=3 uid="uid://dganirw26brfb"]
+[gd_scene load_steps=5 format=3 uid="uid://dganirw26brfb"]
 
 [ext_resource type="Script" path="res://addons/dialogic/Editor/Settings/settings_editor.gd" id="1"]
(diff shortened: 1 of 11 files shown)


coppolaemilio:
Thanks! Looks great :)

PR #1536, opened 2023-04-25
Stop First Event Skip

mechPenSketch:
When an action button is used to start a timeline, this PR will prevent the button press from skipping the first event by accident.

### DIFF
--- a/addons/dialogic/Events/Text/default_input_handler.gd
+++ b/addons/dialogic/Events/Text/default_input_handler.gd
@@ -10,13 +10,16 @@ func _input(event:InputEvent) -> void:
 	if Input.is_action_just_pressed(DialogicUtil.get_project_setting('dialogic/text/input_action', 'dialogic_default_action')):
 		if Dialogic.paused: return
 		
-		if Dialogic.current_state == Dialogic.states.IDLE and Dialogic.Text.can_manual_advance():
-			Dialogic.handle_next_event()
-			autoadvance_timer.stop()
-		
-		elif Dialogic.current_state == Dialogic.states.SHOWING_TEXT:
-			if Dialogic.Text.can_skip():
-				Dialogic.Text.skip_text_animation()
+		match Dialogic.current_state:
+			Dialogic.states.IDLE:
+				if Dialogic.Text.can_manual_advance():
+					Dialogic.handle_next_event()
+					autoadvance_timer.stop()
+			Dialogic.states.SHOWING_TEXT:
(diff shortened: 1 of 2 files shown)


addons/dialogic/Events/Text/default_input_handler.gd
```diff
@@ -10,13 +10,16 @@ func _input(event:InputEvent) -> void:
...
+		match Dialogic.current_state:
+			Dialogic.states.IDLE:
+				if Dialogic.Text.can_manual_advance():
+					Dialogic.handle_next_event()
+					autoadvance_timer.stop()
+			Dialogic.states.SHOWING_TEXT:
+				if Dialogic.Text.can_skip():
+					Dialogic.Text.skip_text_animation()
+			Dialogic.states.JUST_BY_ACTION:
+				Dialogic.current_state = Dialogic.states.SHOWING_TEXT
```
mechPenSketch, line 22: A disclaimer on this line: I've set this one as `SHOWING_TEXT` as my first event happens to be the text one. There might be a state to switch to that could encompass all event, but I don't know which.

Jowan-Spooner:
Hmm... I have to think about this for a moment to see if this is the best way. I don't particularly like using the idx_or_label argument as a way to pass this information. Will take a look though.

Jowan-Spooner:
Have you tried just setting the input as handled? Not sure how you start your dialog but in code like this it works for me:
```gdscript
func _input(event):
    if Input.is_key_pressed(KEY_ENTER):
        Dialogic.start("res://test-project/example-timeline.dtl")
        get_viewport().set_input_as_handled() # without this the first line is skipped
```

mechPenSketch:
I did not expect the alternative, but now that I've tried it, it works for me, too. Is there anywhere in the wiki that mentions this?

Jowan-Spooner:
Not in the dialogic wiki I think. There is lot's missing there...
Also it was considered multiple times to use the _unhandled_input() method instead of the _input() method so that clicking buttons while dialog is visible works without advancing dialog, but it has turned out complicated as it would require a clickable control for dialog advancing mouse-clicks to work.

mechPenSketch:
I see. I'm closing this PR now that it's no longer needed.

Jowan-Spooner:
Yeah, sorry. Thanks for the work you put in anyways! I'm very thankful you are testing things and try to find solutions. If you have trouble with other things, you can always create an issue first or ask in discord, just to see if anyone knows a solution.

PR #1542, opened 2023-04-30
Fix No One setting being ignored in text event

Jowan-Spooner:
When having set a character for the text event (in visual editor), setting it to No One again didn't work, because character_from_directory didn't deal with an empty string correctly.


coppolaemilio:
Thanks!

PR #1547, opened 2023-05-02
Add animation support

### LINKED ISSUE #1203: Textbox hiding and animations
## Hiding the textbox
Currently the textbox is always visible. It's not hidden when the timeline ends or a End timeline event is encountered. Also it's shown by default, not only with the first text event (I believe that's how it was in 1.x). 

## Textbox animations
It should be possible to add animations to the textbox for opening and closing.

Jowan-Spooner:
So this is a system where the animations are completely implemented in the layout scene by the user. All they have to do is to call Dialogic.Animation.now_animating() when receiving the signal and call Dialogic.Animation.animation_finished() once done. This way dialogic knows exactly when to continue.

I had to change some things. For example the textboxes now are hidden by default until some text is shown. This can be changed with a setting called "Hide textbox when empty"
This makes scenes with no text much easier. Currently the way to trigger this is to add a wait event that hides the text box. Then it will again only be shown once any text event updates it.

For all of this to work I had to make sure the clear method is called each time a dialog is newly instanced. Unfortunately that meant I had to specify to the clear method what to clear because Variables for example should NOT always be cleared. This introduced a ClearFlags enum on the dialogicgame handler. I hope it's a good addition. It could prove useful in the future.

- Fixes #1203


coppolaemilio:
Thanks!

PR #1548, opened 2023-05-02
Rework event dragging and moving

Jowan-Spooner:
- Allows moving all selected events up and down with ALT+UP and ALT+DOWN
- movement with ALT+DOWN and ALT+Up can now be undone
- Allows moving all selected events by dragging
- Changes the drag preview to not being a preview, but a line
- This new dragging is not blocked by event node fields

# Additional
- Adds an indicator for empty text events (an empty text event is saved as <Empty Text Event>)
- fix for the event body indent sometimes not being there


coppolaemilio:
Thanks!

Issue #1552, opened 2023-05-07
There is no way to remove Auto-pauses


coppolaemilio:
![image](https://user-images.githubusercontent.com/2206700/236705495-0e1a7127-9310-4ab7-a3ea-a0eb19cdd323.png)
Once you add them you can't remove them.

PR #1563, opened 2023-05-19
Addressing Issue #1552 remove auto-pauses

CristianRos:
Hey, this is my first PR so I'm a little bit lost, I hope everything it's okay, tell me if I need to change anything :)

I felt fairly confident I could help with this issue in particular. But as I was approaching it I thought it could've been done in different ways:

image

I went with the 'D' case, with the 'B' one being the second I like the most. I feel like if the remove button is to the right of the time input I would delete the auto-pause set trying to change the value with the arrows.

Also while addressing this I realized 2 things:

- When you add a lot of auto-pauses the UI breaks, I don't think anyone is gonna need that much but maybe is worth to know: 
image

- It only saves the auto-pauses when swapping the main tabs:
image

I'm not sure if that's already known but I wanted to say them in case they aren't

### DIFF
--- a/addons/dialogic/Modules/Text/settings_text.gd
+++ b/addons/dialogic/Modules/Text/settings_text.gd
@@ -88,7 +88,7 @@ func add_autopause_set(text:String, time:float) -> void:
 	%AutoPauseSets.add_child(hbox)
 	var remove_btn := Button.new()
 	remove_btn.icon = get_theme_icon('Remove', 'EditorIcons')
-	remove_btn.pressed.connect(_on_remove_autopauses_set.bind(hbox))
+	remove_btn.pressed.connect(_on_remove_autopauses_set_pressed.bind(hbox))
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
Hey, thanks for contributing. I think third option (text input, time input, remove button) makes the most sense. 

The autopauses only saving when switching tabs is by design but I would be fine with changing it. 

If you feel like it you can add a Scroll container into the settings page to keep it from breaking, but yeah, so many pauses are unlikely.

CristianRos:
I don't feel confident enough about touching the saving part, I mentioned it because I kind of thought it would save changes directly and I was struggling a bit thinking I introduced a bug or something like that haha

But I changed what you said, now the autopauses follow [text input, time input, remove button] and there is a scroll bar when you add a lot of them.

Hope that's okay!

Jowan-Spooner:
Looks great! Thanks!

PR #1568, opened 2023-05-21
Make event categories more flexible

Jowan-Spooner:
This makes event categories strings and dynamically adds the sections in the button list. This allows extensions to add their own sections.

It also adds a way to reorder the sections in the settings.


coppolaemilio:
+191 -1,335 and we get better functionality
feels gooood :)

Issue #1383, opened 2022-12-28
Autofocus text editor when Text event is created 


Jowan-Spooner:
Requested by a discord user.

If you create a text event in the visual editor, it would be nice if you could instantly start editing.

coppolaemilio:
I was going to create this one now 😅

PR #1573, opened 2023-05-22
Fix issue #1383: Generic autofocus on event creation with button

CristianRos:
I added this 3 properties to the event base class: `created_by_button`, `autofocus_field_index` and `autofocus_relative_path`.

I created a flag `created_by_button` so the `build_editor()` function in _**event_block**_ script knows when it's being created with a button or not. Setting it up to true before calling `add_event_node()` inside `_add_event_button_pressed()` in the _**timeline_editor_visual**_ script, and defaulting it back to false after the node is created.

Then when `build_editor()` is creating fields we check if the new one it's the right one by index and get it's path to grab the focus. We know which one it is because of `autofocus_field_index` (**-1** being the **default** index for none). Additionally if we need to access a child of the field we can provide a relative path with `autofocus_relative_path`

Now we can give autofocus to any event field with something like:
image
Which are the lines of code I added to the **_event_text_** script

If you want me to add autofocus to any other event, or to change anything else let me know!


Jowan-Spooner:
Looks great. Will try to test it later and merge it then!

Jowan-Spooner (approved):

coppolaemilio:
@CristianRos I love it! Thank you so much for the contribution 🥳

PR #1585, opened 2023-05-29
Make styles require style.cfg to be detected.

coppolaemilio:
Styles now require a `style.cfg` to be detected where you declare the basic information.
This is the first step for allowing for people to make their own layouts in a more "shareable" way. 

image


Jowan-Spooner:
Moved the scan() method into the index.gd class so that other modules (e.g. custom) can use it as well. Otherwise looks good and more self contained, always a plus.

PR #1586, opened 2023-05-30
Multiple fixes (all over the place)

Jowan-Spooner:
This is an unfortunate mess. I've been working on the example projects and making changes as needed along the way. This is the result of that.

Here is a list of the most prominent changes:

## Choices
- you can now choose how/when choices are revealed. either
    - instantly when the text finished (default, because previous behaviour)
    - after a certain time (very useful if text speed is instant (0))
    - after another input
    - after an input OR a certain time

## Input
- The input handler now emits a dialogic_action signal that can be listend to
- The input handler reference in the Text subsystem is now named input_handler (previously input_handler_node)

## Textbox
The DialogText node now has three new settings: `enabled`, `hide_when_empty` and `textbox_root`. 
- if enabled is false, this node will ignore calls to `reveal_text()` and the text subsystem will ignore the node
- if hide_when_empty() is true, the node will hide itself (without animation though) whenever set to an empty text
- textbox root determines what node should be hidden/revealed when the textbox shows/hides. It defaults to "self", but on the default layout it is set to the Panel containing the node.

- fixed a bug where text effects at the very end of a text event would be ignored

## Other
- Add hide_textbox setting to [await input] event 
- adds autocompletion for the wait events hide_text setting
- Make shortcode suggestions add a space if it has settings
- Add some static typing
- Removed the "Special" style from the default layout, as it was in bad shape and isn't really needed

### DIFF
--- a/addons/dialogic/Editor/Settings/settings_modules.gd
+++ b/addons/dialogic/Editor/Settings/settings_modules.gd
@@ -73,7 +73,7 @@ func _on_search_text_changed(new_text:String) -> void:
 		
 		for i in range(child.get_button_count(0)):
 			child.erase_button(0, child.get_button_count(0)-1)
-		var any_visible = false
+		var any_visible := false
 		var counter := 0
 		for sub_child in child.get_children():
 			if sub_child.visible:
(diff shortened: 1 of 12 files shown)


Jowan-Spooner:
Will merge this although it's messy. Hope nothing breaks

PR #1591, opened 2023-06-01
Portrait modes update (+other stuff)

Jowan-Spooner:
# Character portrait modes
... have been removed! Hurray!
Seriously tho. Instead portrait containers can now choose between two modes: Position and Speaker.
- POSITION mode is like previously. It has an index and can be joined/moved to. 
    - A character cannot be on multiple position containers at the same time.
- SPEAKER mode will automatically join the current speaker whenever there is one. 
     - speaker mode can have a portrait prefix setup. This allows using a different set of portraits for the speaker container (e.g. instead of "happy" a prefix of "Face/" would show the "Face/happy" portrait). 
-> this means both modes can now be used at the same time
-> it also means that this is now dependent on the layout scenes setup (no external setting)

To make this work I had to rewrite a lot of the portrait subsystems methods.
This means there is now a set of methods that allow manipulating a portrait based on the actual portrait node (not the character). There is also another set of methods for the position mode (VN-style portaits; used by the character event). These methods will make sure the character is joined and wont allow duplicate characters. A join_character(), leave_character() methods have been added to make character event functionality more accessable. The second set of methods uses the first under the hood (with extra checks and functionality around).

# Other changes
Along the way numerous other things have been fixed...
- bbcodes no longer mess up text effects
-  variable and glossary parsing has been moved into the parse_text() method
- text bubble overrides are now correctly applied again
- rpg layout scene has been reworked, removing the texture
- default layout portraits now have better anchors
- added helpful documentation comments to portraits_subsystem
- removed portrait mode setting from portrait settings
- added two preview portraits to the debug character (full character and speaker closeup)
- removed portrait holder (not sure why it was still there) 
- add text editor autocompletion for "Settings"


Jowan-Spooner:
I will merge this and continue on portraits for some more.

PR #1607, opened 2023-06-18
Fixes #1606 logcial_convert function bug fix

### LINKED ISSUE #1606: logical_convert: Cannot convert argument 1 from float to string
## The problem

**Describe the bug**
Error in class VariableFolder _get function:  Cannot convert argument 1 from float to string.

My variable is setup as a group with a number variable, the error occurs when trying to access the number through code: 
`Dialogic.VAR.GroupName.VariableName`

**To Reproduce**
1. Create a basic project with a character and timeline.
2. Add a variable group and subvariable set to 0 (Can also be done using an ungrouped number variable)
3. Add the following logic to main scene:
```  GDScript
extends Node3D

func _ready() -> void:
	print("Before Timeline:")
	print(Dialogic.VAR.Test.TestNum)
	Dialogic.connect("signal_event", _on_timeline_signal)
	Dialogic.start("timeline")

func _on_timeline_signal(data : String) -> void:
	print("After Timeline:")
	print(Dialogic.VAR.Test.TestNum)
```
4. Add the following to the timeline:
```
Join character
The test number will now be increased using Test.TestNum += 1 in dialogic
VAR {Test.TestNum} += 1
[signal arg="test"]
```
5. Run the scene.

**Expected behavior**
Retrieve the variables current value

**Screenshots**

 - OS: Linux
 - Godot Version: 4.0.3-stable
 - Dialogic Version: 2.0 alpha 8

## Solutions

**Workaround**

1. Tried using variables outside of groups with the same result.

**Possible fixes**

Code faulting out:
image

rassweiler:
- Change function parameter from String type to Variant type

### DIFF
--- a/addons/dialogic/Other/DialogicUtil.gd
+++ b/addons/dialogic/Other/DialogicUtil.gd
@@ -117,7 +117,7 @@ static func str_to_bool(boolstring:String) -> bool:
 	return true if boolstring == "true" else false
 
 
-static func logical_convert(value:String) -> Variant:
+static func logical_convert(value:Variant) -> Variant:
 	if typeof(value) == TYPE_STRING:
 		if value.is_valid_int():
 			return value.to_int()


coppolaemilio:
Thanks!

Issue #1609, opened 2023-06-20
Timeline Editor Visual causes errors when gamepad is connected


IkutiDev:
## The problem

**Describe the bug**
When I work with Dialogic 2 on Godot 4, if I have connected PS4 gamepad (guessing it's the same with any kind of gamepad), I get spammed with errors "res://addons/dialogic/Editor/TimelineEditor/VisualEditor/timeline_editor_visual.gd:318 - Invalid get index 'pressed' (on base: 'InputEventJoypadMotion')."
  

**To Reproduce**
1. Connect gamepad to your PC (I have PS4 gampead, xbox gamepad should be fine too I think)
2. Go to dialogic editor window
3. Go to timeline section
4. Get a spam of error messages about InputEventJoypadMotion

**Expected behavior**
No error spam when gamepad is connected

**System (please complete the following information):**
 - OS: Windows
 - Godot Version: 4.0.3
 - Dialogic Version: 2.0 Alpha 8

## Solutions

**Workaround**

Disconnect gamepad when working with Dialogic

**Possible fixes**

Add a check when testing if event is pressed, to only test it if event input is not a joypad motion event (could also just check for mouse events, both should be fine)

`

	func _input(event:InputEvent) -> void:
	# ...
	if get_viewport().gui_get_focus_owner() is TextEdit || get_viewport().gui_get_focus_owner() is LineEdit: 
		return
	if !((event is InputEventKey or !event is InputEventWithModifiers) and is_visible_in_tree()):
		return
	if not event is InputEventJoypadMotion:
		if !event.pressed:
			return
`

Jowan-Spooner:
I suppose I can close this now.

PR #1615, opened 2023-06-21
Fixed input bugs with event pressed and gamepad (#1609)

IkutiDev:
Fixes the issue with #1609 in a better way, by making sure we only check if event is pressed, if event has a pressed property (thus fixing it if someone has connected in gamepad or even possible issues if someone has connected in a midi input device)


Jowan-Spooner:
Looks good to me!

Jowan-Spooner:
Thanks for the fix!

danbolt:
Grateful to see this one fixed! Looking forward to it in the next release!

PR #1618, opened 2023-06-24
Fix extensions_folder typo

oddfacade:
The property is supposed to be `dialogic/extensions_folder`, plural and without the trailing `/`. This broke custom extensions folders.


Jowan-Spooner:
Very good catch! Thank you very much!

PR #1621, opened 2023-06-25
Broken Reference Fixer (alternative to #1605)

Jowan-Spooner:
This PR introduces an interface for fixing broken references. This interface allows listing any possibly broken references and searching for them in timelines and then replacing them. 
grafik

The variable editor will already register changes to this interface and display a warning with a button. 
grafik

Further would have to be done to register reference changes
- on portrait renames (pretty complex RegEx needed)
- character file renames
- timeline file renames

I also wanted to add a way of adding items manually to this list, in case dialogic doesn't correctly register something.

# Timeline processing changes
To make this easier I worked on restructuring timeline processing. If this PR won't make it in, I will at least extract that work, as it makes the whole process_timeline() stuff much more organized.


Jowan-Spooner:
This is ready for review and I would like @CristianRos to confirm/test it works as expected before merging it.

PR #1623, opened 2023-06-25
Fixed file picker in portrait scene exports

oddfacade:
The portrait export editor is unable to edit strings exported with the `@export_file` annotation because they weren't being correctly serialized with `var_to_str` before being stored in the tree item's metadata. The file picker is also ignoring submission events from its `LineEdit` child, which in this instance means the file picker's text box, despite being editable, can't actually be used to define the file.

This commit fixes both issues. However, the fact that these were seemingly special-cased indicates to me that there may have been a conscious decision to make them this way. I was not able to determine what the reasoning might have been, so I leave the question open for reviewers.


addons/dialogic/Editor/Events/Fields/FilePicker.tscn
```diff
@@ -89,14 +89,15 @@ theme_override_styles/read_only = SubResource("StyleBoxEmpty_6b7on")
...
 [node name="ClearButton" type="Button" parent="HBox"]
 unique_name_in_owner = true
 layout_mode = 2
-icon = SubResource("ImageTexture_d1x5q")
+icon = SubResource("ImageTexture_vy6bu")
 flat = true
 
 [connection signal="focus_entered" from="HBox/Field" to="." method="_on_field_focus_entered"]
 [connection signal="focus_exited" from="HBox/Field" to="." method="_on_field_focus_exited"]
+[connection signal="text_submitted" from="HBox/Field" to="." method="_on_file_dialog_selected"]
```
oddfacade, line 103: This is the actual change. I connected the `text_submitted` signal to the same callback used by the file selection dialog.

addons/dialogic/Editor/CharacterEditor/char_edit_p_section_exports.gd
```diff
@@ -97,7 +95,7 @@ func _on_export_number_submitted(value:float, property_name:String) -> void:
 	set_export_override(property_name, var_to_str(value))
 
 func _on_export_file_submitted(property_name:String, value:String) -> void:
-	set_export_override(property_name, value)
+	set_export_override(property_name, var_to_str(value))
```
oddfacade, line 98: This is needed because it's read back and deserialized using `str_to_var`, which will give you `null` if it gets an unquoted string (that doesn't correspond to some other type).

Jowan-Spooner:
Looks very good! Thank you very much!

PR #1638, opened 2023-07-19
Add skip delay

KoB-Kirito:
Adds a timer to the input handler that discards inputs every x seconds, which effectively delays the skipping of events and animations.
Adds config to configure that x delay. Delay is optional, 0.0 results in the default behavior.

Fixes problems with dialogic eating inputs or processing them twice sometimes, leading to accidently skipped events, especially on web builds. (Which I was too lazy to open an issue for)
Enables users to set a general global delay for skipping events.

Tested with some styles and configurations, but I'm not proficient with the dialogic structure.

### DIFF
--- a/addons/dialogic/Modules/Text/default_input_handler.gd
+++ b/addons/dialogic/Modules/Text/default_input_handler.gd
@@ -2,6 +2,7 @@
 extends Node
 
 var autoadvance_timer := Timer.new()
+var skip_delay_timer := Timer.new()
 
 signal dialogic_action()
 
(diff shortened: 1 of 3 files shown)


Jowan-Spooner:
I'll have to test this when I'm back at the end of next week. Sounds good though, Thanks! Also whenever possible please DO report all bugs ;).

Jowan-Spooner:
Works as expected. 

As asked on discord, I'm wondering (because I'm not testing it in a real game scenario, but maybe you did): 
- how does this feel when text speed is 0 (instant)
- and what default-delay would you suggest feels best with the given default speed.
*I don't care so much for keeping behaviour persistent in alphas (as long as previous behaviour is achieveable), but would prefer a good default.*

I would be okay with merging this and fine-tuning the behaviour/default values later.

KoB-Kirito:
Not sure if a respone is better here or on discord xD

Nothing really changes with text speed really as it only effects when you can proceed to the next event. If you have instant text, that's still the same amount of delay. In that case when you proceed to the next text instead of when you can skip the fill animation and then proceed to the next text/event. In both cases you would need some time to comprehent that you skipped and in both you don't want to skip accidently two times. If you want your player to skip the whole section/dialog you would/should solve that differently anyways imho. It just delays input, you can still use autoskip or one input to skip all without any delay.
I felt that you don't notice 0.1 at all, but the double-skip bug was gone. So I would suggest that as a default.

Jowan-Spooner:
Hah, yeah sorry to write multiple times, I just wasn't sure if you'd respond on discord soon and this place is a bit more organized (you never know when the discord discussion suddenly jumps to something new). We can finish this discussion on discord though.

PR #1641, opened 2023-07-21
Silence error on get_meta(..., null)

oddfacade:
As per the `Object` docs, providing a default value of `null` to the `get_meta` call doesn't silence the error in the event that the metadata key doesn't exist. We have a few instances of this pattern throughout the codebase, which were producing unintended error messages under certain circumstances.

This commit replaces them with explicit `has_meta` checks, where appropriate. In situations where the null value should be an error (typically because it was previously checked) I removed the default argument, to avoid confusion (the behavior is the same).

Also did some boy scout rule cleanup of trailing whitespace and refactored some confusing branch logic in `DialogicGameHandler.gd`.

### DIFF
--- a/addons/dialogic/Editor/CharacterEditor/character_editor.gd
+++ b/addons/dialogic/Editor/CharacterEditor/character_editor.gd
@@ -449,7 +449,7 @@ func update_preview() -> void:
 		var offset:Vector2 =current_portrait_data.get('offset', Vector2()) + current_resource.offset
 		
 		if current_previewed_scene != null \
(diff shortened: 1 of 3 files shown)


coppolaemilio (approved):

coppolaemilio:
@Jowan-Spooner want to take a look? looks good to me

Jowan-Spooner:
Generally looks good to me as well. I'm still traveling and access to internet is rare so I can't trst it. I'll be properly back on Friday or Saturday. But this looks like it can be merged.

coppolaemilio:
Thank you @oddfacade !

PR #1661, opened 2023-08-08
Small broken reference fix

CristianRos:
A weird bug was happening with indentation that was making the reference_manager script to not load. I also fixed a typo I found there.


Jowan-Spooner:
Thanks!

PR #1669, opened 2023-08-10
Small shortcut bug fix

CristianRos:
This small bug was preventing the timeline editor to not load properly.


Jowan-Spooner:
Thanks!

PR #1676, opened 2023-08-11
Fixed ERR_INVALID_PARAMETER in animation.gd

endaye:
Fixed an error in `animation.gd`.

## Problem

* If `animation_finished` signal is already connected to `Dialogic.Animation.animation_finished`, it will cause this error  [@GlobalScope.ERR_INVALID_PARAMETER](https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#class-globalscope-constant-err-invalid-parameter)
image

## Solution

* According to [Signal](https://docs.godotengine.org/en/stable/classes/class_signal.html#class-signal-method-connect) docs, use [is_connected](https://docs.godotengine.org/en/stable/classes/class_signal.html#class-signal-method-is-connected) first to check for existing connections.

### DIFF
--- a/addons/dialogic/Modules/DefaultStyles/Default/animations.gd
+++ b/addons/dialogic/Modules/DefaultStyles/Default/animations.gd
@@ -13,16 +13,19 @@ func _on_textbox_show():
 	%DialogicNode_DialogText.text = ""
 	get_node("../DialogTextAnimationParent").modulate = Color.TRANSPARENT
 	play("text_box_reveal")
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
Hey thank you for the PR! I didn't think this would be necessary due to the ONE_SHOT connections (so it should disconnect before it comes up the next time). 
But if you encountered the error message and this fixes it, while maintaining the same functionality, that is great. I'll test it in a moment and merge after.

Jowan-Spooner:
Looks good to me! Thanks.

endaye:
Yes. When you connect a ONE_SHOT signal twice before you emit it, it will return the error.

PR #1693, opened 2023-08-27
Fix input event handling

iscke:
this somewhat fixes a particular bug I've experienced in text handling, in which as far as i can tell:
 - the user triggers `dialogic_default_action` on the same frame they also trigger another input action (say, moving the mouse)
 - `_input` is called for the `dialog_default_action`, `dialogic_default_action` has been pressed this frame, so the coroutine to advance text (`handle_event`) starts
 - `_input` is called for the other action, `dialogic_default_action` has been pressed this frame, so the coroutine to advance text starts again
 - the coroutines resolve, text is advanced twice, and a line gets skipped

i believe this PR fixes the input situation, but the race condition still probably exists. i don't understand how godot godot coroutines work well enough to tell if it's actually an issue, though.


Jowan-Spooner:
Thanks!

PR #1747, opened 2023-09-22
Invalid operands 'Object' and 'String' in operator '!=' fix.

NocturneLight:
The line "Aracelia: foo" followed by "James: bar" would cause the error to occur and Godot to hit a breakpoint.

By converting the speaker value to a string, the error goes away and Dialogic proceeds as normal.


Jowan-Spooner:
This is not the correct solution functionality wise. I've encountered this too and would fix this in an upcoming PR. 
The line should be
```
     if speaker and speaker.resource_path != dialogic.current_state_info['character']:
```
You can change this or you can wait for my PR if you want

coppolaemilio:
Fixed by https://github.com/coppolaemilio/dialogic/commit/c3418b9ce20d34fff0a664b75cf65971a4d9b582 as Jowan suggested.

Issue #1733, opened 2023-09-16
Autoadvance when voiceline finishes


Jowan-Spooner:
In 1.x we had [nw=v], but there isn't yet an equivalent in 2.x.

PR #1755, opened 2023-09-23
Extend Auto-Advance

CakeVR:
# Problem
The current Auto-Advance delay does not respect:
- Text's character count
- Text's word count
- Typewriter time taken
- Voice lines still playing

The developer cannot expose a UI setting the Auto-Advance speed to scale with their reading proficiency.

# Solution
On the user-facing end, this Pull Request adds a new UI to the Text Settings, offering to select the specific scaling values for aforementioned ignored text properties.

Additionally, adds support to ignore specific characters, to prevent punctuation and whitespaces to be exempt from Per Character calculation.
The String input gets mapped into a Dictionary with `null` values, a Hash Set in disguise.

The user-facing settings will be stored in the state info dictionary `autoadvance`, alongside the already added keys `temp_enabled` and `temp_wait_time`. 

Voice lines playing longer than the calculated Auto-Advance delay will override the waiting time, to allow them playing until the very end.

The typewriter will now save their time taken to type out the entire text into a state info. This allows the Auto-Advance to respect the base and speed modifiers that may have taken place and changed the read time experience.

The player can change the delay modifier via `Dialogic.Settings.autoadvance_delay_modifier`.

# Problems
The Auto-Advance strips BBCodes via RegEx, however, it does not differentiate between `[text]` and official/custom BBCodes, for instance `[b]`.
Therefore, if the player wants to use these in their text, the Auto-Advance delay calculation will ignore the square bracketed text.

# Related Issues
This Pull Requests closes:
- https://github.com/coppolaemilio/dialogic/issues/1733


Jowan-Spooner:
Note: The final PR works a bit differently from what is described in the initial comment. That is no problem as hopefully a good wiki page will be created soon.

Issue #1749, opened 2023-09-22
When providing a folder path to Typing Sounds, Dialogic doesn't utilize every sound in that folder


DancingEngie:
## The problem

**Describe the bug**
Learning Dialogic (and Godot), I provided my character with typing sounds from dialogic/Example Assets/sound-effects. However, when playing my game, I noticed the engine only utilized 2 typing SFX (the folder has 5).

**To Reproduce**
1. Create a new character and give it some dialogue
2. Create a new typing sound for that character and provide it with valid SFX folder path. The SFX folder should have more than 2 files
3. Run the timeline

**Expected behavior**
Dialogic should run through a folder and utilize every typing sound provided to it.

**Screenshots**
The sample typing SFX v
![image](https://github.com/coppolaemilio/dialogic/assets/9748479/f56b92eb-5956-4e9a-811a-8a4ebc19cbef)

Console logs of the current AudioStream sound v
![image](https://github.com/coppolaemilio/dialogic/assets/9748479/4aeee9c9-1522-4b95-9a2b-01d01008c595)

...as placed in node_type_sound.gd:
```
#choose the random sound

audio_player.stream = current_overwrite_data.get('sounds', sounds)[RNG.randi_range(0, sounds.size() - 1)]

print(audio_player.stream.resource_path.get_file())
```

**System (please complete the following information):**
 - OS: Windows 10
 - Godot Version: 4.1.1
 - Dialogic Version: 2.0-Alpha-9

## Solutions

**Workaround**

It would appear the code doesn't properly update the "sounds" array, but does update "current_overwrite_data". Forcing the AudioPlayer to use that overwrite data solves this issue:

```
#choose the random sound
var sounds_array = current_overwrite_data.get('sounds', sounds)
audio_player.stream = sounds_array[RNG.randi_range(0, sounds_array.size() - 1)]
```

**Possible Fix**

Properly updating the "sounds" array.

Jowan-Spooner:
@DancingEngie Thank you so much for the fix. I've included it in my latest PR #1763 so it will be in the next release!

Issue #1374, opened 2022-12-21
Allow to enter color codes and names for default background scene 


Jowan-Spooner:

Jowan-Spooner:
I've made this possible in #1763

PR #1763, opened 2023-09-29
Multiple small features & fixes

### LINKED ISSUE #593: Supporting line breaks within choice boxes
For the majority of my game, I have rather long pieces of text in my choice boxes, and find myself in dire need of line breaks.

Supporting line breaks within the choice boxes :)

Using a custom button scene instead? But I don't think that that would be worth the hassle just for a simple line break, and the choice boxes otherwise fit my needs perfectly, so it would be super helpful if it could just be implemented directly into the choice boxes.

Jowan-Spooner:
- color names and codes on default background (#1374)
- fix autopauses and remove hide_textbox_when_empty setting
- support modifiers (br, random selection) in choice texts - fixes #593 
- fix for #1749


Jowan-Spooner:
> When I added an Auto-Pause, Dialogic will report:
> Invalid get index 'text_speed' (on base: 'Node (subsystem_settings.gd)').

@CakeVR I pushed a fix for this. When we merge the autoadvance PR we can choose your fix or this one, idc.

PR #1766, opened 2023-10-01
Fixes conflicts and crashes

eddieataberk:


addons/dialogic/Other/DialogicUtil.gd
```diff
@@ -14,12 +14,11 @@ static func get_editor_scale() -> float:
 	return get_dialogic_plugin().get_editor_interface().get_editor_scale()
 
 
-## Although this does in fact always return a EditorPlugin node,
-##  that class is apparently not present in export and referencing it here creates a crash.
-static func get_dialogic_plugin() -> Node:
-	var tree: SceneTree = Engine.get_main_loop()
-	if tree.get_root().get_child(0).has_node('DialogicPlugin'):
-		return tree.get_root().get_child(0).get_node('DialogicPlugin')
+static func get_dialogic_plugin() -> EditorPlugin:
```
Jowan-Spooner, line 17: We cannot return EditorPlugin, instead we should return Node as explained by the comment above.

addons/dialogic/Other/DialogicUtil.gd
```diff
@@ -14,12 +14,11 @@ static func get_editor_scale() -> float:
...
+	var children = Engine.get_main_loop().get_root().get_children()
```
Jowan-Spooner, line 18: We should use static typing wherever possible although creating the variable here seems unnecessary anyways.

addons/dialogic/Other/DialogicUtil.gd
```diff
@@ -14,12 +14,11 @@ static func get_editor_scale() -> float:
...
-## Although this does in fact always return a EditorPlugin node,
-##  that class is apparently not present in export and referencing it here creates a crash.
-static func get_dialogic_plugin() -> Node:
-	var tree: SceneTree = Engine.get_main_loop()
-	if tree.get_root().get_child(0).has_node('DialogicPlugin'):
-		return tree.get_root().get_child(0).get_node('DialogicPlugin')
+static func get_dialogic_plugin() -> EditorPlugin:
+	var children = Engine.get_main_loop().get_root().get_children()	
+	for child in children:
+		if child.get_class() == "EditorNode":
```
Jowan-Spooner, line 20: Not sure why you are checking for class type instead of the name? Why would the node be named differently?

Jowan-Spooner (changes requested):

addons/dialogic/Other/DialogicUtil.gd
Jowan-Spooner: Alright I understand now, I'm okay with this!

Jowan-Spooner (approved):

Issue #90, opened 2021-03-19
Resources unavailable in HTML game


arnaudvergnet:
* **Dialogic version**: rc_1.0
* **OS**: Ubuntu 20.04

## Summary

Resources such as images are not loaded Web game.

## The problem

**Steps to reproduce**

* Create a new game with this plugin
* Create a new character with a portrait
* Export the game as explained in the [readme](https://github.com/coppolaemilio/dialogic#-why-are-the-dialogs-are-not-working-when-exporting-my-project)
* Start the game on web, you will not see your character portait

**What is the current behavior?**

When exporting a game, characters portraits are not loaded. Did not try scenes but it might have the same problem.

**What is the expected correct behavior?**

Portraits should show like the rest of the game.

## Solutions

**Possible fixes**

According to the [documentation](https://docs.godotengine.org/en/stable/classes/class_file.html#class-file-method-file-exists), `file_exists` method should not be used to check for resources as their names are changed inside the packed resources. Instead, [`ResourceLoader.exists`](https://docs.godotengine.org/en/stable/classes/class_resourceloader.html#class-resourceloader-method-exists) should be used.

So to fix this issue, it should be enough to replace `file_exists` when checking for resources availability by `ResourceLoader.exists`.

coppolaemilio:
Are you going to work on this one or should I?

arnaudvergnet:
I can do it if it saves you some time

PR #1773, opened 2023-10-03
Change file exists method for CharacterResource

lamnguyenkhoa:
If using FileAccess.file_exists(), I can't load character portrait in HTML game. This is the same issue in #90 and fixed in PR #91. It seem like the fix not carried over to Dialogic 2.0.

### DIFF
--- a/addons/dialogic/Resources/CharacterResourceLoader.gd
+++ b/addons/dialogic/Resources/CharacterResourceLoader.gd
@@ -27,7 +27,7 @@ func _handles_type(typename: StringName) -> bool:
 # parse the file and return a resource
 func _load(path: String, original_path: String, use_sub_threads: bool, cache_mode: int):
 #	print('[Dialogic] Reimporting character "' , path, '"')
-	if FileAccess.file_exists(path):
+	if ResourceLoader.exists(path):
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
Looks good and logical, but have you checked for other wrong uses of file_exists?
In the editor file_exists might be correct but every resource access in-game should use ResourceLoader I guess?

lamnguyenkhoa:
No, havent look in other Resource file yet. I can do that if you want

Jowan-Spooner:
If you would, that would be great, otherwise I can also do that.

Jowan-Spooner:
@lamnguyenkhoa do you think you have time to do this? otherwise I'll just merge this and do the others myself.

lamnguyenkhoa:
Sorry for the past day. Yeah, but it less about time and more about Im unable to test it. There is another `FileAccess.file_exists` in TimelineResourceLoader but Im unable to test it so im not sure. You can go ahead and do them. Thanks for following up.

Jowan-Spooner:
Thanks again for bringing this up. No worries, testing hypotheticals that one didn't encounter themselves is always hard.

PR #1777, opened 2023-10-05
fixes the issue choices not appearing on text bubble

eddieataberk:


Jowan-Spooner:
Thanks for your effort. 

I see that on discord you wrote:
> there is an issue with text bubble layout. the choice system works by get_nodes_in_group('dialogic_choice_button') but in textbubble all bubles have choice boxes that are in this group so only the first registered bubble ie the fallback bubble works

*Please include a reasoning or link to an issue in your next PR, as I don't want to search whether any explanation has been posted elswhere. Thanks ;)*

I cannot reproduce the problem you are describing. Multiple text bubbles with choices work fine for me. Textbubbles are never visible at the same time anyways. So they shouldn't interfere with each other. (If they were though, this might be a good solution, although an "enabled" flag might be even better). 

Can you give some details/examples where this happens? Chance is I'm just to dumb to see this, but I need to ask.

Jowan-Spooner:
@eddieataberk I will close this unless you can give some further detail on how to reproduce the problem. Thanks

PR #1778, opened 2023-10-05
Added method to allow for using regex in expressions

opaintner:
I added a method to the "Core Methods" section of the expression subsystem in order to implement regex pattern matching for use in things like conditions, etc.

Not super sure if I did this correctly. I wasn't able to push a new branch, so I forked and then am wanting to merge my fork's branch into dialogic? It's all a bit confusing!


Jowan-Spooner:
Thanks!

PR #1780, opened 2023-10-05
Update TextBubble Layout to fix choice size placement

Jowan-Spooner:
Supposed to be a fix for coppolaemilio#1721, also kindof adresses coppolaemilio#1734.
@Kookiness Maybe you could try out this branch?

Choice size is now correctly taken into account and the text-bubble will get as big as needed. When choices get to long, they will get pushed into the next line.
This also means, we can't say for sure how big the choices will be which results in a size-increase after text finished (when choices become visible).

I also decided to rewrite the bubble-shader to more clearly expose the settings that are of interest for us. This results in a bit of a different look, especially with the same settings.

Also fixes
- text bubble background ignored modulate alpha

### DIFF
--- a/addons/dialogic/Modules/DefaultLayouts/TextBubble/DialogicTextBubbleLayout.gd
+++ b/addons/dialogic/Modules/DefaultLayouts/TextBubble/DialogicTextBubbleLayout.gd
@@ -40,10 +40,10 @@ extends CanvasLayer
 @export_subgroup('Choices Text')
 @export var choices_text_size := 15
 @export_file('*.ttf') var choices_text_font := ""
-@export var choices_text_color := Color.LIGHT_SLATE_GRAY
-@export var choices_text_color_hover := Color.DARK_GRAY
-@export var choices_text_color_focus := Color.BLACK
-@export var choices_text_color_disabled := Color.LIGHT_GRAY
+@export var choices_text_color := Color.DARK_SLATE_GRAY
+@export var choices_text_color_hover := Color.DARK_MAGENTA
+@export var choices_text_color_focus := Color.DARK_MAGENTA
+@export var choices_text_color_disabled := Color.DARK_GRAY
 
 @export_subgroup('Choices Layout')
 @export var choices_layout_alignment := FlowContainer.ALIGNMENT_END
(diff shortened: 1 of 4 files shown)


Kookiness:
Certainly! I'll give this branch some testing. Thank you for working on this!

Kookiness:
Hm, I'm getting an error immediately after confirming dialogue choice selections:

image

Jowan-Spooner:
@Kookiness You probably had the name label disabled. I've pushed a fix for that case. Could you try it again?

Kookiness:
@Jowan-Spooner

I like the changes so far! Very nice.

I noticed that the size of the bubble around its text isn't always consistent.

Here are some screenshots (two separate ones since the bubble didn't all fit onscreen at once) using default font, font size, padding, but with increased max bubble width:

image

image

Here's a screenshot using a different font and reduced max bubble width:

image

Also, would there be the possibility of having an option to left-align the choices?

And maybe the possibility to force each choice onto a separate line, regardless of the choice's length?

I ask because those seem like the ways I would most intuitively expect choices to be displayed, so the player can just scan down the list quickly and not risk the mistake of not realising that one line has two separate choices displayed on it.

Happy to do further testing!

Jowan-Spooner:
It was having trouble finding the problem here, but turns out (as shader parameters are shared across instances) any hidden textbubble would also set the box_size parameter of the shader. However while being hidden they would often be of a different aspect ratio and considerably smaller and thus the box background would be stretched a lot (resulting in different padding and bad border aliasing). 
I will try to also adress the other requests unless it turns out to be a lot of work.

Kookiness:
Great stuff, glad you figured it out. Yeah, I noticed it looks like the speech bubbles were being stretched, rather than using some kind of 9-slice/9-patch type solution. I assumed that was an intentional design decision to be to make the roundness of the corners reflect the dimensions of the bubble, but I'm not certain.

Regarding my suggestion about left-alignment - I just want to mention that I later made a small edit to my comment, to say that having the choices centred might actually be what I would expect - but I think this only makes the most sense if the choices are listed as one per line.

Anyway, no pressure from me - I appreciate you making use of my feedback!

Kookiness:
@Jowan-Spooner Not sure if you want more testing yet, but when I try the latest version of this branch, this is what I get with bubbles that have choices, using default size values:

image

Jowan-Spooner:
@Kookiness Strange, it works fine for me, and I think this has all settings at default size too.
grafik

Kookiness:
@Jowan-Spooner Oh, strange. I tried making sure every single setting was reset, and enabled displaying of the character name too, just to make sure we had everything the same.

If it helps, I notice that when choices are displayed, I also get the character name offset differently than yours, so it sits further away from the bubble, making it seem like the bubble needs to be bigger. The bubble also looks more rounded when there are choices.

Here are screenshots comparing bubbles with choices vs. no choices:

image

image

Kookiness:
Huh, the issue doesn't happen when I just click the "Test Style" or "Play Timeline" buttons; it only happens when I run the game.

image

PR #1789, opened 2023-10-07
Add sound effect settings to default layout

Jowan-Spooner:
Also
- moves choices in front of textbox
- adds a Directory mode for @export_dir settings
- makes file picker in directory mode respect max_lenght too


Jowan-Spooner:
grafik

Issue #1809, opened 2023-10-15
Progress Indicator keeps lowering


CakeVR:
## The problem

**Describe the bug**
Every new Timeline Text Event, the Indicator keeps getting lower.

**Expected behavior**
The Indicator should remain at the same location.

**Screenshots**

**System:**
 - OS: Windows
  - Godot Version: 4.2.beta.1
 - Dialogic Version: 2 Alpha 10
![image](https://github.com/coppolaemilio/dialogic/assets/65981767/162647e7-d7f6-471b-a871-8ca87c49367e)

PR #1827, opened 2023-10-26
Fix next indicator slipping off screen

Pheubel:
Resolves https://github.com/coppolaemilio/dialogic/issues/1809

When a new animation started, it would take the values left behind by the previous tween. By keeping a reference to the current animation tween we can ask it to clean up first before having a new tween take over.

### DIFF
--- a/addons/dialogic/Modules/Text/node_next_indicator.gd
+++ b/addons/dialogic/Modules/Text/node_next_indicator.gd
@@ -19,6 +19,7 @@ extends Control
 		if has_node('Texture'):
 			get_node('Texture').texture = texture
 
+var tween: Tween
 
 func _ready():
 	add_to_group('dialogic_next_indicator')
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
Not entirely sure I understand why this fixes it, but it seems to work pretty well. So 💯 . Thanks for the work!

Pheubel:
> Not entirely sure I understand why this fixes it, but it seems to work pretty well. So 💯 . Thanks for the work!

I think the best way to explain it is to first explain why it broke in the first place. Before it would tween using the position of the indicator, it would change the position up and down. every time the text event got called it would look at it's current position and use that as it's origin for the tween, but because the position was on a different position due to the previous tween it starts slipping off.

So the root of the problem is that the start position of each tween is not equal. What needed to happen was that it got reset before the next tween would begin. If you look at the stop function on a tween object it says:

> Stops the tweening and resets the Tween to its initial state.

This is exactly what needed to be done. All that there was left to do was keep a reference to the current tween so that we can call stop on it and reset it's position before starting the next one.

I hope this clears things up, @Jowan-Spooner

Issue #1828, opened 2023-10-26
[Suggestion] Allow signal event pass a dictionary rather than just a single string


Pheubel:
**Is your feature request related to a problem? Please describe.**
Currently you can only pass one argument in a signal event. This makes it harder to filter the signal and read values from it at the same time.

**Describe the solution you'd like**
By sending a dictionary rather than a string, you can define multiple arguments to be passed into the signal. Inside of the handler you can then filter for specific properties to read out. 

As an example, in one of my games I use a signal to pause the timeline at the start and the end in order to do custom transitions over the background i show in my timeline without having to use dummies in the scene as a fall back. I also use signals in order to unlock CG's in the gallery.

With this proposal, the argument of the signal event would look closer to the call node event, where you are able to add more argument, but instead they would be key-value pairs.

~~There could be the concern of backwards compatibility. There are two ways to go about it:
1. Introduce this as a breaking change, right now dialogic 2 has not been fully released yet, so a stable API is not guaranteed. It can be a bit annoying to people, but this seems to be the easiest to maintain.
2. Add the option to switch between single string input and dictionary input on the signal event. As the signal is already type-less today, it is up to the developer to assume what kind of data will be passed to it. It would require a bit more thought in the text editor, maybe adding a parameter dictating if it is string input or dictionary input mode, with the lack of the parameter defaulting it to string input mode. That way existing signal events should keep working. As for how to handle key value pairs in the text editor, this could be done in multiple ways, one way to do it is to list them like  `key1:value1;key2:value2`. EDIT: since it's already handles dictionaries somewhat, it seems that a syntax won't be that big of an issue. what is a bit of an issue is that it prints dictionaries with pretty print, which is not the greatest~~
EDIT2: after tinkering and seeing what is possible, there isn't a reason to be concerned in regards to backwards compatibility. 

If this suggestion would get accepted, it would be a nice to have feature where if you used a specific key before in a different event, it could suggest it to you when you try to add a key to a different signal event. This would prevent mistakes where you accidentally misspelled a key and would fail the look up for a specific signal event. I don't think that the lack of this feature should block this proposal, as I think the main suggestion could be useful enough on it's own.

**Describe alternatives you've considered**
The alternative would be to keep using the current system, but rely on the developer to come up with ways to handle multiple data passed in a signal.

**Additional context**
N/A

PR #1829, opened 2023-10-27
Allow for signal events to optionally pass dictionaries instead of strings

Pheubel:
This would resolve https://github.com/coppolaemilio/dialogic/issues/1828

I made this mostly to demonstrate how i envision it to work and think it works. I tried playing around with it and am fairly satisfied.

What I added/changed:
* An input scene that acts similar to an array, but instead there are two fields per entry, a key and a value. To keep it simple enough, values are exclusively strings, rather than allowing any data type.
* Added resource type `KEY_VALUE_PAIRS`, i put it under the `OTHER` category in the enum. This type is meant for dictionaries, which are key value pairs. in `to_text()` they are handled under their own clause, as putting them in `var_to_str()` would cause newlines to be added and breaks the syntax highlighting for the text editor. Instead they get serialized densely; example: `arg="{"a":"1","b":"2"}"`
* Added a toggle to switch between the signal using a string as an argument and a dictionary as an argument. It defaults to false and should be compatible with existing signals due to how the syntax works. Since you need to opt-into dictionaries as argument rather than a string, existing code does not break. But in order to take advantage of the feature, they do need to change their code to handle a dictionary argument. With a simple type filter it should work like a charm.
```gdscript
# a simple example of how a type filer could look
func on_timeline_signal(arg: Variant) -> void:
	match typeof(arg):
		TYPE_STRING:
			handle_string_argument(arg as String)
		TYPE_DICTIONARY:
			handle_dictionary_argument(arg as Dictionary)
```

What can still be added:
* For the key value pair inputs, a user could define the type per value, rather than locking them to just strings. However for this to be reliable, it would need a custom serialization step, rather than using Godot's built in JSON stringifier. As it follows the JSON specs, it treats all read values as floats since JSON doesn't distinguish between the two, it only has a number type for numbers. This increases the complexity of implementing this feature and could possibly be added in the future.
* Better wording for the visual editor of the signal node. I found it challenging to come up with a good short description that doesn't confuse people. I'm more a technical minded guy and am generally not good with describing things like this after a code jam.

I hope you give this feature a try and see if this is a good fit


Jowan-Spooner:
I like the idea and the implementation isn't bad either, but UI/UX need improvements. I will try to improve it and then merge it.

Jowan-Spooner:
Okay I made a pass on UI/UX/naming:
## Default look:
grafik
## Default expanded:
grafik
## After switching to Dictionary:
grafik

## With some values:
grafik

## In text:
grafik
grafik

Jowan-Spooner:
The dictionary UI looks really rought, but it's based of the array UI which looks similarly bad, so I'll leave that for a future PR.
I'd also consider both in array and dictionary uses to do a parse_variables pass to allow easy passing of dialogic variables in those.

Jowan-Spooner:
Thanks for this feature!

PR #1836, opened 2023-10-28
Fixed typo on home page

Proggle:
The home page uses "let's" instead of "lets" and it's bugging me.


Jowan-Spooner:
Haha fair enough. Thanks for the PR!

Issue #1831, opened 2023-10-27
Event Background: Transitions with fade time cause scene background to bleed in


Pheubel:
## The problem

**Describe the bug**
When you have a background A and switch to background B with a fade time larger than 0, you can see the scene colors, what is behind the backgrounds, blend in. This makes it hard to make transitions look right.

**To Reproduce**
1. In `Project` > `Project Settings` > `General` > `Rendering` > `Environment` set the default color to something noticeable, like bright red.
2. Create a timeline
3. Add a background event
4. Create another background event
5. Set the fade time of the second background event to 1 second
6. Play the timeline

**Expected behavior**
The two background should blend into each other without the scene bleeding into the view. 

**Screenshots**
![problem](https://github.com/coppolaemilio/dialogic/assets/22686676/4733807b-8e76-4ed6-811b-fc6e973730f6)
When switching backgrounds you can clearly see the red color from the default scene environment bleed into the view.

**System (please complete the following information):**
 - OS: Windows
 - Godot Version: 4.1.1
 - Dialogic Version: 2.0 - Alpha 10

## Solutions

**Workaround**

Not really dialogic, but what i did in my own project i had a need for a similar transition system, for that I use of the image scene A produced and scene B produced and inserted it into a shader and it then lerps the colors of both scene images. It then put's the result on top of both scenes as a sort of overlay in a color rect that hides the scenes behind it. When the transition is over, the overlay is hidden again.

**Possible fixes**

I think that the issue stems from the fact that the backgrounds try to become transparent, rather than each other. With how transparency works this can cause some bleeding through of things behind the images to pop up

Since you have two backgrounds that go between white (indicates that the background is fully visible) and transparent (background not visible).

One way to go about it is to change how backgrounds get transitioned into each other would be change the system with something similar described in https://github.com/coppolaemilio/dialogic/issues/1826#issuecomment-1779667691. In order to get a simple fade transition you could use a full mid gray texture and set the feather to 1, this will fade the entire image at once close to how it's currently done out of the box (minus the bleeding).

A concern with it however is the scale of the backwards compatibility break. I haven't done much research yet, but some points that should be looked at are:
* change in scene structure - it could be that in order to get it to work nicely a change in the scene's structure has to be made, making user made scene possibly incompatible.
* the role of `fade_in()` and `fade_out()` - if the transitions are handled differently for the custom scene using these functions they most likely aren't going to work so well. Their use in the current eco system might need to be investigated.

Issue #1826, opened 2023-10-24
[Suggestion] Allow for more background transitions out of the box


Pheubel:
**Is your feature request related to a problem? Please describe.**
During a visual novel it is common to change the background to go to a different "scene", to make those background changes feel nicer transitions are used. However in dialogic it seems that the only options are either a hard cut to the next background or a fade. It would be nice if it was possible to create your own transitions in order to create a more diverse experience while switching backgrounds. examples include:
* non linear fades (like a sudden flash to white to show a heart racing event)
* pattern based transitions to fit the theme of a game (for example a game with cats could have paw prints covering the original BG)
* a melt fade to creep out the player
* etc.

Additionally the current background fade seems to have a small issue where the background can become noticably darker during a quick fade between two almost identical BG's. 

**Describe the solution you'd like**
I think a nice way to give power into the developer's hand would be to allow shader materials to be passed to the background event. These shader materials would have 3 parameters in common:
* progress (this would be a float between 0 and 1)
* previousBackground (this would be a background set by a previous background event. not always assigned, for example at the start or after a clear with backgrounds turned on)
* nextBackground (this would be the background that is being assigned in the current background event)

When a background event happens, it would make use of the fade time to get the current progress. It sets the previous and next background parameters and from there the mechanism that currently handles fades between background can most likely be used to update the progress.

If a developer wants to create their own transition they can create a shader that has those parameters and even other parameters. Inside of the fragment function they can then decide how the transition should behave and they can tweak settings inside a shader material that they then pass to the background event. if no shader material is passed, it could fall back on a default shader material that does a fade like the current fade between background events.

This way the transitions become fairly plug & play and can easily be shared.

an example of how one of these transition shaders could look would be like so:
```c++
shader_type canvas_item;

uniform float progress : hint_range(0.0, 1.0);
uniform sampler2D previousBackground : source_color, hint_default_transparent;
uniform sampler2D nextBackground : source_color, hint_default_transparent;

uniform float diamondPixelSize = 10.;

void fragment() {
	float xFraction = fract(FRAGCOORD.x / diamondPixelSize);
	float yFraction = fract(FRAGCOORD.y / diamondPixelSize);
	
	float xDistance = abs(xFraction - 0.5);
	float yDistance = abs(yFraction - 0.5);
	
	bool check = xDistance + yDistance < progress;
	
	vec4 color = texture(screenTexture, UV) * float(!check);
	color += texture(transitionTexture, UV) * float(check);
	
[...]

PR #1839, opened 2023-10-28
Fix background bleeding through 

Pheubel:
This resolves https://github.com/coppolaemilio/dialogic/issues/1831

In order to fix this issue i implemented it as described in https://github.com/coppolaemilio/dialogic/issues/1826#issuecomment-1779667691

I had to change a few things:
* The structure of the default default dialog node scene has changed a little bit, I added two nodes in the background holder, a node that contains all the backgrounds that get swapped in. And a color rect called `Overlay` that uses the backgrounds to draw the actual background that needs to be displayed.
* The default background scene has changed a little bit, instead of a color rect with a texture rect child, i have a parent SubViewportContainer with a viewport as child. The overlay relies on the viewport to get a texture of what is happening in the scene to the overlay. The texture provided by the viewport is actually dynamic, so any changes that happen inside the viewport do get drawn, so animated scenes as backgrounds should work just fine.
* The code for the background submodule has been changed a bit, the role of `_fade_in(float)` and `_fade_out(float)` have been reduced to notifications that the background scene is fading in/out, i think this is the best fit for them, as the transition.
* I added a whipe transition shader and material, this is what powers the overlay, it transitions between the previous and next background. For now, the feather and whipe texture have been set in the material, to `1` and a midtone gray respectively. This paves the road for possible custom transitions using whipe textures.

Breaking changes:
* existing custom background scenes will most certainly break, as they do not return a ViewportTexture. In order to fix them, their scene structure has to be changed to have a SubViewportContainer as root with a SubViewPort as child, they can then put everything they had as child of the SubViewPort. Afterwards they need to make sure that the node references are all wired up correctly. 
* Custom transitions in `_fade_in(float)` and `_fade_out(float)` will likely behave in an unexpected way, i didn't have a use case to test how they would behave.

What this allows for in the future:
* Custom transitions using self made textures (easy for developers) - The mechansim implemented for the transitions here makes use of a whipe texture with a feather already, the code needs to be altered in such a way that it keeps the current settings as a default if no whipe texture is provided. There might also need to be an option to indicate how the whipe texture should be applied (toggle between stretch to rect VS scaled to keep the aspect ratio) I don't think it would be a hard task, but requires a bit of UI design.

* Custom transitions using custom shaders (harder for developers) - The shader would have to have a material replace the material in the overlay material, this would require a bit more thought, it will essentially be like the texture transitions, but will require some more code to make sure that the properties are set right. It will require a bit more UI work in order to get this to work well. I would imagine that if this were to be added, it will need to make use of the Key Value pair element that i made for https://github.com/coppolaemilio/dialogic/pull/1829, but more advanced to allow for data types outside strings. I think this feature is possible, but definitely an advanced feature.


Jowan-Spooner:
Note to future visitors: The implementation of this has changed quite a bit.

Jowan-Spooner:
@Pheubel Thanks for all the work and patience!

PR #1843, opened 2023-10-29
Add a separate install step after adding all subsystems

### LINKED ISSUE #1819: Crashes new project on start with "Invalid get index 'Settings'..."
## The problem
Trying [Dialogic 2.0 - alpha 10](https://github.com/coppolaemilio/dialogic/releases/tag/2.0-alpha-10) on a brand new project and it crashes as soon as I run any scene.

**Describe the bug**
Crash on run with error message:
`Invalid get index 'Settings' (on base: 'Node (DialogicGameHandler.gd)').`
at:
`addons/dialogic/Modules/Text/subsystem_text.gd:464`

**To Reproduce**
1. I created a new project and closed it.
2. I installed the dialogic addon from the zip of the alpha release.
3. I enabled the project > project settings > plugins
4. Reloaded the project
5. Check that it is loading in project > project settings > autoload as well.
6. Run the scene
7. See error

**Screenshots**
Screenshot from 2023-10-20 11-16-30

 - OS: Linux
 - Godot Version: 4.1.2
 - Dialogic Version: 2 alpha-10

## Solutions

**Workaround**

Using different dialogue system.

imberny:
* this "install" method can be safely used by subsystems which rely on other subsystems
* fixes #1819 which was caused by a subsystem fetching another subsytem not yet in the tree (introduced by #1718)

Since the subsystems are loaded by parsing the directory, I assume the order might vary by OS (I'm on Linux and could reproduce #1819).

### DIFF
--- a/addons/dialogic/Other/DialogicGameHandler.gd
+++ b/addons/dialogic/Other/DialogicGameHandler.gd
@@ -206,7 +206,7 @@ func collect_subsystems() -> void:
 				subsystem_nodes.push_back(node)
 
 	for subsystem in subsystem_nodes:
-		subsystem.install()
+		subsystem.post_install()
 	
 	# Events are checked in order while testing them. EndBranch needs to be first, Text needs to be last
 	_event_script_cache.push_front(DialogicEndBranchEvent.new())
--- a/addons/dialogic/Other/DialogicGameHandler.gd
+++ b/addons/dialogic/Other/DialogicGameHandler.gd
@@ -202,7 +202,7 @@ func collect_subsystems() -> void:
 		if !Engine.is_editor_hint():
 			
 			for subsystem in indexer._get_subsystems():
-				var subsystem_node := add_subsytsem(subsystem.name, subsystem.script) as DialogicSubsystem
+				var subsystem_node := add_subsystem(subsystem.name, subsystem.script) as DialogicSubsystem
 				if subsystem_node:
 					subsystem_nodes.push_back(subsystem_node)
 
@@ -223,7 +223,7 @@ func get_subsystem(_name:String) -> Variant:
 	return get_node(_name)
 
 
-func add_subsytsem(_name:String, _script_path:String) -> Node:
+func add_subsystem(_name:String, _script_path:String) -> Node:
 	var node:Node = Node.new()
 	node.name = _name
 	node.set_script(load(_script_path))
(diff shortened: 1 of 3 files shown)


imberny:
Changed the method's name to `post_install` to better reflect its usage.

addons/dialogic/Other/DialogicGameHandler.gd
```diff
@@ -199,8 +200,13 @@ func collect_subsystems() -> void:
 		
 		# build the subsystems (only at runtime)
 		if !Engine.is_editor_hint():
+			
 			for subsystem in indexer._get_subsystems():
-				add_subsytsem(subsystem.name, subsystem.script)
+				var node = add_subsytsem(subsystem.name, subsystem.script)
```
coppolaemilio, line 205: type should be specified here, even if it is `Variant`

imberny: Done. Do you think the DialogicSubsystem type should be guaranteed by the `addsubsytsem` (sic) method?

coppolaemilio: Not sure actually. What do you think @Jowan-Spooner ? Probably yes, but I don't know the details on it.

Jowan-Spooner:
Interesting. The implementation looks pretty clean. I think the original issue could've been solved by just adding 
`await get_parent().ready`
before the Settings.connect statements in the Text subsystem.

So now I'm torn between the simple and the proper solution...

coppolaemilio:
@Jowan-Spooner if you think there's a proper solution to this it should be that. The simple solution now might be a headache in the future

imberny:
I guess it depends how often you expect subsystems to rely on eachother. The original issue (I suspect) is due to different OSes loading the subsystems in different orders, meaning it could work for the developer working on a subsystem, but break for others. The benefit of this `post_install` step is to make it explicit that *this* is where you put inter-subsystem code.

But as I am very new to this project it could be that such interdependences ought to be rare and that a simple await would do the trick, as long as people remember to include it.

Jowan-Spooner:
No, I agree with Emi that a proper solution (such as this PR suggests) is likely the better way. Also to answer the question on the review comment, yes all subsystems can be expected to be of tpye DialogicSubsystem so a static declaration would be good. 

Thanks for this work!

imberny:
Ok @Jowan-Spooner, I've added an assert as well as a DialogicSubsystem return type. I think the assert is good here, since we already expect that the script would declare a `dialogic` member variable and thus shouldn't just be *any* script.

Jowan-Spooner:
@imberny Thank you!

PR #1851, opened 2023-10-30
Support saving with encryption enabled

Pheubel:
This allows for developers to opt-into saving their data with a layer of encryption. In order to opt-in, they have to go to the settings, navigate to the `Saving` section and add a password to the `Encryption Password` field.

When setting a password, it is highly likely that existing saves made for testing become invalid and cause errors to pop up. To get rid of the errors, they can remove the files and let it set up new ones. But i would recommend to ignore setting an encryption password until exporting the game in order to make debugging easier.

Along the way i did come across some security concern when it comes to arbitrary code execution when loading variables from a file. It is possible to inject code into a save file and have it run on the end user's computer.https://github.com/godotengine/godot/issues/80562

It should be noted that by making use of encryption, it becomes a fair bit harder to create a code injection, as long as the password does not get known to the attacker.

So for now _I would personally_ recommend developers making use of an encryption password when releasing their game.

### DIFF
--- a/addons/dialogic/Modules/Save/settings_save.gd
+++ b/addons/dialogic/Modules/Save/settings_save.gd
@@ -21,6 +21,8 @@ func _refresh():
 	%AutosaveDelay.visible = %AutosaveMode.selected == 1
 	
 	%DefaultSaveSlotName.text = ProjectSettings.get_setting('dialogic/save/default_slot', 'Default')
+	
+	%EncryptionPassword.text = ProjectSettings.get_setting('dialogic/save/encryption_password', "")
 
 
 func _on_autosave_toggled(button_pressed:bool) -> void:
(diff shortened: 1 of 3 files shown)


Jowan-Spooner:
Hey, thanks for the implementation. It looks pretty good to me, will try to do some testing on it. 
Do you think it should be a concern that we save the password in the project settings? I don't know how easy it is to edit those on an exported project, but I think after a certain point it probably doesn't matter. 

Dialogic is mostly used for single-player games where cheating is of much less concern. Same goes for arbitrary code execution which, while obvioulsy bad, in this case is less of an issue, because most times cheating players would only harm themselves.

Jowan-Spooner:
Okay I've tested this and it seems to do what it should. I've also added a setting "On (non-debug) exports only" which think should be on by default for easier debugging. It can be turned off.

Thanks for your work and patience.

PR #1858, opened 2023-11-01
Events: New call event

Jowan-Spooner:
This PR adds a new call event that replaces the previous Call Node event.

The new event has the following text syntax:
```
do Autoload.method(arg1, arg2, arg3)
```

The new visual event has pickers for autoload and method. The arguments can now be of complex types. There is warnings if the number of arguments is wrong or an argument type is wrong.

This new event ALWAYS awaits. 
Inline features have been dropped in favor of custom text effects.


Jowan-Spooner:
I will merge this as I'd like people to stop using Call Node asap.

PR #1873, opened 2023-11-04
Handle incorrect Portrait Scene path case

Invertex:
When the custom defined portrait path fails to load, the code still tries to interact with the null instance, since it only handled the case where the path was Empty, but not when simply incorrect/broken path.

Instead of `scene_path.is_empty()`, we could also use `FileAccess.file_exists(scene_path)`, but figured I'd avoid the double file access command. But maybe it would be better, to avoid the load exception, I'm not sure.


Jowan-Spooner:
Note to you and myself. FileAccess.file_exists can be tricky on exports. This might not be a. Issue here as these are scenes not the usual resources (which get imported). I will need to chekch that.

Jowan-Spooner:
Besides that, thank you for looking into it and making a PR!

Jowan-Spooner:
I changed it to use ResourceLoader.exists() just to be on the save side. Thanks for the PR!

PR #1888, opened 2023-11-10
Fix Choice Buttons connecting too often

CakeVR:
If a choice gets revisited, it errors because of connecting to the `BaseButton.pressed` again.


Jowan-Spooner:
Thanks!

Issue #1882, opened 2023-11-07
EventField: Allow typed dictionary entries (for signal event mainly)


Jowan-Spooner:

Pheubel:
I was wondering, since the original implementation was based on how Arrays were done, could the same be done again by looking at the face lifted [ArrayValue](https://github.com/coppolaemilio/dialogic/blob/main/addons/dialogic/Editor/Events/Fields/ArrayValue.gd)?

PR #1901, opened 2023-11-14
Add option to use custom transitions for backgrounds

Pheubel:
This PR implements: https://github.com/coppolaemilio/dialogic/issues/1826

It will allow for the background event to use custom transitions using a custom wipe texture or shader.

In order to allow overriding of shader parameters, i think that https://github.com/coppolaemilio/dialogic/issues/1882 is a blocking issue. Once that issue has been resolved, it should be feasible to use it to add override values for shaders fairly easily.


Pheubel:
closed in favor of https://github.com/coppolaemilio/dialogic/pull/1942

PR #1914, opened 2023-11-18
Fix Character Speaker changing

CakeVR:
# Problem
When the speaker changes from a character to none, neither the `speaker_changed` signal changed nor the state info updated.

# Solution
This changes the ternary to fall back to an empty string instead of `null`.

# Context
I am modifying my text whenever there is a valid speaker. I added a text event with a speaker and then a couple of text events without speakers.
However, the state info reported the same speaker for every text event.

# Appendix
I added documentation and adjusted the formatting of the function to be less tense and adhere to the Godot "newline per new section" mindset.


Jowan-Spooner:
Lgtm.

Issue #1816, opened 2023-10-18
Return concrete Type for Submodules


CakeVR:
**Is your feature request related to a problem?**
The lack of proper submodule types introduces unsafe access.

For instance, when we use `Dialogic.Text`, we won't be able to work with the actual type nor does the compiler know whether the accessed properties exist.

An example, `:=` is not accepted:
```
var info := Dialogic.Text.get_autoadvance_info()
```

If we call `Dialogic.Text.random_fantasy_method()`, we won't get any warning either.

**Describe the solution you'd like**
We may need concrete class types for the submodules and a way to access them from `Dialogic`.

Pheubel:
I am not sure how possible this would be in GDScript today, I am not too familiar with the whole code base, but if I am correct, these submodules get added dynamically, which makes it hard for static type system to figure out what they are supposed to be and if they exist at all. For the built in submodules you could get away by declaring them ahead of time, but extensions would have the short end of the stick. 

Using C# as an example, one way to go about it would be with generics.
```cs
public T GetSubSystem<T>(string name) where T : Submodule
{
    T module = GetNode(name) as T;

    if (module is null)
        throw new Exception($"subsystem \"{name}\" not found, make sure the path and type is correct");
    return module;
}
```

But GDScript seems to lack generics at this point in time, so this approach is not possible until then.

If it at some point does get introduced, it could looks something like this (warning, syntax speculation ahead)
```gd
func get_subsystem_as<T>(_name:String) -> T:
    var module := get_node(_name) as T
    assert(module != null)
    return module
```

This would not really a silver bullet for the problem, but would be a possible path a developer can walk with type safety.

CakeVR:
It's possible with internal core submodules, by adding class names to them and then providing properties or methods to access returning declared concrete types. Additionally, using the `setget` feature is a possibility.
Right now, every access on `Dialogic.SubModule` is an unsafe call, this is not really nice.

However, this won't work for custom submodules, hence the solution is limited.

As you have mentioned, GDScript has no generics.

Pheubel:
What naming convention would be good to follow for the submodule class names, something like `DialogicSubmoduleXXX` (for example `DialogicSubmoduleText`) is the first thing to come to my mind.

Jowan-Spooner:
Definitely Subsystem instead of submodule (as Subsystem is what we use for these scripts everywhere else). 

[...]

PR #1919, opened 2023-11-19
Return concrete types in Dialogic game handler

Pheubel:
this PR addresses https://github.com/coppolaemilio/dialogic/issues/1816

There were some concessions to be made, the properties don't match the current way of how the subsystems are accessed. But having a 1-to-1 translation would cause Godot to give an error as it shadows native types.

Another oddity is that Godot (4.1.3) _does_ recognize the properties on `Dialogic`, but _does not_ give proper hints to how the subsystem is structured. However it _does_ recognize when functions have the wrong amount of arguments. if i cast `Dialogic`  to `Dialogic` the hinting system seems to work as expected. So it seems to be a bug on the Godot side.

### DIFF
--- a/addons/dialogic/Other/DialogicGameHandler.gd
+++ b/addons/dialogic/Other/DialogicGameHandler.gd
@@ -80,6 +80,77 @@ var timeline_directory: Dictionary = {}
 var _event_script_cache: Array[DialogicEvent] = []
 
 
+## Built-in subsystems, subsystems from extensions can't be shown currently in GDScript.
+
+var audio: DialogicSubsystemAudio:
+	get:
+		return get_node_or_null("Audio") as DialogicSubsystemAudio
+
+var backgrounds: DialogicSubsystemBackgrounds:
+	get:
+		return get_node_or_null("Backgrounds") as DialogicSubsystemBackgrounds
+
+var portraits: DialogicSubsystemPortraits:
+	get:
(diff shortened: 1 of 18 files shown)


addons/dialogic/Other/DialogicGameHandler.gd
```diff
@@ -80,6 +80,77 @@ var timeline_directory: Dictionary = {}
 var _event_script_cache: Array[DialogicEvent] = []
 
 
+## Built-in subsystems, subsystems from extensions can't be shown currently in GDScript.
+
+var audio: DialogicSubsystemAudio:
+	get:
+		return get_node_or_null("Audio") as DialogicSubsystemAudio
+
```
nlupugla, line 88: Is it necessary to wright `as DialogicSubsystemAudio` here (and similar for all the other systems)? I think since you declared the property as being of type `DialogicSubsystemAudio`, it will automatically be cast to that upon return.

nlupugla (commented):
It seems like this PR now provides two essentially identical ways to access a subsystem. For example, `DialogicGameHandler.audio` and `DialogicGameHandler.Audio` both return the audio subsystem. This seems like it goes against the "have only one way to do things" mantra. I'm not entirely sure what the best solution is though...

Jowan-Spooner:
@Pheubel Thanks for this PR. I've made a slightly altered version of this, so this not relevant anymore but thank you very much anyways, it's been a long needed change!

Issue #1946, opened 2023-12-09
Background Event errors when trying to load a scene as a background


coppolaemilio:
## The problem

**Describe the bug**
When you set a Background event to use a scene as a background, it fails to do so.
First issue here:
![image](https://github.com/coppolaemilio/dialogic/assets/2206700/13cc1b87-7dbf-41f7-8f56-abe123279675)

Second:
![image](https://github.com/coppolaemilio/dialogic/assets/2206700/7058f222-1b4b-46a7-9b03-dc7831afd200)

If I comment out those two lines, then I get these:
![image](https://github.com/coppolaemilio/dialogic/assets/2206700/a27e4112-6c45-4934-b83b-109fa765c6fa)

and when I comment out all both
```
#new_node.get_meta('node')._update_background(argument, fade_time)
#new_node.get_meta('node')._custom_fade_in(fade_time)
```
it finally works and it runs the timeline with the custom background as expected.

**System (please complete the following information):**
 - Godot Version: 4.2.stable
 - Dialogic Version: (latest, main)

## Solutions
Commenting out these lines work, so I guess something might have changed since the last time this was implemented.

Pheubel:
I think that the reason why this problem pops up is because the scene you tried to use did not have a script extending the `DialogicBackground` class.

There is actually a check in place if the script on the root of the scene is (extending) `DialogicBackground`, but this check happens _after_ the damage has been done.

There are really 2 ways for this problem to be solved:
* put the check in the `add_background_node` function and enforce the developer to extend from `DialogicBackground`
* make the code in the subsystem check for the existence of functions instead of relying on typing to dictate functionality.

The assignment of a viewport is not strictly required inside of the background subsystem, as it gets fetched from the newly fabricated node anyways, but it could be used inside of the background scene itself for advanced use. a bit of an edge case. If going with the second way of solving this problem, one way to go about the viewport property is to check if it exists before assigning it, if we want to keep it.

PR #1952, opened 2023-12-14
Alternative Background Transition Implementation

Jowan-Spooner:
Based on @Pheubel's #1942 but quite modified.

Introduces a new "special_resources" system that allows modules to define some files they want to collec and can then easily access via the ResourceUtil. This is now used by both the portrait animations and background transitions. This should allow to limit the amount of methods needed on the indexer_class.

Also adds a bunch of background transitions all inheriting from DialogicBackgroundTransition, very similar to DialogicAnimation in how it works.


addons/dialogic/Modules/Background/subsystem_backgrounds.gd
```diff
@@ -54,51 +60,50 @@ func update_background(scene:String = '', argument:String = '', fade_time:float
...
-			new_node.get_meta('node')._custom_fade_in(fade_time)
-			material.set_shader_parameter("next_background", new_node.get_child(0).get_texture())
+	var old_viewport: SubViewportContainer = null
+	if background_holder.has_meta('current_viewport'):
+		old_viewport = background_holder.get_meta('current_viewport', null)
+
+	var new_viewport: SubViewportContainer
+	if scene.ends_with('.tscn') and ResourceLoader.exists(scene):
+		new_viewport = add_background_node(load(scene), background_holder)
+		if not new_viewport.get_meta('node') is DialogicBackground:
```
Pheubel, line 77: The check to see if the viewport's node is an actual `DialogicBackground` comes a bit too late. This is related to https://github.com/coppolaemilio/dialogic/issues/1946. the check should instead be done inside the `add_background_node` function after the scene has been instantiated on line [113](https://github.com/coppolaemilio/dialogic/blob/background-transitions/addons/dialogic/Modules/Background/subsystem_backgrounds.gd#L113), as that is an ideal place to put it. After it has been instantiated it assumes that the scene node is a `DialogicBackground`

Jowan-Spooner: True. I've moved the check to an earlier moment.

PR #1960, opened 2023-12-17
Restore loading a save when using Styles (#1959)

### LINKED ISSUE #1959: When using Styles, loading a save doesn't complete
## The problem

**Describe the bug**
When loading a save, the UI refresh is incomplete, and the dialogue is stuck and not all resources are correctly updated.

What is expected when
image

**To Reproduce**
1. Create a style
2. Create a Timeline
3. Create a few control buttons in a topmost Canvas Layer for ease of use (Start/Stop/Save/Load)
4. Start the timeline and save at any point
5. Click the Load button
6. See error

**Expected behavior**
The display should be identical to the standard timeline run.

**Screenshots**
Expected (this is an actual timeline screenshot):
image

After load:
image

The background is not shown, the text isn't displayed and the "next" indicator isn't visible.

 - OS: Windows 10
 - Godot Version: 4.2stable
 - Dialogic Version: [e.g. 1.0]

## Solutions
Code changes in the DialogicGameHandler.gd, subsystem_background.gd files (see possible fixes)
Additional code changes in subsystem_styles.bg that applies for any style change with background may be needed too.

**Workaround**
None.

What you tried to get the feature working

**Possible fixes**

***DialogicGameHandler.gd:***

In method load_full_state:

1. The 'Styles' subsystem is referred to as 'Style'. This may be from an older version of Dialogic.
2. The call to "start_timeline" comes too early. If we compare to the "start" method, the timeline must be loaded after the layout is fully loaded and ready.

These two changes restore most of the functionality of the load function.

***subsystem_background.gd***

The previous fix does not fully solve the issue. The background still won't show until there is a background change in the timeline.
The root cause of the issue is that a new Layout scene is loaded and until the previous layout is fully unloaded, which may happen later, there are multiple backgrounds registered as "dialogic_background_holders".

There are multiple solutions to this issue. The most trivial one would be to prevent switching layout, but that may not be what is expected depending on the save. The second solution implies making certain that we target the right layout scene when loading resources.

In method update_background:

1. Make sure that the DialogicNode_BackgroundHolder control that we target is the right one, that is the one belonging to the newly created style.
2. There is an optimization that doesn't work when loading a save. At line 50, there is a check that the argument parameter is different from the current state of the background. However the current state is used as the argument, which makes the method exit without having loaded the background. Removing this check makes the background being able to be loaded again.

Implementing the first change requires filtering the get_tree().get_first_node_in_group() result with only result(s) that belong to the new layout instance. The filtering could be done directly in the update_background method, but I suggest putting this filtering in the Style subsystem as it is may be a global style related need. However some users may want to remove the Style subsystem if they are certain to use only one dialogue style in their app. Therefore it is mandatory to check if the Styles subsystem is used and use the global method in this case. The code could even be made reusable by putting it in a new method in DialogicGameHandler.gd, but I did not see any need for it elsewhere.

Implementing the second change is quite straightforward at first glance. Just remove lines 50-51 in subsystem_background.gd solves the issue. This may cause however an optimization issue. A better solution may be to add an optional parameter to the update_background method to force updating of the background, in this case the background would be updated no matter what.

I have already done most of this changes locally, and I will submit a pull request with my own implementation of the changes.

sudisk:
Fixes #1959

### DIFF
--- a/addons/dialogic/Modules/Background/subsystem_backgrounds.gd
+++ b/addons/dialogic/Modules/Background/subsystem_backgrounds.gd
@@ -19,7 +19,7 @@ func clear_game_state(clear_flag:=DialogicGameHandler.ClearFlags.FULL_CLEAR):
 
 
 func load_game_state(load_flag:=LoadFlags.FULL_LOAD):
-	update_background(dialogic.current_state_info.get('background_scene', ''), dialogic.current_state_info.get('background_argument', ''))
+	update_background(dialogic.current_state_info.get('background_scene', ''), dialogic.current_state_info.get('background_argument', ''), 0.0, default_transition, true)
 
 
 ####################################################################################################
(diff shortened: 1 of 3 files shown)


Jowan-Spooner:
Thanks. The fixes look reasonable. I will test them tomorrow after I got back home.

Jowan-Spooner:
Thanks! I adjusted the code a little, mainly to follow the gd-script styleguide and to make it a bit simpler.

sudisk:
Thank you for accepting my proposed changes. Your version looks much cleaner.

PR #1974, opened 2023-12-28
Update DialogicAnimaResources.gd

TheMeddlingMechanic:
Hi,
This is theMeddlingMechanic from the dialogic discord (sorry for taking a while). I've added a single line of of code (line 39) on the DialogicAnimaResources.gd script. This is to account for developers who use PckEncryption in their game builds.

### DIFF
--- a/addons/dialogic/Nodes/Anima/DialogicAnimaResources.gd
+++ b/addons/dialogic/Nodes/Anima/DialogicAnimaResources.gd
@@ -36,9 +36,7 @@ static func get_available_animations() -> Array:
 
 	for file in list:
 		if file.find('.gd.') < 0:
-			filtered.push_back(file.replace('.gde', '.gd'))
-			filtered.push_back(file.replace('.gdc', '.gd'))
-
+			filtered.push_back(file.replace('.gde', '.gd').replace('.gdc', '.gd'))
 	return filtered #+ _custom_animations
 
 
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
@zaknafean This seems like a pretty simple PR, anything you need to test before we can merge this?

zaknafean:
Unfortunately it may break non encrypted builds as written. I gotta add a branch of some kind that can determine  if you're encrypted or not.

TheMeddlingMechanic:
Hi guys, I believe this shouldn't cause issues like that as it adds a new line, replacing both .gdc and .gde with .gd in the saved list. Feel free to test it, but if I remember correctly, that should work.

Jowan-Spooner:
Hmmm... I'm unable to test this as I don't have a dialogic 1 environment anymore. 
Wouldn't it need to be something like
```
filtered.push_back(file.replace('.gde', '.gd').replace('.gdc', '.gd'))
```
So any file is only added once and depending on what "wrong" extension it has they get corrected?

TheMeddlingMechanic:
Hmm, this could be the case, I'm not home at the moment, but when I am I'll record myself testing this for both encrypted and not (I can print out the list and check the logs). I'll get back to you with the recording once I've got an answer.

TheMeddlingMechanic:
Actually, I doubt this will be an issue, as every script in the pck should be the same type (gdc vs gde), so only one of these lines of code should make a difference,
but I'll still test it once I have the chance, just in case.

TheMeddlingMechanic:
I didn't record this in the end, just added another commit because of the obvious meaning of the docs for that method.

zaknafean:
Confirmed it still works on non-encrypted exports just fine. Thanks for the PR @TheMeddlingMechanic , merging it in now.

PR #1980, opened 2024-01-01
Add Unit Test GitHub Workflow

CakeVR:
This adds the `gdUnit4` framework as a single GitHub workflow CI file to the project.
It fires on new PRs and pushes.

However, I cannot properly test it without trying it once merged thought.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -28,11 +28,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        godot-version: ['4.1.1'] # Insert here the Godot version you want to run your tests with
+        # Insert here the Godot version you want to run your tests with.
+        godot-version: ['4.2.0']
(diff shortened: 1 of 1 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,114 @@
...
+  group: unit-test${{ github.event.number }}
+  cancel-in-progress: true
+
+
+jobs:
+  unit-test:
+    strategy:
+      fail-fast: false
+      matrix:
+        godot-version: ['4.1.1'] # Insert here the Godot version you want to run your tests with
```
coppolaemilio, line 31: Shouldn't we use Godot 4.2 instead?

CakeVR: You are right, I updated the version!

coppolaemilio:
Great! Hope it works :)

PR #1991, opened 2024-01-05
Fix empty-choice-indentation bug

Jowan-Spooner:
Reported by a user on discord, having an empty choice before an indentation change would incorreclty calculate the amount of end branches needed.

Example timeline of failure:
```
if {RedWandCount} > 0 or {BlueWandCount} > 0 or {GreenWandCount} > 0:
	The pedestal is empty.
	- Place red wand [if {RedWandCount} > 0][else="hide"]
	- Place blue wand [if {BlueWandCount} > 0][else="hide"]
	- Place green wand [if {GreenWandCount} > 0][else="hide"]
else:
	The pedestal is empty and you can't do anything about it.
The end
[end_timeline]
```
Would fail previously, but now works.


coppolaemilio:
Thanks!

Issue #2001, opened 2024-01-08
Repeated Background Music Event restarts Music


CakeVR:
**Is your feature request related to a problem? Please describe.**
If you start playing music A and reach another event that starts playing music A as well, music A restarts.

**Describe the solution you'd like**
Sometimes, when jumping or looping, the music playing is important instead of the music starting from anew.
I want the Music Event to have an option to prevent restarting the same audio resource from playing from its beginning.

Adding a property to the event "Restart if Resource plays already": `true` and `false` (default).

**Alternative Solution:**
I cannot think of a good reason when people may want to restart the audio to begin with, changing the behaviour may work as well.

Jowan-Spooner:
Yeah I think for now I would just change the behaviour. I want to avoid any setting which I'm not sure people will actually use.

CakeVR:
I will take a look a this then!

PR #2002, opened 2024-01-08
Add check preventing restarting the same audio resource.

CakeVR:
Implements #2001 and fixes the background music event when its restarting the identical playing audio resource.

### DIFF
--- a/addons/dialogic/Modules/Audio/event_music.gd
+++ b/addons/dialogic/Modules/Audio/event_music.gd
@@ -1,8 +1,8 @@
 @tool
+## Event that can change the currently playing background music.
+## This event won't play new music if it's already playing.
 class_name DialogicMusicEvent
 extends DialogicEvent
-
-## Event that can change the currently playing background music.
 
 
 ### Settings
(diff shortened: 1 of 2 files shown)


CakeVR:
> Seems a bit unnecessary to do the check in the event, couldn't we simply check it in the subsystem?

Personally, I think game logic should happen in events and the subsystem for audio executes the request. \
If someone would like to restart the same audio, they can still make a new event, however, if the subsystem refuses to behave like that, the feature would be impossible; they need to handle their own audio player.

addons/dialogic/Modules/Audio/event_music.gd
```diff
@@ -24,10 +24,21 @@ var loop: bool = true
...
+
+	if not _is_playing_resource_already():
+		dialogic.Audio.update_music(file_path, volume, audio_bus, fade_length, loop)
+
 	finish()
 
 
+## Returns whether the currently playing audio resource is the same as this
+## event's [member file_path].
+func _is_playing_resource_already() -> bool:
```
Jowan-Spooner, line 36: I don't really see why this needs to be a separate method. Is there any case in which one would call this manually on the event? 

I think it would make more sense to either move this method (a bit generalized) to the audio subsystem or desolve it alltogether.

CakeVR: I created this method to keep `_execute` as minimal and simple as possible.
I will add this to the subsystem!

PR #2024, opened 2024-01-16
Fix User Input Auto-Advance

CakeVR:
Fixes Auto-Advance enabled until the next user input not ending on user input.

Fixes the PR https://github.com/dialogic-godot/dialogic/issues/2022.

This PR fixes unit tests relying on the Dialogic `autoload`.

### DIFF
--- a/addons/dialogic/Tests/Unit/test_auto_advance.gd
+++ b/addons/dialogic/Tests/Unit/test_auto_advance.gd
@@ -1,4 +1,3 @@
-class_name AutoAdvanceTest
 extends GdUnitTestSuite
 
 ## Ensure Auto-Advance is enabled properly using the user input flag.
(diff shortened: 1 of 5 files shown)


addons/dialogic/Tests/Unit/test_auto_advance.gd
```diff
@@ -0,0 +1,20 @@
+class_name AutoAdvanceTest
```
Jowan-Spooner, line 1: Does this really need a class name?

CakeVR: Unsure, I just follow what the gdUnit4 examples do! I can research that.

CakeVR: It's not needed, I will remove it then.

Jowan-Spooner:
Lgtm

PR #2025, opened 2024-01-18
fix a function bug in subsystem_variables.gd

endaye:
image

image

When setting a dialogic variable in code, it shows a bug here after `alpha 12`.

The method called `_set_value_in_dictionary` in `outside` as `VariableFolder` has been removed since commit [Variables Rework (second PR attempt) (#1972)](https://github.com/dialogic-godot/dialogic/commit/ccd5e3385db2366bcb8c7eb3702070eef25adedd)

I think this is a missing part.


Jowan-Spooner:
Thank you very much! This was indeed a mistake I made.

PR #2052, opened 2024-01-26
Avoid null get_tree() call

Invertex:
get_tree() can return null when the Dialogic view isn't open but this function is being called by an external system. No node tree exists to wait for or manipulate, so errors are spammed.


Jowan-Spooner:
Thanks, LGTM.

PR #2054, opened 2024-01-28
Allow any Texture type to be assigned as NextIndicator icon

Invertex:
Allow .tres assignment to NextIndicator so things like AnimatedTexture can be used and any other texture time.

Also made the TextureRect conditional checking more clear and consistent too.


Jowan-Spooner:
Thanks. LGTM!

PR #2068, opened 2024-02-03
Fix variable and setting nodes not updating on type change

### LINKED ISSUE #2062: Variable Random Value returning < null >
### Discussed in https://github.com/dialogic-godot/dialogic/discussions/2061

<div type='discussions-op-text'>

<sup>Originally posted by **tryinganon** February  2, 2024</sup>
I was advised to make a dedicated thread for this issue after mentioning it in a previous thread. (Sorry if I'm doing this wrong, new to GitHub)

So, I did a totally fresh install on Alpha 13 after losing a project utilizing Alpha 12, I know it's definitely not an update problem or anything. It's always possible it's a local issue but with a totally fresh install it's unlikely. I also have the latest version of Godot (4.2.1)

When I try to set an expression to a random value, it returns null.

Easy way to recreate it:

- Make a new file
- Create any number variable (default 0)
- Make a new timeline
- Set the variable to a random value, if you don't want to use breakpoints then I recommend "to itself plus random" so the problem will be explained in a regular error message
- Use a conditional branch in the timeline based on the variable's value

When you run it, you'll get the error message: Dialogic: Set Variable event failed because one value wasn't a float! [0, < null >]

(Using breakpoints on the relevant lines will also reveal the value is null but I don't feel like finding which lines that is again)

Alpha 12 does not have this problem (which also rules out my copy of Godot itself), I reinstalled it just to test that lol, I used the exact same steps there with no problem.

If anyone figures out a fix please let me know! I'll also share if I figure it out because I am tinkering with it myself, but I'm pretty new to Godot (I've used Unity and Unreal Engine before this) so my hopes aren't high.</div>

### LINKED ISSUE #2063: Setting a variable to 0 causes it to be treated as null
## The problem

**Describe the bug**
If you have a bool or float (possibly others as well but I've only tested these two) that is set to be between 0 and another number, and place setting it within a condition based on the value of another float, you will get the error: "Dialogic: Set Variable event failed because one value wasn't set!"

**To Reproduce**
1. Make two variables in a new file, one float and one of either a float or a bool (may work with others too, these are just the two I had the problem with)
2. In a timeline, set (one of) the float(s) to a number (I personally used a random, but this requires fixing line 59 in event_variable to be "1,3,4,5: the_value = dialogic.VAR.get_variable(str(value))" for random number generation to work. I did check later to ensure that wasn't the problem, though, it does work with setting a user-chosen value from a fresh install of the plugin.)
3. Create a condition based on that first variable (such as >50), and within the if, set the second variable to 1. Within an else, set it to 0
4. Run the timeline
5: "Dialogic: Set Variable event failed because one value wasn't set!"

**Screenshots**

image

**Expected behavior**

The second variable should be set to 1 (or another number) or 0 based on the conditional statement (including bools being set to true or false).

 - OS: Windows 10
 - Godot Version: [4.2.1]
 - Dialogic Version: [2.0 Alpha 13]

## Solutions

**Workaround**

Setting it to 2 or 1 rather than 1 or 0 fixed it, and, interestingly, upon setting it back to 1 or 0, the error no longer happened, even with new variables. A fresh project, or setting it to a bool in the variable's tab (and even if that's then set back to a float), causes it to happen again.

**Possible fixes**

Line 54: if value and orig != null: in event_variable.gd is returning false. Using breakpoints, I have determined that "value" is the issue. It appears to be failing to get "value" earlier in the code, thus causing it to be null. Unfortunately, despite my best efforts, I can't seem to figure out why that is. I suspect it has to do with storing the variable's default from the variable's tab, though, because I noticed that when reloading the project, "set float to be zero" in the timeline changed to "set any to be (blank)". while "set float to be 1" was fine.

Whatever this is seems likely related to the addition of bools as a variable option, so, whatever aspects of code you guys changed for that is a good start I think?

(Also, I did get it to happen once or twice without the "conditional statement" aspect to this, but it was REALLY inconsistent and after flipping between a bool and a float a few times I couldn't get it to happen again with the same file anymore without the conditional part?? So idk, I wanted to give the most consistent method to trigger it)

(Let me know if I can provide more info or otherwise help)

### LINKED ISSUE #2067: Set var node not updating timeline when set value type is changed.
## The problem

**Describe the bug**

**To Reproduce**
1. Create variable `test_var` of any type
2. Create new timeline
3. Add `Set variable` node to timeline
4. Set `test_var` = `number` 0
5. Save
6. Switch to text editor and see that the timeline is correct
7. Change the set var type to a `bool`
8. Save
9. Switch to text editor and see that the timeline is still using the `number` and not the `bool`

**Expected behavior**
The timeline should be dirtied if the set var type is changed.

**Screenshots**
var_value_type_update

 - OS: Windows
 - Godot Version: 4.2.1.stable.official
 - Dialogic Version: 2.x (main)

## Solutions

**Workaround**
Changing the set var type and then changing the value will force a refresh.

**Possible fixes**
Unsure atm.

Atlinx:
Fixes #2067 
Fixes #2062 (Uses the proposed change in the issue)
May also fix #2063 , since I suspect a null value might be the product of the visual editor not generating the correct code

Changes
- Made _value_type reset the value and force an editor refresh on change
  - Solves the issue of #2067 where changing the value type does not reset the value
- Refactor `event_variable` and `event_settings` to use enums for their different value types to make the code easier to read
- Refactor `event_block` to not call `set_value` on fields that are not visible. This prevents invisible fields from throwing errors when they get injected with the wrong data type in event nodes that have varying value types (such as `event_variable` and `event_settings`, whose value types can be strings, numbers, expressions, etc.)

### DIFF
--- a/addons/dialogic/Editor/Events/EventBlock/event_block.gd
+++ b/addons/dialogic/Editor/Events/EventBlock/event_block.gd
@@ -249,7 +249,16 @@ func build_editor(build_header:bool = true, build_body:bool = false) ->  void:
 
 		# Some things need to be called AFTER the field is added to the tree
 		if editor_node is DialogicVisualEditorField:
-			editor_node._set_value(resource.get(p.name))
+			# Only set the value if the field is visible
+			#
+			# This prevents events with varied value types (event_setting, event_variable)
+			# from injecting incorrect types into hidden fields, which then throw errors 
+			# in the console.
+			if p.has('condition') and not p.condition.is_empty():
+				if _evaluate_visibility_condition(p):
+					editor_node._set_value(resource.get(p.name))
+			else:
+				editor_node._set_value(resource.get(p.name))
 
 			editor_node.value_changed.connect(set_property)
 
@@ -305,9 +314,8 @@ func recalculate_field_visibility() -> void:
 			if p.location == 1:
 				has_any_enabled_body_content = true
 		else:
-			var expr := Expression.new()
-			expr.parse(p.condition)
-			if expr.execute([], resource):
+			var visible = _evaluate_visibility_condition(p)
+			if visible:
 				if p.node != null:
 					p.node.show()
 				if p.location == 1:
@@ -315,8 +323,6 @@ func recalculate_field_visibility() -> void:
 			else:
 				if p.node != null:
 					p.node.hide()
-			if expr.has_execute_failed():
-				printerr("[Dialogic] Failed executing visibility condition for '",p.get('property', 'unnamed'),"': " + expr.get_error_text())
 	%ExpandButton.visible = has_any_enabled_body_content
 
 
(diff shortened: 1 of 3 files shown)


addons/dialogic/Editor/Events/EventBlock/event_block.gd
```diff
@@ -305,18 +314,15 @@ func recalculate_field_visibility() -> void:
 			if p.location == 1:
 				has_any_enabled_body_content = true
 		else:
-			var expr := Expression.new()
-			expr.parse(p.condition)
-			if expr.execute([], resource):
+			var visible = _evaluate_visibility_condition(p)
```
Jowan-Spooner, line 317: Could you add static typing here?

Jowan-Spooner (commented):
Added a tiny comment. Looks great otherwise. I want to take a moment to actually test this and don't have the time right now (will hopefully get to it later or tomorrow), but it looks really good. Thank you so much already.

addons/dialogic/Editor/Events/EventBlock/event_block.gd
Atlinx: Does removing it altogether work as well, since I've used the function directly in all the other cases?

Jowan-Spooner: sure!

Jowan-Spooner:
This seems to work perfectly. Thank you very much!

PR #2072, opened 2024-02-06
Wait one frame before clean to avoid disconnecting internal ONESHOT connections

### LINKED ISSUE #2069: Debugger error after timeline ends when started from "Play Timeline" mode
## The problem

**Describe the bug**
When starting a timeline using the "Play Timeline" button, after the timeline ends, the debugger throws an error. This seems to be harmless, but cleaning up the error would make it more intuitive

This is the error thrown by the debugger:
```
E 0:00:05:0177   event_text.gd:215 @ _on_dialogic_input_action(): Attempt to disconnect a nonexistent connection from '[Text:-9223371957532096845]'. Signal: 'advance', callable: 'GDScriptFunctionState::_signal_callback'.
  <C++ Error>    Condition "!s->slot_map.has(*p_callable.get_base_comparator())" is true. Returning: false
  <C++ Source>   core/object/object.cpp:1420 @ _disconnect()
  <Stack Trace>  event_text.gd:215 @ _on_dialogic_input_action()
                 subsystem_input.gd:79 @ handle_input()
                 node_input.gd:21 @ _on_gui_input()
```

**To Reproduce**
1. Create a timeline
2. Add a text node
3. Press "Play Timeline" and keep on clicking until the timeline is finished
4. See the error in the debugger tab

**Expected behavior**
No errors should appear when the timeline is finished

**Screenshots**
N/A

 - OS: Windows
 - Godot Version: 4.2.1.stable.official
 - Dialogic Version: 2.x (main)

## Solutions

**Workaround**
N/A

**Possible fixes**
Not sure yet.

Jowan-Spooner:
- Should fix #2069

Godot seemed to try to disconnect a signal we had already cleared (in timeline.clean()). Unfortunately it seemed pretty hard to differentiate these signals from other signals (because I wasn't allowed to access it's object). Luckily it is solved by simply waiting a frame because godot will have removed the connection on the next frame.

@Atlinx would love if you could try this!

### DIFF
--- a/addons/dialogic/Core/DialogicGameHandler.gd
+++ b/addons/dialogic/Core/DialogicGameHandler.gd
@@ -272,7 +272,11 @@ func clear(clear_flags:=ClearFlags.FULL_CLEAR) -> bool:
 
 	# Resetting variables
 	if current_timeline:
+		# This is necessary because otherwise INTERNAL GODOT ONESHOT CONNECTIONS
+		# are disconnected before they can disconnect themselves.
+		await get_tree().process_frame
 		current_timeline.clean()
+
 	current_timeline = null
 	current_event_idx = -1
 	current_timeline_events = []


Atlinx:
Yep, the timeline works now with no errors, even when ending on a text event.

Jowan-Spooner:
Thank you!

Issue #2076, opened 2024-02-06
Character event node: Selecting "(No one)" does not close suggestions box.


Atlinx:
## The problem

**Describe the bug**
In the character event node, selecting "(No one)" for the character identifier does not close the suggestions box.

**To Reproduce**
1. Create a timeline
2. Create a character node
3. Click on the dropdown and click on "(No one)"
4. Suggestions box does not disappear

**Expected behavior**
Suggestions box should disappear

**Screenshots**
![not_closing](https://github.com/dialogic-godot/dialogic/assets/25368491/25821e26-1b2a-4353-a181-e163b2b8a4f8)

**System (please complete the following information):**
 - OS: Windows
 - Godot Version: 4.2.1.stable.official
 - Dialogic Version: 2.x (main)

## Solutions

**Workaround**
N/A

**Possible fixes**
Removing `current_value == ""` from the `_on_search_focus_entered` function of `field_options_dynamic`. There doesn't seem to be a need to check the current_value if the search is driven by the search text field (`%Search`) and not the value.

Issue #2066, opened 2024-02-03
Dynamic option fields don't scale width correctly on hiDPI displays


Atlinx:
## The problem

**Describe the bug**
Dynamic option fields don't scale the width of the recommendations box correctly on hiDPI displays. It seems to underestimate the width, leading to text clipping.

**To Reproduce**
1. Launch Godot on a computer with a hiDPI display
2. Add a variable named "test_var"
4. Add a `Set Variable` node
5. Click on the dynamic option field for setting the variable name, and see the recommended box clipping the text

**Expected behavior**
The width should be wide enough to not clip text.

**Screenshots**
![image](https://github.com/dialogic-godot/dialogic/assets/25368491/0e88be09-13a1-449b-be74-6375e0de8265)

**System (please complete the following information):**
 - OS: Windows
 - Godot Version: 4.2.1.stable.official
 - Dialogic Version: 2.x (master)

## Solutions

**Workaround**
N/A

**Possible fixes**
The code for `field_options_dynamic` might need to apply editor_scale somewhere in the line_length calculation.

PR #2077, opened 2024-02-06
Fix hiDPI scaling for dynamic options + empty value triggering suggestions

Atlinx:
Fixes
- Fixes hiDPI scaling for dynamic options. #2066 (Supercedes #2073)
- Fixes dynamic options showing suggestions when set with an empty value #2076 

Code Changes
- Resized `portrait.svg` to 16 x 16
- Make Suggestions box use a fixed icon size of 16 x 16 (Scaled to your editor size)
- Calculates suggestion box size using `line_height`, `v_separation`, `h_separation`, icon texture size, and `icon_margin` instead of using a magic number.
- Update the height and width on every change
- Update suggestions box width on the next frame in case PanelContainer changes its width 
- Removes `or current_value == ""` from the `_on_search_focus_entered` function, since searching shouldn't depend on `current_value`
  - Sometimes the value passed into the field is meant to be empty (""), so we don't want to show the suggestions box in that case.
  - **Ex.** when we want a "(No one)" character_identifier, its actually represented as an empty ("") value.


Jowan-Spooner:
Hey @Atlinx thanks again for the PR. I tried giving some feedback on discord, but I'll post it here again.

When testing your HiDPI PR on my (not HiDPI) screen:

- the suggestions box now sometimes cuts off the suggestions.
grafik

- Also as I was trying to do in my PR the minimum width should be the width of the whole field so we don't get a very short suggestions box
grafik

- the resizing sometimes does some strange stuff (see video)

https://github.com/dialogic-godot/dialogic/assets/42868150/5355047d-bb18-4e10-8c79-7080fc80f7c3

*The video doesn't work on firefox for me idk why, on chrome it works*


Also for this

> Removes or current_value == "" from the _on_search_focus_entered function, since searching shouldn't depend on current_value

While I'm okay with changing this (to fix the box not closing), it's wasn't really a bug but an intended qol feature. It's because it's very typical for these "empty" values to be the defaults and thus to be changed a lot. This line means you can simply click anywhere in the field to open the suggestions in these special cases where it is very likely you want to change them. 
I'm open for removing this though or finding a different solution, just wanted to say it wasn't really a mistake.

PR #2094, opened 2024-02-17
Vector types and number improvements

Invertex:
Added Vector3 and Vector4 types, unified under a common base, so people making custom nodes have a unified way to write that data. Often people will want to use nodes that manipulate 3D space data, like the Camera or 3D character positions for example, so Vector3 is important.

Updated some of the code for the Number field as well to support more display customization.

Main changes:
- Vector3/4 types for custom events to use
- Value adjustment arrows split vertically to allow independent events and highlighting.
- Integrated dedicated Affix/Suffix labels into the Input field that are separate from the actual value and static
- Integrated the value adjustment arrows into the input field visuals as well, keeping everything tied together.
- Value box and arrows respond to arrow keys to adjust value accordingly
- Key or mouse can be held to start a continuous adjustment of value
- Adjustment arrow direction will disable itself when hitting a Min or Max value to visually indicate that limit was hit.
- `"hide_step_button"` extras_option to hide the Adjustment arrows
- `"affix"` & `"suffix"` extras_option inputs for an event to define those labels


CakeVR:
Hello, thanks for the PR.

I would prefer if we keep the arrows as they are for now and discuss the styling in an issue or the Discord.

Invertex:
No problem, removed the changes to the arrows.

addons/dialogic/Editor/Events/Fields/field_number.gd
```diff
@@ -131,7 +131,7 @@ func _on_decrement_button_down(button : NodePath) -> void:
 func _on_value_text_submitted(new_text:String, no_signal:= false) -> void:
 	if new_text.is_valid_float():
 		var temp: float = min(max(new_text.to_float(), min), max)
-		if !enforce_step or is_equal_approx(temp/step, round(temp/step)):
```
Invertex, line 84: This being `is_equal_approx` meant numbers like `0.4300101204999`  would show in the fields despite `enforce_step` being on, since it is `approx_equal` to `0.43`. Further adjustments to the number would retain this large number, since it kept staying within valid stepping.

PR #2108, opened 2024-02-21
Show used Locales in Test Locale Dropdown

CakeVR:
At the moment, the Test Locale in the Translation editor displays all possible locales.
This change will display all *unique* locales found in any of the CSVs.


Jowan-Spooner:
This looks okay and works, so I'm fine with merging it, but I wonder if it's necessary to do all this in the CSV file class. You could alternatively (after "Collect Translations") simply check the created .translation resources for their locales I think (either from their path or loading the resource and calling get_locale()). 

However this is just a thought and I don't know if it's actually simpler.

PR #2114, opened 2024-02-25
Fixed error 'Set Variable event failed' when setting bool false.

### LINKED ISSUE #2112: A change to the way false boolean variables are assigned
I tried to set a Boolean variable to false through the dialogic add on in this way: 
image
image

but received this error: 
image

in order to set the boolean variable to false i put this into the text editor:
image

which resulted in this in the visual editor:
image

I would like a false boolean variable to be easier to set in both the visual and text editor, because the current way is not quite intuitive.   I think the checkbox and the on/off slider make sense visually, it might just be a processing error.

workhorsy:
This bug is in Main and 2.0-Alpha-13 (Godot 4.2+)

This will also fix #2112 .

Basically you can't set a bool Dialogic variable from true to false. This:

Screenshot from 2024-02-24 23-00-16

Will create the error:
```
printerr("Dialogic: Set Variable event failed because one value wasn't set!")
```

This is because the if statement on line 77 fails, because the variable value is false. In order to keep the same behavior, we can check that value is Truthy or equal to false.

### DIFF
--- a/addons/dialogic/Modules/Variable/event_variable.gd
+++ b/addons/dialogic/Modules/Variable/event_variable.gd
@@ -74,7 +74,7 @@ var _suppress_default_value: bool = false
 func _execute() -> void:
 	if name:
 		var orig :Variant= dialogic.VAR.get_variable(name)
-		if value and orig != null:
+		if (value == false or value) and orig != null:
 			var the_value :Variant
 			match _value_type:
 				VarValueType.STRING: the_value = dialogic.VAR.get_variable('"'+value+'"')


coppolaemilio (approved):

Jowan-Spooner:
Thanks for the PR!
I would rather fix the whole thing to check for null (and actually set value to null in the from_text method if nothing is given). Because wouldn't this mean it doesn't allow setting to 0 either? Also this triggers an error if the value is anything else then a bool (e.g. a string, etc.).

Also I feel like I have removed this or a similar check before. Might never have pushed it idk 🙈. I will push an improvement to this branch in a moment.

PR #2119, opened 2024-02-27
Add Layered Portraits

CakeVR:
Adds support for layered portraits to Dialogic.

If you want to get started trying this Pull Request, take a look at this WIP doc page: https://github.com/dialogic-godot/documentation/blob/38df52f0e6d0ba19ebd1705aa13a4d4a7553952e/documentation/layered-portrait.md

Here is the new command syntax available, we need to instruct layer changes via `extra_data` shortcode parameters.
```gdscript
# Show an entire group:
show Group1

# Show a layer, does not hide others:
show Group1/Layer1

# Hide a group:
hide Group1

# Hide a layer:
hide Group1/Layer1

# Show only one layer and hide the others:
set Group1/Layer1
```

Examples in the timeline:
```gdscript
join princess (default) 2 [animation="Heartbeat" extra_data="show Group1"]

update princess [extra_data="set Group1/Layer1"]]
```

The syntax after the command names (`hide`, `show`, `set`) is based on the valid syntax for node hierarchy paths. The different commands can be executed in a single `extra_data` by using the `,` (comma) delimiter.

```gdscript
update princess [extra_data="set Group1/Layer1, set Group2/Layer3"]]
```

You can access this node path syntax by right-clicking the target group/layer and clicking `Copy Node Path`. However, it's a very simple syntax to learn.

## Technical Details

This pull requests fixes the `update` event variant and allows it to set the extra data.

It also corrects the behaviour of the Portrait subsystem. Adding support to properly fade portraits in and out.

Fade Animations doing `XYZ In` and `XYZ Out` were unified as `XYZ In Out`, using the `is_reversed` (POV: Fade In) member on the base.

`guess_special_resource` supports searching for these `In Out` animations by just declaring `In Out`.

Leaving Animations are automatically set to be `is_reversed` `true`.

The Animation Base provides `get_modulation_property` as method, to easily access the correct modulation property on `CanvasGroup`s. `self_modulate` affects the entire internal viewport image instead of individual parented nodes.

The Portrait Settings Tab supports a default Transition Animation. If using the `Change` event on unique scenes (e.g. two Layered Portraits), it will automatically resort to this animation.


CakeVR:
This Pull Request is ready for review.

Jowan-Spooner:
Alright, let's do this.

PR #2124, opened 2024-03-04
Refactor Manual-Advance to `Dialogic.Inputs.manual_advance`

CakeVR:
Evens out the difference between Auto-Advance, Auto-Skip, and Manual-Advance by turning it into a property on the `Inputs`-subsystem.

This has a breaking impact on the API and results in the following changes:

Setting the enable state of Manual-Advance:
```gdscript
# Old:
Dialogic.Inputs.set_manualadvance(true, false)

# New:
Dialogic.Inputs.manual_advance.enabled_forced = true
Dialogic.Inputs.manual_advance.enabled_until_next_event = false
```

Getting the enable state of Manual-Advance:
```gdscript

# Old:
var result := Dialogic.Inputs.is_manualadvance_enabled()

# New:
var result := Dialogic.Inputs.manual_advance.is_enabled()
```

### DIFF
--- a/addons/dialogic/Modules/Core/subsystem_input.gd
+++ b/addons/dialogic/Modules/Core/subsystem_input.gd
@@ -13,16 +13,18 @@ var action_was_consumed := false
 
 var auto_skip: DialogicAutoSkip = null
 var auto_advance : DialogicAutoAdvance = null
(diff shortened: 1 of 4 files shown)


Jowan-Spooner:
Thank you, it's great having them all match!

PR #2128, opened 2024-03-07
Add support for Alias Glossary Dictionary Keys

CakeVR:
Right now, if there is an alias key used in a glossary `Dictionary`, erasing encountered a type-match crash.

This will handle the `String` alias entry keys.


Jowan-Spooner:
Not entierly sure how to test this, but the code looks good enough to me. Thanks!

Issue #2131, opened 2024-03-08
`EditorInterface` in `DialogicUtil` fails compiled Builds


CakeVR:
The `update_autoload_subsystem_access` method invokes this to reimport `EditorInterface.get_resource_filesystem`, in order to reimport files after the custom subsystems were injected. 

Using an early-return if outside of the editor does not work, the sole existence of the `EditorInterface` will cause many errors to appear in the binary console window after a build.

This issue exists on commit 75ac265490b62e0426c95fd58134dc36d67eadc3.

If someone wants to fix this locally, just remove the line in question: https://github.com/dialogic-godot/dialogic/blob/75ac265490b62e0426c95fd58134dc36d67eadc3/addons/dialogic/Core/DialogicUtil.gd#L78

rberends:
Thanks :) This broke an export of mine- removing the line bypasses this issue.

PR #2133, opened 2024-03-08
Prevent `EditorInterface` breaking Builds

CakeVR:
This identifier caused builds to breaks, as it does not exist at runtime outside the Godot editor.

Fixes: #2131 

Thanks to @lyuma for finding a workaround handling the `EditorInterface`.


Jowan-Spooner:
God damnit godot. I hate it when they give you nice features and then are like "well don't use them if you want your game to actually work". Thanks for the fix. Looks good to me.

PR #2136, opened 2024-03-11
Fix history not evaluating variables in names

### LINKED ISSUE #2127: History shows variable names instead of values.
## The problem

**Describe the bug**
When using variables for character's names, the history function will show the variable instead of the name for the character saying the dialogue. It functions as intended for the speech part.

**To Reproduce**
1) Create a variable that you'll use for your character's name. I have done a folder called `mc` and a variable under it called `name`
image
2) Do some simple dialogue that uses `{mc.name}` and look at the history
image
3) See that `{mc.name}` is used in the character's name slot, but `Bob` is correctly used in the speech slot.

**Expected behavior**
The History should return the value of the variable instead of the variable.

**Screenshots**
See above

 - OS: Arch Linux
 - Godot Version: 4.2.1
 - Dialogic Version: 90a889c5ee99b889de861c7988befd96cd95276a

Atlinx:
Added `get_character_name_text` method to the text subsystem. This method evaluates and translates character's name and returns the final text that should be displayed. It's now used in the text event, character event, and in the `update_name_label` of the text subsystem.

Fixes #2127 

https://github.com/dialogic-godot/dialogic/assets/25368491/79452ce9-e4bc-4c39-a4e4-8bad22a6ca0c

### DIFF
--- a/addons/dialogic/Modules/Character/event_character.gd
+++ b/addons/dialogic/Modules/Character/event_character.gd
@@ -72,7 +72,8 @@ func _execute() -> void:
 		Actions.JOIN:
 			if character:
 				if dialogic.has_subsystem('History') and !dialogic.Portraits.is_character_joined(character):
-					dialogic.History.store_simple_history_entry(character.display_name + " joined", event_name, {'character': character.display_name, 'mode':'Join'})
+					var character_name_text = dialogic.Text.get_character_name_text(character)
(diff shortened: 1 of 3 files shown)


Jowan-Spooner:
Looks good to me. Please use static typing as outlined in the styleguide https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html. I've made the necessary changes and also named the method get_character_name_parsed (text was to generic for my taste ;)) and moved it to the helper section.

Thanks a lot for the PR!

PR #2137, opened 2024-03-12
Improve Choices Script and VN Text Box Layer

CakeVR:
Adds the `text_box` property to the `DialogicNode_ChoiceButton` class. If set, the instance will set the `text` property on the `text_box` instead.

Adds exports to the `VN Text Box Layer` to allow turning off each individual override group category. Also differentiates between the `Box Settings` and the `Box Animation Settings`' subgroup.


Jowan-Spooner:
Looks good to me, thanks!

PR #2138, opened 2024-03-12
improve ConditionPicker#is_too_complex

iscke:
Currently, `is_too_complex` requires `condition.count('==') == 1 and condition.count('>') != 1` etc for a condition to be complex, meaning you need something like `== != <= >=` in your expression for it to be counted as complex. This changes the requirement to being empty or having a single comparison operator, which is closer to what the simple condition editor and `complex2simple` use.


Jowan-Spooner:
Thanks, looks good to me!

PR #2146, opened 2024-03-21
don't emit `Dialogic.timeline_ended` until `Timeline#clean` has been called

iscke:
this fixes a bug where starting a timeline in a callback attached to `Dialogic.timeline_ended` will lock, since `Timeline#clean` fires a frame after it's called (and after `timeline_ended` is emitted), which will then disconnect the signals on the _new_ timeline.

(easy repro:
```gdscript
func _ready():
    Dialogic.start("whatever.dtl")
    Dialogic.timeline_ended.connect(Dialogic.start.bind("whatever.dtl")
```

causes the timeline to run once, and then to start a second time without input working.)


Jowan-Spooner:
Looks good to me. Thanks!

PR #2151, opened 2024-03-26
Add Tween Event

CakeVR:
This adds an event to tween values from timelines.

To do:
- [x] Background Support
- [ ] Portrait Support
- [ ] Camera Support
- [ ] Autoload Support
- [ ] Tween Pool

### DIFF
--- a/addons/dialogic/Editor/Events/Fields/field_number.gd
+++ b/addons/dialogic/Editor/Events/Fields/field_number.gd
@@ -7,8 +7,10 @@ extends DialogicVisualEditorField
 @export var allow_string : bool = false
 @export var step: float = 0.1
 @export var enforce_step: bool = true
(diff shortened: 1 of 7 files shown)


suppayami:
I think it would be nice to support loops (also endless loop, but might need to disable await?). Use case would be something:
- Breathing animation
- Flying around portrait
- Something like party frog? 
- ...

CakeVR:
> I think it would be nice to support loops (also endless loop, but might need to disable await?). Use case would be something:
> 
> * Breathing animation
> * Flying around portrait
> * Something like party frog?
> * ...

Good suggestions but I feature-froze this PR. They can be added in future PRs.

CakeVR:
We have found a bug where selecting `Character` and `offset` as property will completely crash Godot.

PR #2153, opened 2024-03-27
Remember the Timeline Sidebar Size

### LINKED ISSUE #2152: Remember the Timeline Right Sidebar Size
tab sizes dont get saved when you relaunch the godot engine

thats all

CakeVR:
When reloading the project, the right sidebar of the timeline editor does not remember its last manually set size.

This pull request saves the size as project setting, assigns the correct size to all relevant containers, and sets the proper event button style.

Closes #2152

### DIFF
--- a/addons/dialogic/Editor/TimelineEditor/VisualEditor/timeline_editor_visual.gd
+++ b/addons/dialogic/Editor/TimelineEditor/VisualEditor/timeline_editor_visual.gd
@@ -159,14 +159,14 @@ func _ready() -> void:
 
 	batch_loaded.connect(_on_batch_loaded)
 
-	#await find_parent('EditorView').ready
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
@CakeVR We have our own "editor settings" for this, stuff like this should not be stored in the project settings, because it's super annoying for vcs, especially in a team. 
You should use:
DialogicUtil.set_editor_setting()
DialogicUtil.get_editor_setting()

Also I'll have to take another look, but I had tried to implement this originally, but somehow it didn't work, not sure if there are any leftovers from that attempt.

CakeVR:
Thanks a lot, I use `DialogicUtil` now and it still works.

Jowan-Spooner:
There is now an error, because initialized is using two different spellings!

Jowan-Spooner:
Looks good to me. Thanks!

PR #2173, opened 2024-04-10
Parse Variables in Glossary Fields

### LINKED ISSUE #2172: Variables in Glossary Entries are not working
If a variable gets used in any glossary field, it stays `{name}` instead of getting replaced by the variable's value.

CakeVR:
This will parse the glossary fields for potential variables.

Closes #2172


Jowan-Spooner:
Looks good to me. I would like to move some of this functionality out of the layer, so that there is an easier way to just get the parsed and translated glossary info (because otherwise this has to be done in any glossary layer). But for now this is a good fix! Thanks.

PR #2176, opened 2024-04-11
Never disable the Colour Picker

### LINKED ISSUE #2175: Glossary Entries cannot change Colours
I can change the default colour, but the colour field on the Glossary Entry does not seem to function anymore.

CakeVR:
The Colour Picker was disabled, if the Glossary Entry was lacking the `color` entry. However, the Entries are kept minimal on purpose, if they use the default colour.

This removes the behaviour of disabling the Colour Picker.

Closes #2175


CakeVR:
There is a button that must be toggled, this was an oversight on my end!

PR #2229, opened 2024-05-14
Fixing relative margin scale

coppolaemilio:
The calculation of the margin based on the editor scale and the new godot editor theme were not properly set (used magic numbers). 

## Before
image

## After
image


coppolaemilio:
Should probably be good to start using `EditorInterface.get_editor_scale()` and replace `DialogicUtil.get_editor_scale()`. I'll add it to this same PR

PR #2232, opened 2024-05-16
Hot reload button for custom events

suppayami:
- [x] Update event cache
- [x] Reload event buttons on Timeline Right bar
- [x] Rebuild all EventBlock in Timeline

### DIFF
--- a/addons/dialogic/Editor/Events/EventBlock/event_block.gd
+++ b/addons/dialogic/Editor/Events/EventBlock/event_block.gd
@@ -334,6 +334,18 @@ func set_property(property_name:String, value:Variant) -> void:
 		end_node.parent_node_changed()
 
 
+func rebuild_editor(build_header:bool = true, build_body:bool = false) -> void:
+	body_was_build = false
+
+	for child in %HeaderContent.get_children():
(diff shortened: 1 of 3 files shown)


Jowan-Spooner:
Hey, thanks for the PR.
The implementation looks good, but I don't really see the point of the feature. The same can be achieved by switching to a different timeline or into text mode and back. This is really only needed when editing a custom events script, in which case you are not in the settings, so having the button in the settings seems like 2 more steps to me. MAYBE a reload button at the top of the visual timeline editor would make sense, but then again, I don't think it would add much. 

Sorry to discuss this after you've already implemented it, but for things like this I'd suggest making a feature request first, so we can discuss it. In this case, we can discuss it here though, idc.

suppayami:
@Jowan-Spooner Hi, it was my bad not discuss before making PR, I got into problem of event block not reloaded and found a thread on Discord discussed about reload button. 

It seems that with latest code in `main` the event block properly reloads without restart the editor now. I think this can be closed.

Jowan-Spooner:
Alright, thanks! Glad it works now ;)

PR #2233, opened 2024-05-17
Update subsystem history.

AnidemDex:
_Hello everyone, did you miss me?_

Update variable type for _result instead of letting the engine guessing the type. 

For some reason in 4.3 (https://github.com/godotengine/godot/issues/92021 probably) can't guess the type of _that very specific line_, maybe because can't even parse the referenced class so it just throw it as variant and let it work until fails?

This _kinda_ works (as is just a solution in place that solved Emi traceback). Tested with https://github.com/dialogic-godot/test-project exporting to linux, using Godot v4.3.dev6.official [64520fe67].

It may fix https://github.com/dialogic-godot/dialogic/issues/2231 ? I'm not sure, that traceback seems more related to Editor* classes being referenced in exported game

### DIFF
--- a/addons/dialogic/Modules/History/subsystem_history.gd
+++ b/addons/dialogic/Modules/History/subsystem_history.gd
@@ -65,7 +65,7 @@ func _update_saved_connection(to_connect: bool) -> void:
 	if to_connect:
 
 		if not DialogicUtil.autoload().Save.saved.is_connected(_on_save):
-			var _result := DialogicUtil.autoload().Save.saved.connect(_on_save)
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
Tbh I would prefer just removing the _result capturing. I think Cake added this, because they enabled very rigorous godot warnings, but I truly feel that at that point it's messing with the readability of the code if we were to actually capture the "result" everytime we connect a signal.

AnidemDex:
Totally forgot that you made this request. Thank you

Jowan-Spooner:
Np ;) Thank YOU!

PR #2235, opened 2024-05-17
Update CharacterResourceLoader and CharacterResourceSaver.

AnidemDex:
As #2234 , this just add very generic code to create an error handler on save/load of character resources (one of those already had half of one, I just updated it)


Jowan-Spooner:
Thanks!

PR #2237, opened 2024-05-17
Field Color for Event

### LINKED ISSUE #2236: Color field for custom event
## Description
Currently event fields don't support color picker. The use case would be something like tint, flash... portrait and background

suppayami:
Closes #2236 
- [x] Added `ValueType.COLOR`
- [x] FieldColor


Jowan-Spooner:
This looks very clean to me. Thank you very much. And congrats on your first commit to dialogic ;)! 🎉

PR #2252, opened 2024-05-23
Improve choice logic

Jowan-Spooner:
This separates the choice showing logic into three parts (that can be used somewhat independantly of each other):
- Collecting question information (choice subsystem)
- Showing choices (choice subsystem)
- Loading choice info (choice node!)

This also adds logic to store visited choices in the already_visited history, and adds info on whether a choice has been visited to the choice_info dictionary. This should make it pretty easy to change the style of choices, when they have already been visited.


Jowan-Spooner:
I guess I'll merge this now. I initially wanted to also add a feature where you could add more custom info to choices, but that can happen at a later point.

Issue #2256, opened 2024-05-29
Highlight Full Name when registered as a nickname


firedfromlife:
**Is your feature request related to a problem? Please describe.**
If a character's display name is a subset of a nickname, only the display name will be highlighted. This appears to only be the case when whitespace is used in the nickname. For example, the display name is just the first name, but the character's full name is used as a nickname to ensure it is also highlighted. See below example.

**Describe the solution you'd like**
I would like Dialogic to use a greedy Regex when searching for nicknames or display names. This would allow the longest available name to be selected even when it contains a shorter name that is a subset of another.

**Describe alternatives you've considered**
I have considered using the Display Name to contain the full name, but this looks very awkward when a character's full name is not how a player or their avatar would likely refer to a character. Just feels oddly formal.

**Additional context**
Example:

- **Display Name:** John
- **Nicknames:** John Smith, Johnny
- **Result:** Hello, I am _John_ Smith. Call me _Johnny_.
- **Expected Result:** Hello, I am _John Smith_. Call me _Johnny_.

Jowan-Spooner:
Hey, thanks for the report. 
I will look into this, it's a regex issue, and possibly also an issue of sorting these options by length first.

firedfromlife:
If you point me to the correct area to look at, I'd be more than happy to look into this and make a contribution.

PR #2258, opened 2024-06-01
Ensure all names are highlighted despite length

firedfromlife:
Solution to the following issue: 
#2256 

This change takes `collect_character_names()` and sorts the collected names by length, descending, prior to compiling the regular expression. This allows longer nicknames containing display names to function as expected. (See #2256 for an example)


Jowan-Spooner:
Thank you so much! Sorry it took me so long. Seems to work great. 
Also sorry you are having trouble with the portrait rn, hope we find the problem 🙈!

PR #2261, opened 2024-06-04
Add close button to reference manager

### LINKED ISSUE #2260: There is no way to close the reference manager window in i3
## The problem

**Describe the bug**
There is no way to close the reference manager window in i3. A Windows or Linux user would have a close button (X) in the title bar.

**To Reproduce**
1. Open the reference manager

**Expected behavior**
There should be a close button somewhere, maybe close to the "Check Selected" button.

**Screenshots**
image

 - OS: Arch Linux (i3 wm)
 - Godot Version: v4.2.2.stable
 - Dialogic Version: v2.0-Alpha14

## Solutions

**Workaround**

N/A

**Possible fixes**

Add a close button at the bottom.

I'll work on it myself.

juliohq:
This PR adds a close button to the reference manager popup, so it looks like this in a window manager (like i3):

image

Fixes #2260.

### DIFF
--- a/addons/dialogic/Editor/Common/reference_manager.gd
+++ b/addons/dialogic/Editor/Common/reference_manager.gd
@@ -7,6 +7,7 @@ func _ready() -> void:
 		return
 
 	add_theme_stylebox_override("panel", get_theme_stylebox("Background", "EditorStyles"))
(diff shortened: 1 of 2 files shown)


Jowan-Spooner:
Looks good to me. Thanks, I never realized 🙈.

PR #2263, opened 2024-06-04
Remove trailing whitespace before end of line

juliohq:
This PR removes all trailing whitespace (tabs, in this case) before end of line.


Jowan-Spooner:
Thanks!

PR #2264, opened 2024-06-04
Add missing void return types

juliohq:
This PR adds the missing `void` return types of several functions and callbacks.


juliohq:
In this specific case there is no need to convert `Array` into `PackedStringArray` actually, Godot figures it out automatically, maybe due to the built-in function signature.

Jowan-Spooner:
Alright. Thanks!

PR #2266, opened 2024-06-04
Fix static call warnings for `get_editor_interface()`

juliohq:
This PR fixes warnings caused by calling `get_editor_interface()` (`STATIC_CALLED_ON_INSTANCE`).


Jowan-Spooner:
I've not tested this one yet, but using EditorInterface directly usually results in errors on exported projects, as godot decides not to include the editor classes/singletons in the export. 
Maybe this works, but it needs to be properly tested.
The other PR's seem cool though. Thanks a lot for all this cleanup😍!

juliohq:
I think we can hold off on it in this case. Do those errors happen even for plugin code that isn't executed directly in an exported game?

Jowan-Spooner:
I think so, unless the whole script goes unused (for example the DialogicUtil script is used in exported games while the plugin.gd script is only used in editor). It's usually parse errors (when the script is loaded), due to finding an "unknown" class name.

Jowan-Spooner:
Yeah sorry @juliohq, I can confirm that this breaks exports, so I would close this PR unless you object.

I'm curious though, when are you even getting these warnings? I don't think I'm getting them or I'm missing it somehow 🙈.

juliohq:
Alright, in this case this PR should be closed. As for the warnings, they suddenly stopped to show up for me. I'm closing it rn.

Jowan-Spooner:
If the warnings come up again, feel free to create an issue for it and we will see what can be done!

PR #2268, opened 2024-06-04
Add scroll container to Dialogic UI

### LINKED ISSUE #2267: Shrinking the editor breaks dialogic UI
## The problem

**Describe the bug**
Dialogic UI breaks when you shrink the .

**To Reproduce**
1. Resize the editor window to its minimum size

**Expected behavior**
There should be scroll bars horizontally and vertically.

**Screenshots**
image

 - OS: Arch Linux (with i3 window manager)
 - Godot Version: v4.2.2
 - Dialogic Version: v2.0-Alpha-14

## Solutions

**Workaround**

N/A

**Possible fixes**

Wrap Dialogic UI with a `ScrollContainer` node to add scroll bars.

juliohq:
This PR wraps the Dialogic UI in a `ScrollContainer` node in order so that the contents are restricted and do not overflow. Fixes #2267.

image

### DIFF
--- a/addons/dialogic/Editor/Common/reference_manager_window.gd
+++ b/addons/dialogic/Editor/Common/reference_manager_window.gd
@@ -5,7 +5,7 @@ extends Window
 ## Other scripts can call the add_ref_change() method to register changes directly
 ##   or use the helpers add_variable_ref_change() and add_portrait_ref_change()
 
-@onready var editors_manager := get_node("../Margin/EditorsManager")
+@onready var editors_manager := get_node("../Scroll/Margin/EditorsManager")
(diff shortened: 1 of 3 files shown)


Jowan-Spooner:
Hey @juliohq, thanks for the PR. From my testing all this does is prevent the dialogic UI from overlapping with godot which might be more easily achieved with the `clip contents` flag enabled on the root node of the dialogic main editor. I never got scroll-bars to appear, probably because the parent node is a Control, not a container. I'm gonna try to fix the issue in a more simple way in a separate PR, but wanna thank you anyways!

PR #2271, opened 2024-06-07
Fix bug where 2 copies of a character are created on state load

AmphiBlocks:
When you load from state, currently the portraits get loaded twice - once during style load, and a second time during portrait subsystem load.

The reason is this line here:
`dialogic.current_state_info.portraits = {}`

when a character is added, it's supposed to clear out old copies of itself, but since it's not in the current state info on the second pass, it doesn't see that the character is already there, and makes a second copy.

This gets awkward when you try to continue the story and oops - the character animates out but the second copy stays, so the character doesn't leave.

Bug fix is that, if we're going to be setting portraits to {}, first have all present characters instantly leave.

### DIFF
--- a/addons/dialogic/Modules/Character/subsystem_portraits.gd
+++ b/addons/dialogic/Modules/Character/subsystem_portraits.gd
@@ -30,6 +30,8 @@ func load_game_state(load_flag:=LoadFlags.FULL_LOAD) -> void:
 		dialogic.current_state_info["portraits"] = {}
 
 	var portraits_info: Dictionary = dialogic.current_state_info.portraits.duplicate()
+	if load_flag != LoadFlags.ONLY_DNODES:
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
Hey @AmphiBlocks,
sorry to have taken so long. Thank you very much for contributing this solution! I have been hesitant to merge this, as I'm touching this anyways in my [Portrait/PortraitPositions PR](https://github.com/dialogic-godot/dialogic/pull/2123)  and wanted to see if this is even happening anymore afterwards. 
I have now found that my PR did not solve it, but decided that it should rather be solved properly by not causing this double-loading in the first place. I will try to add that to the Portrait position PR and get it merged rather soon.
Thanks again,
Jowan

Jowan-Spooner:
I'm gonna close this, as I think the issue has been fixed on main now. Thanks again for your work!

PR #2288, opened 2024-06-20
Fix "it's own" mistakes

juliohq:
This PR fixes all "it's own" orthographic mistakes.


Jowan-Spooner:
Are you sure this is correct? I'm not a native english speaker, but a quick google search makes it seem as if "on its own" is correct?

Jowan-Spooner:
OMG I'm so dumb, excuse my comment ;)

Jowan-Spooner:
Thanks, looks good to me.

PR #2294, opened 2024-06-21
Set tooltip for buttons in event blocks

juliohq:
This PR fixes the `tooltip` field of `display_info` being ignored (and hence not shown) for event block buttons, like the external link button to the variables editor:

image


Jowan-Spooner:
LGTM, Thanks!

PR #2295, opened 2024-06-21
Set timeline resource tooltip to its path in side bar

juliohq:
Every resource in Dialogic has a tooltip that shows its path in the file system, but timelines are different: they just show their name in the side bar. This PR makes that behavior consistent by displaying the timeline resource path as tooltip.


Jowan-Spooner:
Makes sense, thank you very much for noticing and fixing this inconsistency!

PR #2309, opened 2024-06-24
Export follow_viewport for default layout base

juliohq:
This PR adds a `follow_viewport` property that is able to be changed through the layer settings under the styles tab. `follow_viewport` sets `follow_viewport_enabled` for `CanvasLayer` layout base scenes.


Jowan-Spooner:
This is a simple feature, so I don't really mind, but could you elaborate when this would be useful? Is it a common thing to need? I guess it could be used on text-bubble like layouts, but those use a different base layer anyways?

juliohq:
That will make it possible for the text box to be positioned on top of the character manually via code (without being anchored to a specific corner of the screen). Think of something similar to the textbubble style:

image

Jowan-Spooner:
Sure. I really don't want to bloat this scene because if you really want anything custom from it it's SO easy to make a custom version, but I don't mind adding this if you feel it's useful.

juliohq:
I believe that's very useful since you'll be able to configure that option through the style settings. I think that's also in line with the layer setting that exists in the scene script alredy.

PR #2311, opened 2024-06-25
Fix follow viewport export order

juliohq:
Fixes the order of the `follow_viewport` export added by #2309. I'm sorry for the silly mistake!


Jowan-Spooner:
No problem.

PR #2319, opened 2024-06-30
Fix autosaving in ON_TIMER mode

jrb0001:
- Make it possible to configure and programmatically set the autosave interval.
- Fix default save corruption during subsystem init.
- Pause the timer while Dialogic.paused.
  - Let me know if you prefer to have an option for it.


Jowan-Spooner:
Lgtm. Thanks.

Issue #1316, opened 2022-11-07
Add way to let subsystems save on Dialogic.get_full_state()


Jowan-Spooner:
Maybe with a signal or by calling a method like save_full_state() on all subsystems that have it.
This would allow the subsystems to easily save stuff to the full state before it is "exported" or saved by the save system.

thebardsrc:
and to copy my notes here, what would need to be figured out is a way to hold for all of the subsystems to be ready once the save signal is emitted.  something like this:

1) when DGH is initialized, say 4 subsystems register to the signal  that they want to be able to save
2) saving is called
3) signal is sent out for all the subsystems to report in, and the save routine pauses
4) once all 4 subsystems have reported in that theyve written to the full state, then the full state is ready and it can be saved 

while checking into the Signal documentation, was able to see that there is a `get_connections()` function, which returns an array. so that works for the getting the number of them count. now just need to figure out how to do the waiting for it to report back as ready part 

a similar kind of Signal->Wait system also had come to a couple other things people had requested, like being able to do custom text processing with  #1112. now that i know that there's a way to see what's listening, ill loook around probably can find some good description of how to completely implement it

Jowan-Spooner:
Wouldn't it be possible to just call the method on every subsystem similar to this:
```gdscript
for subsystem in get_subsystems():
    if subsystem.has_method('save_full_state'):
        current_state_info = subsystem.save_full_state()
```

thebardsrc:
that would be a way to do it for just the save function, but that wouldn't solve the other example of doing it in other places, like post-processing of the text to handle otehr custom BBcode and such, or for any future similar things we may want to add.  would also mean that if they want to use the single save for everything, and have any other game data in the same save it would also have to register as a subsystem too

but i think I did figure out more of the signal -> wait idea as alternative:
- queue handling signal to emit, that other functions 
- timer (for timing out and moving on just in case one of them takes too long)
- go/nogo signal for the listeniers to return when they're ready

then a couple functions to do the handling. would then go like this :

1. some function (save for example) calls the fucntion to start the system
2. function notes how many listeners there are with `get_connected()`, and signal is emitted with the name of what's being called  to start
3. each of the functions listening for that signal receives it, does their thing, and then sends back the go/nogo signal that they're done
4. once it receives the same mumber of go/nogo back as listeners, then it's ready to finish out its own work 
[...]

PR #2320, opened 2024-06-30
Add option to include simple and/or full event history in savegame

jrb0001:
Simple and full event history can be toggled separately. If one option is enabled:
- Include that history in Dialogic.get_full_state() / Save.save().
- On Dialogic.load_full_state() / Save.load() replace that history with the history from the savegame.
  - If the savegame doesn't contain that history, clear it.
  - Events appended to the full event history between the clear_game_state() and the load_game_state() are dropped intentionally.
- On Dialogic.clear(FULL_CLEAR), clear that history.

The changes to the text event are needed to avoid duplicating the last entry in the simple history after a save+load. This also makes the text show up immediately after loading instead of the normal typing style.

Feel free to suggest a better option name or tooltip.

### DIFF
--- a/addons/dialogic/Core/DialogicGameHandler.gd
+++ b/addons/dialogic/Core/DialogicGameHandler.gd
@@ -321,6 +321,9 @@ func get_full_state() -> Dictionary:
 		current_state_info['current_event_idx'] = -1
 		current_state_info['current_timeline'] = null
 
+	for subsystem in get_children():
+		(subsystem as DialogicSubsystem).save_game_state()
+
 	return current_state_info.duplicate(true)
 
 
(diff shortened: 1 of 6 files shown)


Jowan-Spooner:
This looks good to me on the surface, will try to take a closer look the coming days.

Jowan-Spooner:
I think this makes sense. I made some changes to follow the gdscript style guide. 
I'm not entirely sure whether having these histories get erased on clear() is a good thing, but am willing to merge this and potentially adjust the behaviour in future PR's. 

This incidentally fixes #1316

Jowan-Spooner:
Thanks and congrats on your first PR!

jrb0001:
Thanks for the cleanup!

I thought about adding another option for the clear behavior but decided against it in the end. Enabling the save/load behavior more or less makes it part of the game state, like variables. So I wanted to mirror their behavior for consistency.

I can only see the use case where a game has many relatively short and repeatable timelines and uses dialogic variables as temporaries while keeping the real state outside dialogic. And unless such a game supports saving inside such dialogue, it wouldn't use the dialogic save system anyway so it has no reason to enable the option.

I am not against adding a separate clear option, but I also didn't want to clutter the settings tab with something that might never be used (independently).

PR #2337, opened 2024-07-10
fix default portrait animations

endaye:
`Fade Up In`/`Fade In Out`/`Fade Down Out` is outdated. Change them to the new ones.


DankDX:
this fix is very sigma

PixelMariachi:
I was wondering why did I get those errors all the time...

Jowan-Spooner:
Hey @endaye thanks for the PR! I've looked into it and this is not quite the correct fix. 
It would be correct to change the defaults to "Fade In Up", "Fade Out Down" and "Fade Cross".

However the main issue is that "Fade Up In" was actually also intended to work (mainly to allow for some flexibility when specifying animations). This seems to have been a mistake on my end and is relatively easily fixable.

PR #2343, opened 2024-07-12
Fix new text event getting stuck if Dialogic.start() is called during another text event

jrb0001:
The text event stores its state in `dialogic.current_state_info['text_sub_idx'] `which is needed for save+load to work correctly. Unfortunately it stays across a `Dialogic.clear()` (flags doesn't matter) so the first text event of a new timeline tries to resume in the middle of the event which would only work correctly directly after a `Dialogic.load_full_state()`.

1. Start the Visual Novel from dialogic-test-project.
2. Click on 'New Game'.
3. Textbox appears correctly.
4. Click the menu button and then the door button.
5. Click on 'New Game'.
6. Textbox is missing.

`clear_state()` is called on every `Dialogic.clear()` and also before the next event is executed so events can centralize their cleanup code. In case of a `Dialogic.load_full_state()`, it is called right at the start so when it starts executing the event, there is no previous event to clear and the next event sees the loaded state.

The text event uses this to erase `dialogic.current_state_info['text_sub_idx']` and to disconnect signals.

This PR also fixes the issue where loading into the middle of a multi section text event caused it not to update the textbox when advancing to the remaining sections.

### DIFF
--- a/addons/dialogic/Core/DialogicGameHandler.gd
+++ b/addons/dialogic/Core/DialogicGameHandler.gd
@@ -259,8 +259,7 @@ func handle_event(event_index:int) -> void:
 	if not current_timeline:
 		return
 
-	if has_meta('previous_event') and get_meta('previous_event') is DialogicEvent and (get_meta('previous_event') as DialogicEvent).event_finished.is_connected(handle_next_event):
-		(get_meta('previous_event') as DialogicEvent).event_finished.disconnect(handle_next_event)
+	_cleanup_previous_event()
 
 	if paused:
 		await dialogic_resumed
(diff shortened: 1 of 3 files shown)


Jowan-Spooner:
This is interesting. Thanks for the report and the PR! I will try to test this this weekend, the code looks good though. I think _clear_state() should start with an underscore. We've not been super thorough with that yet, but that's typical for methods that should be overwritten.

jrb0001:
Calling an underscore method from outside looks weird to me. But it is more consistent for overriding, that's true.

Another option would be to have both, but that looks weird as well:
```gdscript
func clear_state() -> void:
	_clear_state()

func _clear_state() -> void:
	pass
```

Jowan-Spooner:
I get that it looks weird, but the idea for overwritten methods is kindof that they are unaware that these things are "called" by someone. Like with _ready() and _process() the script that implements them just assumes some higher power will know when to activate them 😆.

I mainly care about consistency and tbh we have not been to consistent with this rule even though it is part of the style-guide, which is why I brought it up.

jrb0001:
@Jowan-Spooner The method name is changed. Is there anything else I should change?

Jowan-Spooner:
Hey @jrb0001, this PR is good, I just didn't find a lot of time for dialogic these past weeks.
Thanks again for this!

Issue #1894, opened 2023-11-11
Editor: Better Resource List + Sidebar


Jowan-Spooner:
The sidebar is currently sorted by alphabetical order. This is somewhat okay, but often doesn't actually show what you are working on. 
Characters and Timelines are also mixed which can be annoying. 

Better options for sorting should be considered. Possibilities:
- sort by last used (latest used always at the top)
- sort by alphabet but split characters and timelines
- allow sorting by folder (taking into account the file system). This would allow more complex sorting.

It should also be allowed to hide the sidebar for users who prefer to use quick-open (SHIFT+ALT+O) or the file system.

firedfromlife:
I would like to work on this feature.

firedfromlife:
@Jowan-Spooner Any input on the progress so far? Still need to implement the highlight on search before a PR is ready.

- Resources are now split into characters and timelines, then sorted alphabetically.
- I have added a collapse button beside the CurrentResource.
    - ![imgA](https://share.c02.co/data/images/CXB6-Godot_v4.3-beta3_win64_woYdDuEp0p.png)
- This image shows the UI while the sidebar is collapsed.
    - ![imgB](http://share.c02.co/data/images/YTX1-Godot_v4.3-beta3_win64_eTqE3zOVwR.png)

Jowan-Spooner:
@firedfromlife Hey, that looks pretty cool. Did you convert the item list into a tree? Or how are you making those sections?

If you open a PR, I could test it and give more detailed feedback. But I like the look so far!

firedfromlife:
It is currently in a very broken state, by that I mean, it seems to work completely fine, but floods the console with errors when an item is clicked in the tree. Loads the resource up just fine though.
I will get that resolved then post a pr.

Issue #2033, opened 2024-01-19
Add Groups for Timelines & Characters


AutumnRivers:
**Is your feature request related to a problem? Please describe.**
I'm currently working on a visual novel remake in Godot, using Dialogic. The main issue here is that the visual novel I am remaking is *massive*. Multiple books, multiple chapters, a whole slew of characters. The current issue I have right now is that the editor shows all these characters and timelines together, with no option to organize them.

**Describe the solution you'd like**
The ability to group together timelines and characters together in the main editor. This doesn't really have to change anything with how files are handled - this can be nothing more than a cosmetic change. Similar to how variables can be grouped.

There would be a separate icon next to "Create Timeline" and "Create Character" that would allow the developer to create a group. They could rename this group to whatever, and put either timelines or characters in it. Both, if they so wished.

**Describe alternatives you've considered**
Right now, I organize everything I need by starting each timeline/character's name with the category they're in. ex. Book 1 Chapter 1's timeline would be named `Book1_Ch1`, and Character A would be named `char_CharacterA`. This still leads to a fairly messy look, but it's better than nothing.

**Additional context**
I'm currently using Dialogic 2. As I stated above, this can be just a purely cosmetic change, and still require the usual unique identifies across all groups.

CakeVR:
Thanks for your issue.

Just a small note, variable groups are implemented using `Dictionary`s and are more than just cosmetics.

Jowan-Spooner:
My mistake, I used the wrong number. 
_There is only two proplems in computer science: Off by one errors._
Or something like that.

Jowan-Spooner:
The sidebar has gotten some really nice updates and can now be set to auto-group the characters and timelines by the folder they are in. I feel like this should already improve the issues described here a lot.

PR #2350, opened 2024-07-24
Sidebar upgrade #1894

firedfromlife:
**PR for:** "Editor: Better Resource List + Sidebar" #1894

- Added collapse functionality to sidebar.
- Added select after search to sidebar search.
- Updated sidebar sorting.
    - Now breaks resources into groups by "Character" and "Timeline" 

**Notes:**
- Due to odd behaviors and known issues of the godot "Tree" node, resources must be double clicked to open.
    - When using item_selected, a nullptr error occurs causing many problems including crashing of the editor.
- Tree node does not support Middle mouse events, so middle click to remove no longer works.


Jowan-Spooner:
I would love if we could get single-click to open working again. I will try to rebase this on the main branch first though.

firedfromlife:
@Jowan-Spooner 
Fixed. Single click now works.
One function wrap and a call_deferred was able to fix the issue. Found an issue in godot from 2018 that someone had that ended up being the answer.

firedfromlife:
Before merging this PR I would also like to add one more sort option, sorting by folders. Seems this would address a few issues including #2033 .

Jowan-Spooner:
I think it might also make sense to have an option to automatically collapse the Character and Timeline sections when switching to the character or timeline editors (so that you kinda only see the timelines when in timeline editor, but could still manually expand the character section). No need to add this though, it can also be done later.

addons/dialogic/Editor/Common/sidebar.gd
```diff
@@ -113,6 +122,13 @@ func clean_resource_list(resources_list: Array = []) -> PackedStringArray:
 
 
 func update_resource_list(resources_list: PackedStringArray = []) -> void:
+	print_rich(
+		(
+			"[b]Updating resource list[/b]:\n\t [i]Resource List: [/i]%s\n\t[i]Resource List D: [/i]%s"
+			% [resources_list, DialogicUtil.get_editor_setting("last_resources", [])]
```
firedfromlife, line 128: This should be removed when the merge happens. This was me troubleshooting an issue that was caused by having the editor scenes open.

Jowan-Spooner:
@firedfromlife do you consider this ready for merging or would you like to continue working on this? The only "bug" I noticed was that if you have two folders with the same name, they get merged (e.g. if I have `res://Chapter1/Home/tml.dtl` and `res://Chapter2/Home/some_tml.dtl` both get put into one `Home` section). It might be better to add the parent folder to the names (`Chapter1/Home` and `Chapter2/Home`), but that is kinda annoying to code, so I'd be okay with leaving it like it is for now.

firedfromlife:
I would like to fix that bug, sorry been busy with work recently. Plan to work on the issue this weekend.

firedfromlife:
@Jowan-Spooner Alright, so I added another sort option called "Path" This works closer to what you recommended. It currently just prints the whole relative path as the folder name, which may not be ideal.

I kept the "Folders" option working as before. I think this could be useful depending on how someone has their game organized.

Jowan-Spooner:
Hey @firedfromlife,
finally published alpha 15 so now back to looking at this. 
It all seems to work quite well. I'm fine with keeping the path option as is for now, although I agree it probably isn't ideal to just use the whole path. I think I would rename "Default" to "Type" though. 

I think we could potentially even implement using the folder colors set in the file system, not sure though (and not as part of this PR)

firedfromlife:
> "and not as part of this PR"

**Oops**... I was already working on that. Haha.

I also updated "Default to "Type" and "Path" now displays only the directory and the parent, if it has one.

Colors can't be added to the "Folder" mode without making that its own setting, which would be possible, and I am not apposed to. It merges contents of multiple directories, so it has no way to determine which color it would use.
The folder colors are simply stored in the project.godot file as follows: 

``` 
[file_customization]

folder_colors={
"res://assets/folder_1/timelines/": "yellow",
"res://assets/folder_b/timelines/": "orange"
}
```

If we really wanted to add colors to the folder option, we could add it like so:
```
[dialogic]
folder_colors={
"timelines":"red",
"assets":"teal"
}
```

I think this would be worth adding, and it would be (hopefully) simple enough to add to the existing right click menu for the sidebar.

Not pushing changes yet, still cleaning up some messy debug code.

image

Jowan-Spooner:
@firedfromlife Oh nice. I think for now showing the file system folder colors in the "Path" option would be enough. 

Should we also change the background color of the files then, like the file system does it? Maybe slighly less opaque?

Thanks for all the work on this.

firedfromlife:
@Jowan-Spooner 
Alright, I think that should be it for this PR. 

Let me know if it looks good to you 👍

Jowan-Spooner:
@firedfromlife Thank you so much for all this work. I will test it again over the next days.

Jowan-Spooner:
It looks good to me. I'd like to simplify some of the code, but I think I'll do that in a follow-up PR. These are working pretty well and while there are some further improvements that I feel could be great this is already so much better then before. Thank you very much!

PR #2363, opened 2024-08-09
Loading files at startup improvement

LBranco27:
Related to https://github.com/dialogic-godot/dialogic/issues/2362

### DIFF
--- a/addons/dialogic/Core/DialogicGameHandler.gd
+++ b/addons/dialogic/Core/DialogicGameHandler.gd
@@ -154,8 +154,6 @@ var Voice := preload("res://addons/dialogic/Modules/Voice/subsystem_voice.gd").n
 
 ## Autoloads are added first, so this happens REALLY early on game startup.
 func _ready() -> void:
-	DialogicResourceUtil.update()
-
 	_collect_subsystems()
 
 	clear()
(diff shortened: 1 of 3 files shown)


LBranco27:
Maybe the `DialogicResourceUtil.update()` can be in the _entre_tree() function.

exelia-antonov:
this looks solid to me but I'm not going to merge it since im a bit out of it on the current code. how much speedup have you been seeing with this by chance?

LBranco27:
> this looks solid to me but I'm not going to merge it right now myself since im a bit out of it on the current code. how much speedup have you been seeing with this by chance?

On a modern PC, the improvement may not be very noticeable due to the high read speed of storage devices. However, on an Android device (SM-M146B), it reduced the startup time of my project by about 20 seconds (from 25 to 5, as everything else doesn't demand extensive file reading). In any case, the larger the project, the longer the startup time would be.

Jowan-Spooner:
@LBranco27 Thanks for the PR!
This looks quite good and re-listing all the files is truly unnecessary here, but instead of the cashe from the project settings, I think it should use `DialogicResourceUtil.get_character_directory().values()`. I don't think it makes a big difference in practice, but it seems cleaner to me then getting it manually there.

LBranco27:
> @LBranco27 Thanks for the PR! This looks quite good and re-listing all the files is truly unnecessary here, but instead of the cashe from the project settings, I think it should use `DialogicResourceUtil.get_character_directory().values()`. I don't think it makes a big difference in practice, but it seems cleaner to me then getting it manually there.

I agree with you. It's cleaner and makes more sense since there's an existing function to do what I was doing "manually." I just made a new commit with that change.

PR #2380, opened 2024-08-30
Fix conditions contained in {brackets}

### LINKED ISSUE #2378: Inline condition statement cannot compare the value of variables
## The problem

**Describe the bug**
The inline conditional [if] statement added in #2142 cannot compare the value of a variable to any other value. Any attempt to use a comparison operator like `==, or, not, !` on a variable fails.

For my example I have created a boolean variable in Dialogic called 'myVar' and set its default value to true. In Dialogic, if I simply write an expression like `{{myVar} == {myVar}} ` or  `{{myVar} == true}` it correctly evaluates to true. However, if I write those same statements into an inline conditional [if] statement, it fails to evaluate, and the [if] statement defaults to its false condition. Dialogic throws the following error or something similar:

```
[Dialogic] Expression "{true or true}" failed to parse.
Expected ':'
```

Notably this **only** happens when attempting to compare the value of a variable. Comparing non-variable values seems to work just fine.  E.g `{1 == 1}`.

**To Reproduce**
- Create a boolean variable in dialogic and set its default value to true.
- Attempt to compare it to any value using operators like `==, or, not, !` inside of an inline conditional [if] statement

I have put together a small battery of tests using the `myVar` variable mentioned above. All of these lines are expected to evaluate to true. However, all of the [if] statements that attempt to call an operator on the `myVar` variable fail to parse. You should be able to paste the following into a Dialogic text event to reproduce this error.

```
{{myVar} == true} 
{1 == 1}
[if {1 == 1} Statement is true./broken]
[if {1 == 1 or 1 == 1} Statement is true./broken]
[if {myVar}  Statement is true./broken]
[if {{myVar} or {myVar}}  Statement is true./broken]
[if {{myVar} and {myVar}}  Statement is true./broken]
[if {{myVar} or 1 == 1}  Statement is true./broken]
[if {{myVar} and 1 == 1}  Statement is true./broken]
[if {{myVar} == true}  Statement is true./broken]
[if {{myVar} == {myVar}}  Statement is true./broken]
```

**Expected behavior**
Comparing the value of variables inside an [if] statement should behave the same way as it does in a regular text event expression or in the conditional expression.

 - OS: Windows
 - Godot Version: 4.2.2
 - Dialogic Version: 2.0 Alpha 14, downloaded most recent pre-release files as of August 25th, 2024

Jowan-Spooner:
Conditions that where sorrounded  by brackets (like is required in the conditional modifier) would sometimes not be correctly resolved. This should fix this.

- fixes #2378

### DIFF
--- a/addons/dialogic/Modules/Core/subsystem_expression.gd
+++ b/addons/dialogic/Modules/Core/subsystem_expression.gd
@@ -20,6 +20,9 @@ func execute_string(string:String, default: Variant = null, no_warning := false)
 	for res in regex.search_all(string):
 		var value: Variant = dialogic.VAR.get_variable(res.get_string())
 		string = string.replace(res.get_string(), var_to_str(value))
+	
+	if string.begins_with("{") and string.ends_with('}') and string.count("{") == 1:
+		string = string.trim_prefix("{").trim_suffix("}")
 
 	var expr := Expression.new()
 


Jowan-Spooner:
Seems to work quite well. Will just merge 🙏

PR #2393, opened 2024-09-12
CharacterEvent: Fix position/size not working correctly when joining

### LINKED ISSUE #2389: New 'pos' and 'size' features don't appear to work as intended
## The problem

**Describe the bug**
Using 'pos' and 'size' on the same line as a character update causes character position to move in unintended ways.

**To Reproduce**
Using 'pos' and 'size' like this will cause the character portrait to size properly:
```
join Character pos=x0.7 y1.0
update Character size=y500px
```
This will cause the portrait to stay in the indicated position and change size, as intended.
image
image

Using 'pos' and 'size' on the same line causes the portrait to jump up the screen:
```
join Character pos=x0.7 y1.0 size=y500px
update Character (Expression2)
```
image

...and then when `update` is called, the portrait snaps to the proper location.
image

**Expected behavior**
The portrait should remain in the correct screen position when a new scale is applied.

**Screenshots**
See above

 - OS: Windows 10
 - Godot Version: v4.3.stable.mono.official [77dcf97d8]
 - Dialogic Version: 2.0-Alpha-15

## Solutions

**Workaround**

Updating portrait scale and position separately allows the correct behavior, but with an uncomfortable gap of time, which causes the portrait to suddenly "snap" to the desired size after entering the scene.

Jowan-Spooner:
- should fix #2389

### DIFF
--- a/addons/dialogic/Modules/Character/subsystem_containers.gd
+++ b/addons/dialogic/Modules/Character/subsystem_containers.gd
@@ -182,7 +182,7 @@ func resize_container(container: DialogicNode_PortraitContainer, rect_size: Vari
 			tween.finished.connect(save_position_container.bind(container))
 	else:
 		container.position = container.position + relative_position_change
-		container.set_deferred("size", final_rect_resize)
+		container.size = final_rect_resize
 		save_position_container(container)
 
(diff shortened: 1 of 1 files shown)


McStebb:
Tested and working!  Good work.

PR #2405, opened 2024-09-17
Tag dtl and dch files with their resource class

### LINKED ISSUE #2226: Cannot quick load DialogicCharacter .dch custom resource file
## The problem

**Describe the bug**
I reference a DialogicCharacter from my own script:
@export var dialogic_character: DialogicCharacter

but using Quick Load in the inspector doesn't list it.

image

Load, however, works (but I must navigate to the file and select it).

image

Interestingly, *.dch seems automatically added to the list of recognized files in Godot (maybe it scans all files, or was it manually registered from the plugin?)

image

Anyway, using either All Recognized or .dch, I could select and assign the file properly.

image

I wonder if it's a limitation from Godot. According to this closed issue: https://github.com/godotengine/godot/issues/67119

custom resources should now support Quick Load (since 4.0), however it says nothing of custom extensions like .dch. Is it the source of the issue?

I searched through the doc (https://docs.dialogic.pro/), but I found no page showing how to use `register_character` to pass a DialogicCharacter so the bubble shows this character's portrait. Only enough, only the sample bubble (text_bubble_base.tscn, shown when calling `Dialogic.start('timeline')` alone) mentions this method:

```gd
layout.register_character(load("res://path/to/my/character.dch"), $BubbleMarker)
```

The sample code gives a hardcoded path to simplify, so there is no such issue, but in production I would export such a variable. But then I cannot Quick Load the exported DialogicCharacter var. I also checked in code how the Dialogic Panel worked, but it does some custom scan of all .dch files (we could do the same to implement a custom Quick Load maybe?).

Otherwise I could export a String var to use as path to resource; but then again I wouldn't benefit from resource type filtering and I would have to navigate and select the file anyway.

In practice it's OK because all my characters are put in the same folder and I don't have many of them, so it would really matter for a bigger project.

**To Reproduce**
1. Create a script with `@export var dialogic_character: DialogicCharacter`
2. Create some character in Dialogic panel, saved as .dch file
3. In Inspector, try to Quick Load dialogic_character with that file
4. See that file doesn't show

**Expected behavior**
Character resource files should be shown

 - OS: Linux Ubuntu 22.04
 - Godot Version: v4.2.1.stable.official [b09f793f5]
 - Dialogic Version: 2.0-Alpha-13 (Godot 4.2+) (installed manually)

## Solutions

**Workaround**

Do a classic Load and select file in popup

**Possible fixes**

It may only be possible on Godot side, waiting for diagnostic from dev

iscke:
(this makes them visible to quick load, should fix #2226)


Jowan-Spooner:
Hey @iscke, thanks you so much for this, this sounds awesome!

Jowan-Spooner:
I like this, but I'd need someone to test if this actually works, e.g @zaknafean.

zaknafean:
Testing out PR #2405 is good! The UI for Timeline COULD get a bit wild, but its a great and simple QoL. 

https://github.com/user-attachments/assets/d358a7c3-ed85-426a-875d-8cee50b76ada

Cons:
UI gets a bit annoying to deal with when you have alot of timelines. I think this is more a godot issue rather than an implementation. I do not believe that should prevent merging at this time.

https://github.com/user-attachments/assets/eb93b65c-f3fa-4ca5-90cd-decbf4bbb965

zaknafean (approved):
Fully tested, and a good QoL for everyone involved. Should merge!

PR #2418, opened 2024-09-28
Make variable editor right click menu hidden by default

salianifo:
The RightClick menu showed up in the top left of the editor each time the editor was started because `visible` was set to `true`. This fixes that so the menu doesn't show up by default.


Jowan-Spooner:
Thank you very much!

PR #2423, opened 2024-10-02
Add timeline_active bool to DialogicGameHandler.gd

### LINKED ISSUE #2422: Boolean to check if a timeline is active
For a lot of games I have to know if there's an active dialog, so I can disable other interactions and character movement, a boolean that checks if there's an active timeline would be really useful.

A bool in the DialogicGameHandler.gd file that we set to true in the start_timeline() function, then set it to false in end_timeline().

First I added this variable to my own Global script, then gave it a new value at the beginning and end of every Dialogic timeline using Dialogic's function calls, which was something I had to do in every timeline, so I gave up and modified the DialogicGameHandler script.

polgar1:
Added a variable called timeline_active to DialogicGameHandler.gd, which is set to 'true' if there's an active timeline.
Resolves #2422

### DIFF
--- a/addons/dialogic/Core/DialogicGameHandler.gd
+++ b/addons/dialogic/Core/DialogicGameHandler.gd
@@ -45,6 +45,8 @@ var current_state := States.IDLE:
 		current_state = new_state
 		state_changed.emit(new_state)
 
+##'true' when a timeline is active
+var timeline_active := false
 ## Emitted when [member current_state] change.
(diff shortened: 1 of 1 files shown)


CakeVR:
Thanks for your Pull Request.

I don't this is necessary, you can check `Dialogic.current_timeline` for being `null`. This will tell you if a timeline has been set or not (`null`).

Issue #2420, opened 2024-10-01
Unable to select Typing Sound Audio Bus via Layer Settings


anis-mahsoume:
I am not able to change the bus for the character typing sounds.

I have searched on the previous commits and found this one:
https://github.com/dialogic-godot/dialogic/pull/344/commits/e83d7e6a431d74e843c7c3e8458de29df5c20a99

Seems like there use to be a field to set the bus for typing sounds. is there an equivalent now ? I cannot find it in the documentation or in the issues neither.

CakeVR:
Right now, the Layers having the Node with the Typing Sounds cannot be changed from the Layer Configuration. If you want, you are free to add such possibility to all built-in Dialogic Styles.

However, you can turn the Layer with the Typing Sounds Node into a *Custom Layer* and then simply change the Audio Bus on the Node. The Node is named `DialogicNode_TypeSounds`.

anis-mahsoume:
Thanks @CakeVR  for your response. I have made a pull request to add the option into character settings:
#2425

anis-mahsoume:
Hi @CakeVR, concerning my PR, since it's my first participation I am not aware of the process. Can I get it merged ?

PR #2425, opened 2024-10-03
Add: Typing Sound Audio Bus in Character Settings

anis-mahsoume:
Add a new field "Audio Bus" into the character settings typing sounds section:
image
The Audio Bus is an OptionButton field which values are populated from AudioServer buses.
I decided to make the change after opening the issue:
#2420

### DIFF
--- a/addons/dialogic/Modules/Text/character_settings/character_moods_settings.gd
+++ b/addons/dialogic/Modules/Text/character_settings/character_moods_settings.gd
@@ -234,6 +234,6 @@ func preview() -> void:
 
 func _load_audio_bus_options() -> void:
 	%AudioBus.clear()
-	for i in AudioServer.bus_count:
+	for i: int in AudioServer.bus_count:
 		var bus = AudioServer.get_bus_name(i)
 		%AudioBus.add_item(bus)
--- a/addons/dialogic/Modules/Text/character_settings/character_moods_settings.gd
+++ b/addons/dialogic/Modules/Text/character_settings/character_moods_settings.gd
@@ -60,8 +60,8 @@ func _ready() -> void:
 	%Duplicate.icon = get_theme_icon("Duplicate", "EditorIcons")
 	%Play.icon = get_theme_icon("Play", "EditorIcons")
 	%Default.icon = get_theme_icon("NonFavorite", "EditorIcons")
-
 	%NameWarning.texture = get_theme_icon("StatusWarning", "EditorIcons")
+	_load_audio_bus_options()
 
 
 func update_mood_list(selected_name := "") -> void:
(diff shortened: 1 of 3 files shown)


addons/dialogic/Modules/Text/character_settings/character_moods_settings.gd
```diff
@@ -228,3 +230,10 @@ func preview() -> void:
 		await preview_timer.timeout
 
 	preview_timer.queue_free()
+
+
+func _load_audio_bus_options() -> void:
+	%AudioBus.clear()
+	for i in AudioServer.bus_count:
```
CakeVR, line 237:
```suggestion
	for i: int in AudioServer.bus_count:
```

addons/dialogic/Modules/Text/character_settings/character_moods_settings.tscn
```diff
@@ -163,6 +163,18 @@ size_flags_horizontal = 3
...
 file_mode = 3
 
+[node name="AudioBusLabel" type="Label" parent="VBox/Settings/Grid"]
+layout_mode = 2
+text = "Audio Bus:"
+
+[node name="AudioBus" type="OptionButton" parent="VBox/Settings/Grid"]
+unique_name_in_owner = true
+layout_mode = 2
+tooltip_text = "Audio bus in which the type sounds will be sent"
```
CakeVR, line 173:
```suggestion
tooltip_text = "Audio bus on which the typing sounds will be played."
```

addons/dialogic/Modules/Text/node_type_sound.gd
```diff
@@ -41,6 +41,9 @@ func _ready() -> void:
 
 
 func _on_started_revealing_text() -> void:
+	var default_bus_index: int = AudioServer.get_bus_index(bus)
+	var bus_index: int = current_overwrite_data.get('audio_bus', default_bus_index)
+	bus = AudioServer.get_bus_name(bus_index)
 	if !enabled or (get_parent() is DialogicNode_DialogText and !get_parent().enabled):
```
CakeVR, line 47:
```suggestion
	bus = AudioServer.get_bus_name(bus_index)
	
	if !enabled or (get_parent() is DialogicNode_DialogText and !get_parent().enabled):
```

CakeVR (changes requested):
Just some minor improvements, you can simply apply all of them.

Thanks for your Pull Request.

anis-mahsoume:
> Just some minor improvements, you can simply apply all of them.
> 
> Thanks for your Pull Request.

Thanks @CakeVR for your review: I have applied the changes you suggested.

CakeVR:
The problem I have with this solution is that you use `_on_started_revealing_text` to set the Audio Bus which gets triggered every single text message.

I don't think this is a good solution.

Jowan-Spooner:
I don't think this should be a setting in the character editor, rather a global setting or a style setting. I cannot see a use-case where you'd need to change this per type sound mood.

CakeVR:
Hello, is there any interest in continuing on this PR?
As I agree with Jowan, there would have to be Global Setting in the Settings tab instead.

anis-mahsoume:
Hi,
This issue has been fixed in
https://github.com/dialogic-godot/dialogic/pull/2456
So closing this PR

PR #2426, opened 2024-10-04
Fix various music player issues

salianifo:
In reviewing the `subsystem_audio.gd` file for implementing music channels, I found the following bugs:
- Previous music track gets duplicated and only the copy gets faded out, leaving behind the original track
- `is_music_playing_resource` erroneously checks the `base_music_player` instead of the `current_music_player` and thus always returns `false`
- `current_music_player` gets `stop()` and `queue_free()` called on it when there is no path, not allowing for it to be faded out.
- Removed unnecessary `pass` at the end of `_on_dialogic_timeline_ended`

Thought I would just fix it instead of make an issue :smiley:

### DIFF
--- a/addons/dialogic/Modules/Audio/subsystem_audio.gd
+++ b/addons/dialogic/Modules/Audio/subsystem_audio.gd
@@ -79,7 +79,6 @@ func resume() -> void:
 func _on_dialogic_timeline_ended() -> void:
 	if not dialogic.Styles.get_layout_node():
 		clear_game_state()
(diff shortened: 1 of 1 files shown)


salianifo:
This PR is superseded by the multiple music channels #2427, so you don't need to merge this one if you merge that one.

zaknafean (commented):
As per the author, we will be ignoring this PR and looking at the other PR instead. We will close this one out.

PR #2427, opened 2024-10-06
Implement multiple music channels

salianifo:
Supersedes #2426

This PR adds the ability to specify what music channel to play on in the Music event. Adds a settings page for Audio with only one setting (max music channels). I made sure to code in backwards compatibility (will play on the first channel).

I would appreciate any feedback, as well as some testing to make sure everything works correctly.


zaknafean (changes requested):
Looks good and I love the functionality.

The timelines do break if the channel setting is adjusted and called for. I'd like to see some error handling around 287 in the linked image to avoid that if at all possible. Otherwise we're good to go

image

salianifo:
Added a check to the `is_music_player_resource` function that you requested, plus some error info in the `update_music` function, and a note to the Audio settings that lowering the max channels may invalidate existing music events.

zaknafean (approved):
Alright, I'm satisfied with this! It's as protected from bad decisions as it can be! Approved and merging

PR #2428, opened 2024-10-06
Choices: add array boundary checking

endaye:
* to fix this bug below
image


Jowan-Spooner:
This looks alright, but it fixes a symptom, I'd be really curious why this happens.

endaye:
I am not sure. Maybe it was caused by what I called `Dialogic.clear()` when it's running.

zaknafean:
Does that cause it to happen consistently? Can you share the code?

zaknafean (changes requested):
While I generally like bounds checking, this feels like a bandage to a problem that maybe shouldn't exist. We really need an example of when this happens here specifically, and why. Do you have a small project we could review?

PR #2432, opened 2024-10-14
handle some null/freed item cases to avoid bugs

### LINKED ISSUE #2381: Timelines will save over each other
**Describe the bug**
If you click between different timelines fast then one timeline will show up as that timeline for example if I have timeline A and B; and I switch between them fast then timeline A will become timeline B, if you reload the project it's not an issue but if you save then it'll save timeline A as B 

**To Reproduce**
Just spam click between them this is the fastest way I am able to recreate it (not that I spam between timelines)

**Expected behavior**
Timeline A to show Timeline A and Timeline B to show B 

 - OS: Linux
 - Godot Version: 4,3
 - Dialogic Version: 2.0 Alpha 14

## Solutions

just reload don't save

Invertex:
May fix #2381  but don't know for sure yet.

Regardless, this gets rid of some potential error spam that isn't useful to the end-user.

Sometimes the `opener_events_stack` items can be in the process of being freed, or have become Null by the time the code got there and causes an error trying to feed them into the `create_and_branch_event()` function.
Same with the `_batches` content.

### DIFF
--- a/addons/dialogic/Editor/TimelineEditor/VisualEditor/timeline_editor_visual.gd
+++ b/addons/dialogic/Editor/TimelineEditor/VisualEditor/timeline_editor_visual.gd
@@ -112,9 +112,11 @@ func batch_events(array: Array, size: int, batch_number: int) -> Array:
 var opener_events_stack := []
 
 func load_batch(data:Array) -> void:
-	var current_batch: Array = _batches.pop_front()
+	# Don't try to cast it to Array immedietly, as the item may have become null and will throw a useless error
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
Looks good generally, some testing would be nice.

zaknafean (approved):
Overall this does no harm and likely does fix some issues some people see, though I can't recreate the issues this was supposed to address consistently. My concern is this may be more a bandage to a core problem. But it isn't enough of a problem in my eyes to not merge up.

PR #2436, opened 2024-10-15
Fix test timeline scene error with Dictionary signal event

salianifo:
Fixes an error when using the Play Timeline button when you have a signal event that uses a `Dictionary` instead of a `String`.


zaknafean (approved):
While the error wasn't crashing anything, this does fix a very real error that was being thrown. Merging it up.

PR #2439, opened 2024-10-15
Add skippable option to Wait event

salianifo:
Closes: #1029 

Added a skippable option to the wait event. Default value is false. I don't think that the `is_instance_valid` is needed for the `if _is_instance_valid(_tween):` but I put it there just in case.


zaknafean (approved):
Good extension of the wait event without messing with the point of it.

Issue #1072, opened 2022-08-02
Implement an audio preview


Jowan-Spooner:
Implement an audio preview like in 1.x

PR #2440, opened 2024-10-15
Implement audio previews

salianifo:
Closes: #1072 

Behaves essentially the same as Dialogic 1. An `AudioStreamPlayer` gets added as a child to the `editor_node` in the `_enter_visual_editor` function.


Jowan-Spooner:
I'm wondering if we could just add an editor-field that does this and can then be added in the build_event_editor() method.

salianifo:
Closing in favor of #2448.

Issue #2372, opened 2024-08-16
Extreme zoom when trying to create new portrait


Shadoudoune:
I'm using Godot version 4.3 and Dialogic 2.0 Alpha 14 and when I try to add a new portrait to one of my characters, it zooms in my editor but i can't unzoom. Sometimes, I manage to escape the glitch by resizing the wiindow like a maniac, but when I go back to the character menu, it gives me the glitch again.
This may already be fixed but I don't know why it happens
![Capture d'écran 2024-08-16 104115](https://github.com/user-attachments/assets/f6988ce5-7e21-438e-a074-239131a59665)

Shadoudoune:
Apparently it happens because of this : character_editor_portrait_tree.gd:86 apparently this script calls a method that can't be called

PR #2442, opened 2024-10-16
Fix portrait tree drag drop bugs

salianifo:
Maybe Fixes: #2372 

Added check to make sure something is selected when trying to create drag preview in the Character editor's Portrait tree. Also renamed some function arguments to avoid shadowing builtin `position` variable.


salianifo:
I'm actually wondering if instead of using `get_selected()` if we should check for an element under the position and get that. After testing this fix, if you have an item selected, but you drag in the empty space at the bottom, it can reorder that item, which is likely not intended, or expected by the user.

Jowan-Spooner:
Not sure if it fixes #2372, but the changes look good.

PR #2444, opened 2024-10-16
Remove stray print statment, fix extension creator mode

salianifo:
A stray print statement was leftover from a recent merge, so I removed it.
The Extension Creator's setup mode had two options reversed. Selecting `Event+Subsystem` would only create an event and `Subsystem only` would create both an event and subsystem.


zaknafean:
Thanks for the double check!

Issue #2443, opened 2024-10-16
Portraits don't respect the window stretch mode with percent sizing


salianifo:
## The problem

**Describe the bug**
When sizing character portraits using percentage based sizes, the window stretch mode is not considered. The portraits are scaled based on the window size only, the `content_scale_size` is not taken into account. This causes issues when trying to set a target render size that is different than the window size (for example having the resolution be 1080p for fullscreen, but having windowed mode be 720p).

**To Reproduce**
1. Under Project Settings -> Display -> Window set
 - Size -> Viewport Width: 1920
 - Size -> Viewport Height: 1080
 - Size -> Window Width Override: 1152
 - Size -> Window Height Override: 648
 - Stretch -> Mode: `canvas_items` or `viewport`
2. Join a character with position set to `pos=x0.25 y1 size=y1`
3. Change from windowed mode to fullscreen
4. Join a second character with position set to `pos=x0.75 y1 size=y1`
5. Observe different sizing on the second character based on the new window size

**Expected behavior**
Portraits should be sized differently depending on the Stretch -> Mode setting (and possibly Stretch -> Aspect setting)

**Screenshots**
Minimal reproducible project is better than screenshots ^_^
[dialogic-scaling-mrp.zip](https://github.com/user-attachments/files/17402076/dialogic-scaling-mrp.zip)

**System (please complete the following information):**
 - OS: Linux Mint 21.3, Windows 10
 - Godot Version: 4.3-stable
 - Dialogic Version: 2.0 Alpha 16-WIP

## Solutions

**Workaround**

There is no workaround as only the window size is considered when sizing portraits relative to the screen size.

**Possible fixes**

The window's `content_scale_size` should be used at [subsystem_containers.gd:260](https://github.com/dialogic-godot/dialogic/blob/880a3e297a795edf77074435ffd55635bc5cacd6/addons/dialogic/Modules/Character/subsystem_containers.gd#L260) instead of just the window size directly depending on the above mentioned settings. I would fix this myself, but I'm unsure of the correct way to deal with the various combinations of stretch mode and stretch aspect. Additionally, this may be a breaking change for some projects.

salianifo:
@zaknafean Ping!

PR #2445, opened 2024-10-17
Fix portrait scaling

salianifo:
Closes: #2443 

Based on conversation with @zaknafean, I checked how this was done in Dialogic 1, and it uses `get_viewport().get_visible_rect()`, which according to my testing does respect the window stretch mode in Project Settings.

More detailed testing may be required, I only tested that the size matches expectations based on the stretch mode settings of `disabled` and `canvas_items`. When set to `disabled` the size reflects the actual window size, but on `canvas_items` the viewport size is returned instead, regardless of the window size. If I have time for further testing I will post any results as well.


Jowan-Spooner:
This seems like a sensible change. Thank you very much for looking into this and making the PR 🎉!

PR #2448, opened 2024-10-18
Implement audio preview fields

salianifo:
Supersedes: #2440 Closes: #1072 

Changed the audio previews to be built-in fields.
Added background previews as well since they both needed to be added to `event.gd` and `event_block.gd`.

There is an issue with the background event (unrelated to this PR directly though):
- If using a custom scene with a string argument, `_arg_type` isn't saved, so reloading the timeline thinks that the argument is an image again, causing the preview to show up with no image. This should be addressed separately though as it is unrelated to this PR.

The background preview does cause a warning about fit expand modes not being supported by Flow Containers, but that doesn't effect the functionality at all. It is set this way as the image is given a predefined height for the preview, but can adapt to whatever width the image has.

Edit:
Removed background previews for now until we can decide the implementation details.


salianifo:
Removed background previews for now until we can decide the implementation details. They can be added in a new PR later.

Jowan-Spooner:
Looks good to me for now. Possible changes to the audio events might require more work, but this is a cool addition.

PR #2449, opened 2024-10-22
Add accessibility settings, implement image preview field

salianifo:
Adds a new accessibility settings page with the following settings:

- Image preview height [default: 50 px] (used for the image preview field)
- Event block bottom margin [default: 0 px] (used to add some extra space between event blocks to improve readability)
- Show event names in timeline [default: false] (adds a header label to the beginning of all events with the event name to help distinguish events better)

The image previews are automatically unloaded when the event block body is hidden, and not loaded at all if image preview height is set to 0. Image preview size and event block bottom margin are scaled according to the editor scale (`DialogicUtil.get_editor_scale()`). The settings page includes hints noting that some settings require reloading the visual timeline for changes to take effect.

### DIFF
--- a/addons/dialogic/Editor/Events/EventBlock/event_block.gd
+++ b/addons/dialogic/Editor/Events/EventBlock/event_block.gd
@@ -55,6 +55,8 @@ func _ready() -> void:
 func initialize_ui() -> void:
 	var _scale := DialogicUtil.get_editor_scale()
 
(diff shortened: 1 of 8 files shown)


Jowan-Spooner:
I made a bunch of changes:
- Renames the Accessibility Page to "Editor"
- Moves the Color Palette and Section Order settings to the new Editor page
- Stores all those editor settings as DialogicUtil Editor settings instead.
- Makes all the relevant editor settings reload the timeline editor (including color palette and the new settings).

PR #2450, opened 2024-10-23
Fix Publish Test Report failures

salianifo:
Tested these changes on my main branch and it seems to fix the publish test report Github Action failing.

Also updated to the latest versions of the three actions (checkout, upload-artifact, and dorny/test-reporter) to fix warnings about outdated node.js versions.

Adds the artifact name to the test report, and changes the path so it can find the files to generate the report.

Link to passing publish test report -> [salianifo/dialogic/runs/31923968779](https://github.com/salianifo/dialogic/runs/31923968779)

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -31,7 +31,7 @@ jobs:
 
     steps:
       - name: "Checkout your Repository"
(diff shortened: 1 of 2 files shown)


Jowan-Spooner:
Thanks!

PR #2454, opened 2024-10-24
Fix events teleporting around with fast mouse clicks

salianifo:
Discovered an issue where events would randomly teleport in the visual timeline when clicking too fast for the `_process` function to run. Additionally, sometimes the mouse position would be in the dead space between events causing `drag_to_position` not to get updated when `_process` did get called.

This fix ensures that `drag_to_position` has been updated at least once after starting a drag operation, otherwise cancels the drag on completion.

This was a pretty difficult bug to reproduce on my machine, but my tester has a very fast click and could consistently reproduce the issue, to the point where their timeline constantly was getting garbled.


Jowan-Spooner:
Thanks. Looks reasonable to me.

Issue #2455, opened 2024-10-25
Copying Multiple Items from one timeline to another sends the first item to the bottom of the ordering when pasted


adrientremblay:
## The problem

**Describe the bug**
Copying Multiple Items from one timeline to another sends the first item to the bottom of the ordering when pasted

**To Reproduce**
1. Create a new timeline with multiple distinct items
2. Select all items and copy them to the clipboard
3. Create another timeline and paste the items into the timeline
4. Observe the first item in the original sequence is at the bottom

**Expected behavior**
The pasted items to maintain their ordering

**Screenshots**
N/A

**System (please complete the following information):**
 - OS: Linux
 - Godot Version: 4.3 Stable
 - Dialogic Version: Dialogic 2.0 - Alpha 15

## Solutions

**Workaround**

N/A

**Possible fixes**

N/A

Jowan-Spooner:
I cannot recreate this on windows on the main branch. Could you share more about your setup, maybe a video, screenshots, etc.

adrientremblay:
Heres a video of me producing the bug: https://youtu.be/TUHs4zfXs4o

I am on Linux Mint. Maybe this is a linux only issue? Wild Guess: but maybe there are some particularities to the linux clipboard.

salianifo:
Confirmed. Happens on both Windows and Linux.

Jowan-Spooner:
That makes sense, I wasn't pasting into an empty timeline, my bad for not following the steps precisely 🙈.

adrientremblay:
Thanks guys!!!! @Jowan-Spooner @salianifo

Issue #2407, opened 2024-09-18
Box containers not copied when performing copy paste


AtlaStar:
1. Select an If branch in the visual editor
2. Copy it and paste it somewhere, cut and paste works properly
3. See the following

![image](https://github.com/user-attachments/assets/8291d34f-186b-4c22-9580-ce8d327f86cc)

**Expected behavior**
To copy the box container node at when copying

System:
 - OS: Windows 10
 - Godot Version: 4.3.stable
 - Dialogic Version: 2.0 alpha 15

Workaround:
![image](https://github.com/user-attachments/assets/1e1d93ba-83e2-4558-9084-00a1c5d7fe8d)
ctrl  click the box so it is also selected in the copy operation before copying. As you can see it is highlighted gray, showing it is selected, unlike it being purple like the other image (probably not intended behavior either)

Since the cut and paste operation works correctly, the logic error has to be something broken in the copy operation itself

AtlaStar:
I should note; it isn't so much that they aren't copied as that appears to be the intended behavior, but rather that the GUI doesn't populate new UI controls on copying which is expected behavior

PR #2459, opened 2024-10-29
Fix visual editor event copy/paste issues

salianifo:
Closes: #2455
Closes: #2407

When pasting into an empty timeline, the paste position would be set to -1, making the first item pop down to the end of the list.

### DIFF
--- a/addons/dialogic/Editor/TimelineEditor/VisualEditor/timeline_editor_visual.gd
+++ b/addons/dialogic/Editor/TimelineEditor/VisualEditor/timeline_editor_visual.gd
@@ -1108,8 +1108,8 @@ func _input(event:InputEvent) -> void:
 			var paste_position := 0
 			if selected_items:
 				paste_position = selected_items[-1].get_index()+1
-			elif %Timeline.get_child_count() > 0:
-				paste_position = %Timeline.get_child_count()-1
+			else:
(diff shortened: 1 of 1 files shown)


salianifo:
Since I'm looking at the copy/paste features of the visual timeline, I noticed there's a few other things that need to be fixed as well, such as nested events not being nested when pasted.

salianifo:
I noticed that when you paste into a timeline with events (but none selected), it pastes right before the last event instead of after it. This last change fixes this as well. I think the other copy/paste/move issues need discussion, and should be addressed in another PR.

Jowan-Spooner:
Why did you change it from using `from_text` to `_load_from_string`? It sounds correct to me, but it makes me wonder if we should also use `_store_as_string()` instead of `to_text()`, to avoid the translation ids getting lost. Was there something specific that mad you change it?

salianifo:
Yes, when you have custom defaults set, they get loaded in `_load_from_text ` which then calls `from_text`. Without this, if you copy an event that has and uses custom defaults, when pasting it will paste the original defaults instead. My tester set a default audio bus for Voice events, but when copying and pasting a brand new event it would paste it with Master instead of Voice for the audio bus.

Jowan-Spooner:
Looks good to me. Tysm!

PR #2463, opened 2024-10-31
Add Screen Shake feature

salianifo:
Adds a new screen shake feature, allowing you to set the amplitude/frequency for both horizontal and vertical screen shakes, independently. 

~~*Note: This feature requires adding the Screen Shake Layer to your styles, and it has NOT been added to the default styles at this time.*~~ The screen shake layer has been added to the default Visual Novel Style and Speaker Textbox Style, below the text boxes in priority, but above the background and portraits (where applicable)

Amplitude/frequency can both be faded over time, and they have a separate duration that does not include the fade time. This allows for a smooth transition into and out of shaking the screen, while still allowing the full strength of the shake to last for a particular duration.

There is an additional option to wait for completion if the duration is greater than 0, or if the amplitude is 0 and there is a fade time greater than 0.

If the duration is 0 and the amplitude is not 0, the screen will continue shaking until the next screen shake event. To end a screen shake, you can either set the amplitude to 0 in another screen shake event, or use the clear event (this will clear both directions simultaneously).

The edges of the screen are drawn with the color specified at `rendering/environment/defaults/default_clear_color` in the Project Settings.

Amplitude is in the range of 0 to 1, with a step value of 0.01. An amplitude of 1 will move the entire size of the screen at its peak. 

Frequency has a minimum of 0.1 (a frequency of 0 would cause it to no longer update), and has no upper bound. It is measured in hertz or cycles per second, so higher numbers means faster shaking. This means a value of 1 will move the screen from one side to the other and back to the center in 1 second.

Here is an example of shaking the screen in both directions at different rates, then using the clear event.

https://github.com/user-attachments/assets/03133ad7-87cb-46a9-b68b-2db23bd11ea2


UsernameIsInUse:
So there's a chance I've done something wrong, but pulling this request as is and making no changes results in this error:
`event_screen_shake.gd` line 31 `dialogic.ScreenShake.update_shake_x(amplitude, frequency, fade, duration)`
`Invalid access to property or key 'ScreenShake' on a base object of type 'Node (DialogicGameHandler)'.`
I added the layer to the style, just dropping the default screenshake onto the timeline results in this issue.

salianifo:
Yup, that is indeed a mistake on my part. I will update this when I get some time.

salianifo:
Rebased and fixed merge conflicts, no other changes.

salianifo:
Rebased on latest main, fixed differing values for the `duration` in the event, added the Screen Shake layer to the default styles (VN and Speaker Box, will shake portraits and background, but not textbox by default), and added UIDs for 4.4.

Issue #2458, opened 2024-10-28
portrait overlapping


ssw940521:
## The problem

**Describe the bug**
I have two charater with two styles and the only difference in the style is portrait position.When the character changes ,the previous portrait will under the new one.

**To Reproduce**
![image](https://github.com/user-attachments/assets/a7eec9dc-bef8-4730-8723-febc14baea43)
![image](https://github.com/user-attachments/assets/5bc38524-8dd5-471e-a982-91e3e3665af9)
![image](https://github.com/user-attachments/assets/3d6200d1-4a4e-42d7-9977-8404c1fced38)
![image](https://github.com/user-attachments/assets/5dc9c7dc-b560-433e-94b1-43359350d85c)
![image](https://github.com/user-attachments/assets/440f8f6e-aaf9-4b81-aaa2-f265a77ef309)

**Screenshots**
![image](https://github.com/user-attachments/assets/7d5d177c-82aa-4204-970f-83751b55fd11)
![image](https://github.com/user-attachments/assets/8736e05a-00da-4699-8d71-b339e5f04ac4)
![image](https://github.com/user-attachments/assets/83a09e7b-edc1-4222-bb4d-85cead015623)

**System (please complete the following information):**
 - OS: windows11
 - Godot Version: [e.g. 4.3]
 - Dialogic Version: 2.0 Alpha 16

## Solutions

**Workaround**

reopen the project

**Possible fixes**

have no idea

ssw940521:
a reproduce video
https://github.com/user-attachments/assets/69fd51d9-332f-4f31-8d4e-bf1519603ad8

ssw940521:
the problem code maybe in subsystem_portraits.gd line :672 & 673:
![image](https://github.com/user-attachments/assets/50018c27-ac0d-418d-9e70-6949f4b723e5)
When await get_tree().process_frame is called, the timeline has moved forward, and the dialog box for the next text has already loaded. At this point, it overlaps with the previous one. 
After commenting out these two lines of code, the issue disappeared, but I'm not sure if this might cause other problems
@Jowan-Spooner

ssw940521:
My previous idea might have been wrong. I have a new idea now, but I'm not sure if it's correct:
When the style changes, the following code will be called:
![image](https://github.com/user-attachments/assets/68f34bb1-d859-4cce-b280-7fe6f94ff87a)
[...]

PR #2464, opened 2024-11-02
fix: portrait overlapping #2458 

ssw940521:
By switching to a new speaker before changing the style, I solved the issue of overlapping portraits. This way, when changing the style, the object calling change_speaker is already the next speaker.


Jowan-Spooner:
Thanks. And congrats on your first merged PR!

PR #2473, opened 2024-11-11
Fix file field clobbering path on unfocus without changes

salianifo:
The file event field would cause the path to get changed to just the file name without the directory when you accidentally click on the field and click off the field without changing anything. This changes adds a check to make sure the current value and field text value aren't the same before calling `_on_file_dialog_selected`, or, if `file_mode` isn't directory open mode, that the current value as just the filename isn't the same as the field's value.


Jowan-Spooner:
This looks very reasonable. Thank you for the PR!

PR #2479, opened 2024-11-13
Styles: Fix persistent info getting lost

Jowan-Spooner:
Now persistent info is always merged with the previously stored info. This means switching to a non-textbubble style doesn't clear all registers.


zaknafean:
This is similar to how the fixes worked in Dialogic 1, so its already battle tested! yay! 

Confirmed it works to do what it says it does.

PR #2480, opened 2024-11-14
Fix crash when creating DialogicLabelEvent in gdscript

shazzner:
Adding a label via gdscript will cause Dialogic to crash in `get_unique_identifier()` when attempting to `get_extension()` on a null `file_path`. This PR checks if it's null and returns `""`.


Jowan-Spooner:
Sorry this took so long, I was really inactive these past months. This looks like what you'd expect the method to return, so I'm fine with doing this, though I wonder if we should stop it before it get's there in your situation. 

Will merge anyways. Thanks!

PR #2485, opened 2024-11-20
Add Choices to `last_question_info`.

CakeVR:
This was missing, and the History Layer never displayed any Choices.


Jowan-Spooner:
Thanks!

PR #2486, opened 2024-11-21
Fix instancing new Character Nodes on Simple Highlight Portrait

CakeVR:
This portrait is simple and allows for only one image, there is no underlying node hierarchy that may need to change. Therefore, we need to ensure to not recreate a new character node.

We should look at handling this behaviour, to keep highlights working for other portrait scenes, that create a new character instance, too.

### DIFF
--- a/addons/dialogic/Modules/HighlightPortrait/simple_highlight_portrait.gd
+++ b/addons/dialogic/Modules/HighlightPortrait/simple_highlight_portrait.gd
@@ -19,6 +19,10 @@ func _ready() -> void:
 		self.modulate = unhighlighted_color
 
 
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
This seems correct. Looks like I changed the default for this method a while ago and didn't change this implementation. Thanks.

I think custom portraits that create new scenes every time should just check against the current speaker in _ready to generate the right highlight, but we could possibly add a helper method like is_current_speaker() either directly to the DialogicPortrait class or to the Portrait subsystem...

salianifo:
This uses spaces for indentation instead of tabs causing Godot not to parse the file correctly.

Issue #1306, opened 2022-11-06
Proposal: Unify Sound and Music event


Jowan-Spooner:
These events are currently VERY similar except:
- Music has a loop and fade option
- They use different audio players

It could make sense to make it one event. However this is just an idea. Would like to hear your thoughts. Thought of this as it's how it looks in @coppolaemilio's mockup (that is used in the readme).

thebardsrc:
one event would probly be fine, but they would still need two separate buses/channels/whatever (idk not familiar with how Godot audio works), for the purpose of players being able to adjust the volume and sound effect levels independently

zaknafean:
@KvaGram is probably a good one to have in this discussion.

Jowan-Spooner:
Well I would still have two modes (music and sound)  on the event. Audio bus can be set by the user anyways, but maybe adding a default (changeable in the settings) for both modes would be good as well.

KvaGram:
KvaGram the Autistic Duck reporting in.


> one event would probly be fine, but they would still need two separate buses/channels/whatever (idk not familiar with how Godot audio works), for the purpose of players being able to adjust the volume and sound effect levels independently
> 

The problem is with Dialogic. As @Jowan-Spooner pointed out, each even have their own player.
A player can only play one soundfile at a time. So if you want music and sound effects at the same time, there must be separate players. Same goes with voiceovers and type-sound effects. That's 4 players in all, unless I'm missing something.

If we were to unify the events, we would need to be able to define what player to use in the event. That requires aditional datafield in the UI of the event.
A common complaint of some of the events, is the clutter.

We could have dialogic just have a pool of players, using them one by one. That would solve the problem of having to set one. But that would introduce a new problem. What if you wanted to stop one specific track?
There are fixes to that too, but that again would mean adding another datafield to an event.
There is no way around that.

> For the voice overs too, for that matter, thafs one people would want to control as well. Three separate things with independent volume

I am dreaming of means to better control voice overs. I am kinda out of it right now. But when I get my shit together and start coding again, I got some ideas I itch to try out.

> Well I would still have two modes (music and sound) on the event. Audio bus can be set by the user anyways, but maybe adding a default (changeable in the settings) for both modes would be good as well.

The audiobus don't matter in this case. We already have an option for the audiobus, well it's more like a compulsory field that defaults to the master-bus. It already has to be set. Again, the limit is in dialogic's audio players.

KvaGram:
Here is my proposal to fix this.
[...]

PR #2488, opened 2024-11-25
Combine Music/Sound event into single Audio event

salianifo:
### _Only test on backup timelines, not backwards compatible!_

Closes: #2082
Closes: #1306 

This PR combines the Music and Sound events into a new Audio event. It will automatically read the old sound/music event formats and convert them into the new audio event format.

#### Features
- Named audio channels
    - Channel names do not need to already exist, they are created dynamically
    - Channel names may contain `a-z A-Z 0-9 _ -`, but they may not be only `-`
    - Default settings for `fade`, `volume`, `audio bus`, and `loop` can be configured for named channels in the Audio section of the Dialogic Settings page, or updated/added by clicking the favorite icon next to the channel name in the event itself
    - Selecting/typing a channel name with default settings will override the current settings when the channel dropdown loses focus
    - Playing audio on a channel that is already playing audio will cross-fade from the previous audio to the new audio
- One-Shot SFX channel
    - This channel is used for short sounds that don't need to be tracked
    - Can play as many audio files as needed simultaneously
- Sync with channel
    - Allows you to start an audio file at the same position as the currently playing audio on a different channel
    - Can be used for layered audio tracks
    - Cannot be used with WAV files imported using Ima-ADPCM compression due to a [Godot bug](https://github.com/godotengine/godot/issues/18878)
- Looping implemented for WAV files

#### Text Format
```
# Plays "file_path" on channel_name with settings in shortcodes,
# defaults for anything not in shortcodes
audio channel_name "file_path" [shortcodes]

# Plays "file_path" as a One-Shot SFX with the default settings
audio "file_path"

# Stops audio playing on channel_name
audio channel_name -

# Stops all One-Shot SFX
audio -
```

#### Missing Features
- Syntax highlighting for the text timeline editor
- Code completion for the text timeline editor

#### Additional Changes Made
- Dynamic options field now has a new `ANY_VALID_STRING` mode that accepts a `validation_func` in the `extra_info` paramter of the event editor build functions
    - Expected return for the validation function is a `Dictionary` with `valid_text` and `error_tooltip` as keys, which are both optional if the text is already valid
    - If `valid_text` exists in the return from the `validation_func`, the `current_value` will be updated


Jowan-Spooner:
Notes on my changes for completeness:
- Removed the ability to change a channels defaults from the event. Instead just have a button to go to the audio settings
- Volume, Fade, Bus, Loop and Sync are now considered override settings. This makes the workflow with the channel defaults feel more logical.
- Added syntax highlighting and code completion
- Various stuff I did while working on this 🙈

PR #2497, opened 2024-12-24
Check for UIDs in strings as well on `start_timeline`

coppolaemilio:
If you use an exported variable with a path to a file (let's say a timeline):

`@export_file var timeline = ""`

Godot will automatically use a UID path (something like `uid://d0kqf1bb2ev0` instead of `res://timeline.dtl`) file to avoid having issues when moving that file in the future. This check in Dialogic is only expecting for the string `res://` in the argument, so if you do something like `Dialogic.start(timeline)` it will fail to run.

The fix just adds a check for UID as well. 

I suspect there are many other places where this is happening, so it would be good to replace all those instances of checking for "res://" to also check for "uid://"


Jowan-Spooner:
Thanks, this looks good to me.

PR #2499, opened 2024-12-28
Added focus_choice_button_node function to Choices subsystem

DJLink:
Recently added this function to my own code base for something, thought it could be useful for OG codebase, the possibility of setting a different choice focus later after showing the question.

User could set this for example to be called after choices are shown by getting the signal question_shown.

```
Dialogic.Choices.focus_choice_button_node(button_index);
```

### DIFF
--- a/addons/dialogic/Modules/Choice/subsystem_choices.gd
+++ b/addons/dialogic/Modules/Choice/subsystem_choices.gd
@@ -191,7 +191,10 @@ func show_current_question(instant:=true) -> void:
 	if missing_button:
 		printerr("[Dialogic] The layout you are using doesn't have enough Choice Buttons for the choices you are trying to display.")
 
-
+func focus_choice_button_node(button_index:int) -> void:
+	var node: DialogicNode_ChoiceButton = get_choice_button_node(button_index)
+	if node:
+		node.grab_focus();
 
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
Hi @DJLink, this looks pretty good. I thought I had added a similar helper method for selectting/pressing/activating the focused choice before but now I can't find it. 
I would consider shortening those method names (the addition of button_node seems a bit over-kill).
- focus_choice()
- get_choice_button()
- select_choice() 
- select_focused_choice()
the last two don't exist, but I think would make sense, so feel free to add them if you want.

Also no ; at the end of lines please.

DJLink (commented):
renamed to select_focused_choice

Jowan-Spooner:
Hey @DJLink, it took me a while, but I finally found the time to take another look at this. I decided to go ahead and implement the other helper methods I mentioned, and rename the focus one as well.

So now the Choices subsystem has these methods:

- focus_choice(index)
- get_choice_button(index)
- select_choice(index) 
- select_focused_choice()

Which should all be very useful if one tries to do some custom input logic regarding choices.

Thanks again for this PR, hope you're doing well 👋

DJLink:
Awesome, thanks for checking and your work on this. Hope you are well too. Cheers

PR #2501, opened 2024-12-31
Ignore directories ignored by godot (.gdignore file)

jrb0001:
Godot ignores directories with a `.gdignore` file, but Dialogic still scans them.

In case of the android build template, Godot copies all files into `./android/build/assets` and the gradle build copies them again. In my case Dialogic had 3 copies of every timeline and every character. The name of those duplicates can change inside `project.godot` which is annoying for version control.


zaknafean:
Hm thats interesting! I ran into this issue ALOT in Dialogic 1 and Godot 3, and couldn't personally figure out why multiple copies were appearing in the Android build. Will try to test this one asap.

PR #2503, opened 2025-01-03
Feat: background file can now be loaded from user:// scope too

orangemelon69:
Godot resource load works only for imported Resources and background changer accepted only res:// paths before. This change allows for an image from user:// path to also be used.


zaknafean:
Thanks a lot for the feature! It looks good and will be merged.

PR #2513, opened 2025-01-11
Fix a typo when applying italic font in text bubbles

mmatyas:
I was just browsing through the code, and noticed a small typo.


Jowan-Spooner:
Thank you very much!

PR #2523, opened 2025-01-28
Add Signal text_started that aligns with the text appearing

Snowdrama:
The `await` on line 124 can take some time depending on if the textbox fades in or has an animation when opening, and because of that the `about_to_show_text` signal has a delay before the text begins to render.

Delay from before the await to after the await measured by `Time.get_ticks_msec()`
image
```gdscript
dialogic.Text.about_to_show_text.emit({'text':final_text, 'character':character, 'portrait':portrait, 'append': is_append})

print("About to await dialogic.Text.update_textbox %s" % Time.get_ticks_msec())
await dialogic.Text.update_textbox(final_text, false)
print("Done Awaiting dialogic.Text.update_textbox %s" % Time.get_ticks_msec())
```

If you use `about_to_show_text` to enable an animation, as an example for lip flaps, the delay makes it look awkward because mouth begins moving before the text begins rendering. This PR adds a signal after the await so that it's closer to when the text starts rendering. 

Here's an example where I swap between a normal mouth to an `AnimatedSprite2D` for the character to talk:
```gdscript
extends Node2D

#this goes on the group in a custom layered character scene that has the mouth

@export var normal_sprite : Sprite2D
@export var talking_sprite : AnimatedSprite2D

func _ready():
	Dialogic.Text.about_to_show_text.connect(start_talking)
	# The text_started signal is a drop in replacement that feels much better for things
	# That need to respond to the text being drawn
	# Dialogic.Text.text_started.connect(start_talking)
	Dialogic.Text.text_finished.connect(stop_talking)
	
func start_talking(info:Dictionary):
	normal_sprite.visible = false
	talking_sprite.visible = true
	
func stop_talking(info:Dictionary):
	normal_sprite.visible = true
	talking_sprite.visible = false
```
I saw while I was researching there was similar methods to this that may have been in previous versions like `started_revealing_text` but those seem to have been removed or something.

### DIFF
--- a/addons/dialogic/Modules/Text/event_text.gd
+++ b/addons/dialogic/Modules/Text/event_text.gd
@@ -127,7 +127,9 @@ func _execute() -> void:
 			_try_play_current_line_voice()
 			final_text = dialogic.Text.update_dialog_text(final_text, false, is_append)
 
-			#this is after the await for the textbox so the text starts rendering here
+			# This is after the await for the textbox so it's after any text box 
+			# opening animations that may make animations reliant on knowing when
+			# the text starts 
 			dialogic.Text.text_started.emit({'text':final_text, 'character':character, 'portrait':portrait, 'append': is_append})
 
 			_mark_as_read(character_name_text, final_text)
--- a/addons/dialogic/Modules/Text/event_text.gd
+++ b/addons/dialogic/Modules/Text/event_text.gd
@@ -127,6 +127,9 @@ func _execute() -> void:
 			_try_play_current_line_voice()
 			final_text = dialogic.Text.update_dialog_text(final_text, false, is_append)
 
+			#this is after the await for the textbox so the text starts rendering here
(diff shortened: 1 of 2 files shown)


Jowan-Spooner:
I like this, though we really need to improve the documentation on all those signals, cause now the names alone are a bit confusing.

Snowdrama:
> I like this, though we really need to improve the documentation on all those signals, cause now the names alone are a bit confusing.

I definitely can understand, the `about_to_show_text` signal is a bit confusing for sure because if the text window isn't open that await means it's like "preparing to show text" while revealing the textbox. The second time it's called though the text box is already open so it more correctly corresponds to the text appearing, definitely unintuitive.

I can definitely look at updating the signal documentation with descriptions to help clarify the difference in the future!

Jowan-Spooner:
Okay I put in a bunch more docs comments. Thank you again for this PR!

PR #2530, opened 2025-02-12
Immediately load persistent info to prevent overwritting

### LINKED ISSUE #2526: `register_character` calls for existing characters are ignored during the same frame the layout scene is created
## The problem

**Describe the bug**
If you have previously called `register_character()` and re-register the character as a new node on the same frame that `Dialogic.start()` is called, the character will not be updated. Visually, this causes text bubbles to float around the original node instead of the new one.

**To Reproduce**
1. Start a timeline with `Dialogic.start()`
2. Call `register_character()` on the layout.
3. Complete the timeline.
4. Start a new timeline and register a new node as the **_same character_** on the **_same frame_** that `Dialogic.start()` is called.

**Expected behavior**
The character is registered as the new node.

 - OS: (Platform independent)
 - Godot Version: 4.3
 - Dialogic Version: 2.0 Alpha 16

## Solutions

**Workaround**

1. Call `Dialogic.start()`.
2. Register the character(s).
3. `await Engine.get_main_loop().process_frame`
4. Register the character(s) again.

**Possible fixes**

This issue is caused by a race-condition between `DialogicLayoutBase._enter_tree()` and `text_bubble_base.register_character()`. Currently, the following happens when `Dialogic.start()` and `register_character()` are called on the same frame:

1. `register_character()` updates `registered_characters`.
2. `_enter_tree()` is invoked by Godot, calling `text_bubble_base._load_persistent_info()` and overwriting `registered_characters` with old data.

To fix this, `_load_persistent_info()` can just not call `register_character()` is new data is already present by adding `and character not in registered_characters` after the `is_instance_valid()` check:

```
func _load_persistent_info(info: Dictionary) -> void:
	var register_info: Dictionary = info.get("textbubble_registers", {})
	for character in register_info:
		if is_instance_valid(register_info[character]) and character not in registered_characters:
			register_character(character, register_info[character])
```

This fix seems pretty straightforward, but I do not know if there are other scenarios where `_load_persistent_info()` should overwrite `registered_characters`, so I wanted to create this issue before any PR.

HuntJSparra:
Fixes #2526 by calling `_load_persistent_info()` in `_init()` instead of `_enter_tree()`. Previously, `_enter_tree()` was called the frame after `Dialogic.start()`, causing new changes (e.g., `register_character` to be overwritten with the old persistent info.

The fix here is more general than the one proposed in #2526 (which only addressed text bubbles). During testing `_init()` was called after `_exit_tree()`, so there should be **_no_** race conditions with saving and loading persistent info on the same frame.

### DIFF
--- a/addons/dialogic/Resources/dialogic_layout_base.gd
+++ b/addons/dialogic/Resources/dialogic_layout_base.gd
@@ -58,7 +58,7 @@ func _apply_export_overrides() -> void:
 #region HANDLE PERSISTENT DATA
 ################################################################################
 
-func _enter_tree() -> void:
+func _init() -> void:
 	_load_persistent_info(Engine.get_meta("dialogic_persistent_style_info", {}))
 
 


Jowan-Spooner:
Thank you very much!

Issue #572, opened 2021-10-08
method to call to close dialog window of currently opened timeline


Zoo4125:
**Is your feature request related to a problem? Please describe.**
I want the dialog window to close if I walk away too from my NPC  in my game, without having to click the default action key again.

**Describe the solution you'd like**
a method to call upon that closes the currently opened timeline dialog window

**Describe alternatives you've considered**
telling the Godot to press my default action key after leaving an area 2d that the player is in.

**Additional context**
thank you for your time :)

Jowan-Spooner:
So I guess you want to have the usual fade out animation and agree, that would be a cool thing. 

If you just want to remove the dialog, you can just delete the node with queue_free().

MatthewHinds:
New to Dialogic, was this ever addressed? I'm currently using `Dialogic.end_timeline` and it's working fine, just wish it had the closing animations.

Jowan-Spooner:
This can be done with end_timeline() now

PR #2536, opened 2025-02-16
Implement proper timeline ending logic

### LINKED ISSUE #2509: Audio doesn't stop at timeline end
## The problem

**Describe the bug**
If you add music to your timeline, the music doesn't stop at the end of the timline. It also doesn't stop if the timeline is ended with the call Dialogic.end_timline()

**Workaround**

Would be great to get a workaround for the time being

Jowan-Spooner:
Dialogic now uses a timeline to perform the end of dialog. By default this timeline simply contains a default clear event, but it can easily be swapped out. 
This should fix #2509 and #572 because `end_timeline()` now does an actual nice closing by default.

### DIFF
--- a/addons/dialogic/Core/DialogicGameHandler.gd
+++ b/addons/dialogic/Core/DialogicGameHandler.gd
@@ -70,6 +70,10 @@ var paused := false:
 
 			dialogic_resumed.emit()
 
+## A timeline that will be played when dialog ends.
+## By default this timeline only contains a clear event.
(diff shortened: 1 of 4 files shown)


Purplotl:
@Jowan-Spooner I have a setup with a global audio player for background music, and a pause menu that pauses, clears, ends timeline and boot to main menu. For some reason with this setup end timeline does not stop typing sounds from playing.

image

image

PR #2537, opened 2025-02-17
Improve runtime resources (timelines and characters created through code)

### LINKED ISSUE #2491: Create a character in gdscript without requiring saving a local file
In the game I'm developing, all data is created and stored in a sqlite database, including any scripts. This is due to the fact that the game is very data-heavy. Generally this works fine; although, there are a few quirks I have to work around. Unfortunately, I'm not able to create characters in the same manner as it requires creating a new Character in gdscript then saving it as a character resource locally. This causes issues as I have to ship individual dch files for updates and track them separately from my data.

Should be able to declare a character, their portrait, and any additional properties in gdscript without needing to save the character resource locally. Then reference that character in any subsequent timelines.

As far as I know, there isn't a way to create a character without any disk writes. Instead of the data-driven approach I'd prefer, I have to create each character manually in the editor or via a long gdscript which requires a save to file locally. The other alternative is to use an entirely separate system for just the portraits and manually call show/hide functions in the timeline script.

**Additional context**
I understand this is an unusual case from the standard dialogic game; however, as more games get developed with Dialogic that an established architecture pattern, such as being data-driven, that this feature would see more use.

### LINKED ISSUE #2492: jump not working on timeline created by code
## The problem

Using jump node in a timeline created by code results on error that says:
`Trying to assign value of type 'Nil' to a variable of type 'String'.`

Stacked Fotograms:
`0 - res://addons/dialogic/Core/DialogicResourceUtil.gd:65 - at function: get_unique_identifier`
`1 - res://addons/dialogic/Modules/jump/event_label.gd:22 - at function: _execute`

**To Reproduce**
1. Create a scene with a Node2D.
2. Create a script with:
```
func _ready():
	var timeline: DialogicTimeline = DialogicTimeline.new()
	var evento = get_event(preguntas)
	timeline.events = evento
	timeline.events_processed = true
	Dialogic.start(timeline)
	Dialogic.timeline_ended.connect(_on_dialog_ended)
	
func _on_dialog_ended() -> void:
	Dialogic.timeline_ended.disconnect(_on_dialog_ended)
        print("dialogue completed correctly")
        
func get_event() -> Array:
	var event: Array = """
	hellooo
	jump test
	good bye!
	
	label test
	test text
	""".split("\n")
	return event
```
3. Assign the script to the scene node created before.
4.  Press *Execute current escene* button (F6)

**Expected behavior**
I expected to get the correct sequence of dialogue:
- hellooo
- test text
- good bye!

**Screenshots**
imagen_2024-12-03_121627174

 - OS: Windows 11
 - Godot Version: 4.3.estable
 - Dialogic Version: 2.0. Alpha 16

## Solutions

I didn't reach any solution

I tried write jump/label "test", jump/label TEST, jump/label Test

**Possible fixes**

I have no idea sorry.

Jowan-Spooner:
This makes both `DialogicTimeline` and `DialogicCharacter` extend `DialogicIdentifiableResource` which basically adds a get_identifier and set_identifier method. If an identifier is set at runtime for a non-saved resource, the resource is registered to the directories directly (without a path) allowing for the use of those resources in timelines from then on.

```gdscript
        var characterA := DialogicCharacter.new()
	characterA.display_name = "Code Character"
	characterA.color = Color.WEB_PURPLE
	characterA.set_identifier("CodeCharacter")

	var timelineA := DialogicTimeline.new()
	timelineA.from_text(
	"""
		CodeCharacter: Timeline A!!!
		jump CodeTimelineB/
		label Section2
		CodeCharacter: About to end timeline A.
	""")
	timelineA.set_identifier("CodeTimelineA")

	var timelineB := DialogicTimeline.new()
	timelineB.from_text(
	"""
		Timeline B!!!
		jump CodeTimelineA/Section2
	"""
	)
	timelineB.set_identifier("CodeTimelineB")

	Dialogic.start(timelineA)
```
- fixes #2492 
- fixes #2491

This also once again allows using "invalid" character names in timelines (a new character resource will be created on the spot) and will now even interpret the portrait as a color for those, which allows to have one-off characters say something. However the portrait picker is not shown in the visual editor for those characters for now so this is kinda a text-editor exclusive rn.
```
"Random person" (webpurple): Wait who's that?
"Other person" (darkblue): I think that's the new guy.
```

Also fixes a bug with dictionaries as export overrides for custom portraits.

### DIFF
--- a/addons/dialogic/Core/DialogicGameHandler.gd
+++ b/addons/dialogic/Core/DialogicGameHandler.gd
@@ -439,6 +439,6 @@ func print_debug_moment() -> void:
 	if not current_timeline:
 		return
 
-	printerr("\tAt event ", current_event_idx+1, " (",current_timeline_events[current_event_idx].event_name, ' Event) in timeline "', DialogicResourceUtil.get_unique_identifier(current_timeline.resource_path), '" (',current_timeline.resource_path,').')
+	printerr("\tAt event ", current_event_idx+1, " (",current_timeline_events[current_event_idx].event_name, ' Event) in timeline "', current_timeline.get_identifier(), '" (',current_timeline.resource_path,').')
 	print("\n")
 #endregion
--- a/addons/dialogic/Core/DialogicResourceUtil.gd
+++ b/addons/dialogic/Core/DialogicResourceUtil.gd
@@ -63,7 +63,7 @@ static func add_resource_to_directory(file_path:String, directory:Dictionary) ->
 
 ## Returns the unique identifier for the given resource path.
 ## Returns an empty string if no identifier was found.
-static func get_unique_identifier(file_path:String) -> String:
+static func get_unique_identifier_by_path(file_path:String) -> String:
 	if not file_path: return ""
 	var identifier: Variant = get_directory(file_path.get_extension()).find_key(file_path)
 	if typeof(identifier) == TYPE_STRING:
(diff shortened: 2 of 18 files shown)


shazzner:
Great work!

PR #2554, opened 2025-03-14
Choices: Add override for custom choice buttons

HuntJSparra:
The choice buttons for the `Centered Choices` and `Textbubble Layer` layers can be replaced with a `PackedScene` descending from `DialogicNode_ChoiceButton`. This is intended for users who want custom behavior for their buttons (the choice must be pressed 3 times, on-hover animations, multi-`Control` UI, etc.).

**New Settings:**
_Centered Choices:_ `Choices/Behavior/Custom Button`, `Choices/Behavior/Maximum Choices`
_Textbubble Layer:_ `Choices/Behavior/Custom Button`, `Choices/Behavior/Maximum Choices`

**Changes to `DialogicNode_ChoiceButton`:**
The on-pressed behavior of the button can be customized by overriding `_pressed()`. By default, this preserves the existing behavior by emitting the new `choice_selected` signal when the button is pressed. The Choices Subsystem has now listens for the button's `choice_selected` signal instead of `pressed`.

**Changes to VN/Centered Choice Layer:**
The centered choice layer now instantiates its buttons when the style is changed (functionally similar to Textbubble Layer) so that custom buttons are applied, including while the timeline is running.

### DIFF
--- a/addons/dialogic/Modules/Choice/node_choice_button.gd
+++ b/addons/dialogic/Modules/Choice/node_choice_button.gd
@@ -12,6 +12,10 @@ extends Button
 ## not supported on buttons at this point.
 
 
+## Emitted when the choice is selected. Unless overridden, this is when the button or its shortcut is pressed.
+signal choice_selected
+
+
 ## Used to identify what choices to put on. If you leave it at -1, choices will be distributed automatically.
 @export var choice_index: int = -1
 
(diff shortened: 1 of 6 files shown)


Jowan-Spooner:
Hey @HuntJSparra, sorry it took me so long.
This looks mostly good. My only requests would be to:
- Add a layer setting on the amount of choice buttons that get added. This seems more intuitive then having to change a number somewhere in the code.
- get rid of the _on_pressed method and connection in favor of buttons built in `_pressed` method
I can also make those changes if you don't have the time.

HuntJSparra:
Nothing to apologize for! Dialogic has been a great help to me, so the least I can do is show a little patience.

I'll rebase and make the changes sometime this coming week.

HuntJSparra:
@Jowan-Spooner There is now a layer setting for the number of choice buttons (defaults to 5 for text bubbles and 10 for centered choices) and `on_pressed`+`connect` has been replaced with `_pressed`, as you asked. I've also added a `printerr` if a custom button fails to load and moved the Centered Choice Layer settings to a new "Choices" group so they match Text Bubble Layer.

Everything should be ready for another review when you have the time, but I did have a question: should we pass the choice info or at least additional data to the choice buttons? This would allow custom choice events (or even the regular choice event) to pass data that could then affect a custom button's behavior (e.g., a choice costs 5 gold to select).

HuntJSparra:
I had not thought of overriding `_load_info()`, but that makes a lot of sense. I added some inline documentation before `_load_info` that mentions this. Do you want this also added to the class reference for `DialogicNode_ChoiceButton` or somewhere else in the online docs?

Jowan-Spooner:
Tha looks good to me for now. The class reference is generated from the built-in doc-comments so that should be good for now (might have to add a bunch of [br] to the doc comments, not sure). More tutorials in the online docs would be cool, but absolutely not necessary for this PR.

Jowan-Spooner:
I think this is ready to be merged, right @HuntJSparra?

HuntJSparra:
Yes. Looks good to me.

PR #2556, opened 2025-03-21
Apply new overrides when switching styles if there's a common ancestor

Galo223344:
A very simple change. When switching between styles if the new applied style inherits the old one then the exports of the old style get transferred to the new one, this PR makes this also happen when both old and new styles inherit from the same style.

The use-case, at least in our case, was that we had a different style for each character in our game, so every time the characters switched in the dialogue the whole layer was destroyed and rebuilt, causing problems with the textbox animations. So we made all the styles inherit from one base style that way when we switch between styles the exports simply change, but we had the problem that when changing from a character style to a different character style since the second style technically didn't inherit from the first one it didn't work, this PR fixes that.


Jowan-Spooner:
Looks good to me. Thank you!

PR #2571, opened 2025-04-15
fix one-off name getting lost in visual editor

### LINKED ISSUE #2570: one-off names break when using the visual editor
## The problem

**Describe the bug**
Visual editor breaks temporary names.
When filling in a character name inside of the visual editor, playing the timeline will not display those character names:

Additionally, setting up names in the text editor (which do work) and switching to the visual editor clears the names from the character field in the text events. If you play the timeline there and then, it appears to still work. However, when you switch back to the text editor, all the names have been removed.

**To Reproduce**
1. create a timeline and in the text editor create a text event like: "jimmy: my name is jimmy!" (important, jimmy does not have a character resource)
2. switch to the visual editor. (you should see that the speaker has changed to blank)
3. add or remove an event, let's say a text event that says "after a change in the visual editor, the name is gone"
4. switch back to the text editor. 
5. the text even should now have corrupted into "my name is jimmy!" (notice how the character prefix is missing?)
**Expected behavior**
temporary character names created in the visual editor should show up when playing the timeline and switching between editors should not delete the temporary character names 

**Screenshots**
example:
Image

will display the text fine, but will not have the name box

Image

 - OS: windows
 - Godot Version: [e.g. 4.4.1]
 - Dialogic Version: 2.0 alpha 17 (commit e098088)

## Solutions

**Workaround**

Working purely in the text editor goes around this issue, however it is not ideal because i no longer have the convenience that the visual editor gives me.

**Possible fixes**

I don't know, but if i were to make a guess, the event display builder might play a part in this problem.

Pheubel:
fixes: #2570 

This does not fix the bug where characters aren't set if you don't click them in the drop down. that it's own can of works

### DIFF
--- a/addons/dialogic/Modules/Text/event_text.gd
+++ b/addons/dialogic/Modules/Text/event_text.gd
@@ -325,12 +325,15 @@ func from_text(string:String) -> void:
 		else:
 			character = DialogicResourceUtil.get_character_resource(name)
 
-			if character == null and Engine.is_editor_hint() == false:
-				character = DialogicCharacter.new()
-				character.display_name = name
-				character.set_identifier(name)
-				if portrait:
-					character.color = Color(portrait)
+			if character == null:
+				if Engine.is_editor_hint() == false:
+					character = DialogicCharacter.new()
+					character.display_name = name
+					character.set_identifier(name)
+					if portrait:
+						character.color = Color(portrait)
+				else:
+					character_identifier = name
 
 	if not result:
 		return


Jowan-Spooner:
There seems to be an `if` missing in your code.

Pheubel:
> There seems to be an `if` missing in your code.

i was a little bit too fast with VS code, found that out as well and has been corrected

Jowan-Spooner:
Thanks!

Issue #2580, opened 2025-04-26
Style editor not working and Character portrait not showing up


Alex900v:
## Style editor not working

**Describe the bug**
I've started my first project with Dialogic. I've set up a timeline, characters and selected the premade visual novel style. But after choosing the colours and fonts in the style editor,  it's still showing the premade style.

**Expected behavior**
For the dialogue textbox to change to the colour I chose.
**Screenshots**

![Image](https://github.com/user-attachments/assets/a5ee0c73-c74b-4a87-8144-4dc03be93aed)

![Image](https://github.com/user-attachments/assets/ce6dc604-bc42-462f-880d-1ae8da7016c3)

![Image](https://github.com/user-attachments/assets/66d01ab0-b230-4496-849e-0b0a3c6ea141)

**System :**
 - OS: Windows
 - Godot Version: 4.4.1
 - Dialogic Version: 2.0 Alpha 17

Jowan-Spooner:
I think this is fixed on main. If not, please reopen!

PR #2582, opened 2025-04-27
Fix Custom Styles not working

### LINKED ISSUE #2577: Cannot change styles
## The problem

**Describe the bug**
Dialogic will not load custom styles, either through the 'Change Style' event or through the 'Test Style' button in the style editor. Dialogic always defaults to the standard visual novel style.

**To Reproduce**
- Create a new style in Dialogic's style editor
- Edit style so that it is visually distinct from any default style
- Attempt to preview this new style with the 'Test Style' button **OR**
- Use the 'Change Style' event in a Dialogic timeline with this new style selected

**Expected behavior**
Dialogic should display a timeline with the newly created style.

 - OS: Windows 10
 - Godot Version: 4.4.1, both direct download and Steam versions
 - Dialogic Version: 2.0-Alpha.17 WIP - Downloaded on 04/15/2025 from Github

## Solutions

**Workaround**
Dialogic does not appear to be locating the style resource file correctly. In the text editor for a timeline, manually supplying the full path of the style resource will cause the new style to load correctly.

E.g this does NOT work:

`[style name="New_File"]`

but this DOES work:

`[style name="res://styles/New_File.tres"]`

Jowan-Spooner:
Some mistakes on a recent PR made custom styles not be used at all 🙈.

Should fix #2577 and #2580 as well as a bunch of peoples issues on discord.

### DIFF
--- a/addons/dialogic/Modules/Style/DialogicStylesUtil.gd
+++ b/addons/dialogic/Modules/Style/DialogicStylesUtil.gd
@@ -8,7 +8,7 @@ static var style_directory := {}
 ################################################################################
 
 static func update_style_directory() -> void:
-	return ProjectSettings.get_setting('dialogic/layout/style_directory', {})
+	style_directory = ProjectSettings.get_setting('dialogic/layout/style_directory', {})
 
 
 static func build_style_directory() -> void:
@@ -24,9 +24,11 @@ static func build_style_directory() -> void:
 			continue
 		# TODO this is bad
 		var resource: DialogicStyle = load(style_path)
-		style_directory[resource.name] = resource
-	ProjectSettings.set_setting('dialogic/layout/style_directory', style_directory)
-	ProjectSettings.save()
+		style_directory[resource.name] = style_path
+
+	if Engine.is_editor_hint():
+		ProjectSettings.set_setting('dialogic/layout/style_directory', style_directory)
+		ProjectSettings.save()
 
 
 static func get_default_style_path() -> String:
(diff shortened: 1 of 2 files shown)


addons/dialogic/Modules/Style/DialogicStylesUtil.gd
```diff
@@ -46,16 +48,16 @@ static func get_fallback_style() -> DialogicStyle:
...
 static func get_style_path(name_or_path:String) -> String:
-	var path := ""
+	print(style_directory)
 	if name_or_path.begins_with("res://"):
 		if not ResourceLoader.exists(name_or_path):
 			name_or_path = ""
 
 	if name_or_path in style_directory:
-		path = style_directory[name_or_path]
+		name_or_path = style_directory[name_or_path]
```
nlupugla, line 57: This function seems like it has a side effect of modifying the input argument `name_or_path`. Is that intentional?

addons/dialogic/Modules/Style/DialogicStylesUtil.gd
```diff
@@ -46,16 +48,16 @@ static func get_fallback_style() -> DialogicStyle:
 
 
 static func get_style_path(name_or_path:String) -> String:
-	var path := ""
+	print(style_directory)
```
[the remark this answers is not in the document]
Jowan-Spooner: No, my bad 🙈 Will push a fix later!

addons/dialogic/Modules/Style/DialogicStylesUtil.gd
nlupugla: Makes sense. Worth testing quickly though. Strings can potentially be very large, so I would have assumed they are passed by reference in GDScript but happy to be proven wrong.

PR #2589, opened 2025-05-11
Runtime Resources: Improve runtime characters ability to have portraits

Jowan-Spooner:
Changes the portrait subsystem to index characters by their identifier instead of the resource path and removes an unnecessary threaded loading loop.

Also adds a helper method `character.add_portrait(name, image, _scene)` that allows code like this to very easily register a portrait:

```gdscript
        var characterA := DialogicCharacter.new()
	characterA.display_name = "Code Character"
	characterA.set_identifier("CodeCharacter")
	characterA.color = Color.WEB_PURPLE
	characterA.add_portrait("default", "res://characters/npc/portrait1.png")
	characterA.default_portrait = "default"

	var timelineA := DialogicTimeline.new()
	timelineA.from_text(
	"""
		join CodeCharacter center
		CodeCharacter: I have a portrait!
		
	""")
	Dialogic.start(timelineA)
```

### DIFF
--- a/addons/dialogic/Editor/Settings/CoreSettingsPages/settings_tools.gd
+++ b/addons/dialogic/Editor/Settings/CoreSettingsPages/settings_tools.gd
@@ -30,7 +30,6 @@ func execute_tool(method:Callable) -> void:
 
 	tool_thread = Thread.new()
 	tool_progress_mutex = Mutex.new()
-	print(method)
 	tool_thread.start(method)
 
 	await tool_finished_signal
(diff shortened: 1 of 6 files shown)


shazzner:
Hey @Jowan-Spooner thanks for the quick update! I tested your branch and I was able to create characters and link portraits using your helper method.

However, I seem to be having a new issue where when I end a script with a joined character I get an error that looks similar to the previous one:

```
E 0:00:18:293   subsystem_portraits.gd:617 @ get_joined_characters(): Resource file not found: res://liz (expected type: )
  <C++ Error>   Method/function failed. Returning: Ref<Resource>()
  <C++ Source>  core/io/resource_loader.cpp:344 @ _load()
  <Stack Trace> subsystem_portraits.gd:617 @ get_joined_characters()
                event_character.gd:130 @ _execute()
```

This results in the character portrait not loading again.
(liz is the name of the character, I believe it's trying to load the resource_path here again?)

In this method inside subsystem_portraits.gd:

```
## Returns a list of the joined charcters (as resources)
func get_joined_characters() -> Array[DialogicCharacter]:
	var chars: Array[DialogicCharacter] = []

	for char_path: String in dialogic.current_state_info.get('portraits', {}).keys():
		chars.append(load(char_path))

	return chars
```

shazzner:
That worked, thank you!

PR #2593, opened 2025-05-13
Fix reference to test-project

salianifo:
There were some path references to the `test-project` folder that doesn't exist, causing some error output, updated to their actual paths. I did a search to make sure there weren't any other instances of this, this was the only spot. I know it doesn't break anything, figured I'd fix it anyway.


Jowan-Spooner:
Thanks, good catch!

PR #2598, opened 2025-05-27
Visual Editor Performance Enhancements

Jowan-Spooner:
This PR is a bunch of experiments on my end to improve the performance of the visual editor loading times. 
It's been a while since I did these, here is the main points I remember:
- Includes a small fix for temporary characters in visual editor (similar to #2571)
- I made adjustments that allow running the dialogic editor as a scene. This scene has many problems, but it allows using the profiler. This mostly meant putting access to Editor classes/singletons into Engine.is_editor_hint() checks.
- Adds a cache to Editor Settings, improving read performance of those.
- Caches Event icons to prevent constant loading
- Reduces Calls to indent_events()
- Changes the way Extra-Spacing is added to the bottom of the visual timeline editor (because it got recalculated way to much)
- Reduced Calls to update_content_list
- increases the loading batch_size from 10 to 100 because that made things significantly faster (it waits a frame between each batch)
- preloads the common editor fields in the event_block script

There is also an attempt at colored icons in here (for characters) but I scrapped it because unless we have a character color cache this would mean loading all the characters just to get their colors. 

Testing and feedback would be very much appreciated.


zaknafean:
This thing still valid and need testing/updating @Jowan-Spooner ?

PR #2606, opened 2025-06-07
Assorted Bug Fixes

salianifo:
Fixes the following bugs:

- Audio loop default had incorrect value in the `update_audio` function causing looping to not work at all
- Portrait changing during Character Update event was not `await`ing `Portraits.change_character_portrait` which is now an asynchronous function due to threaded loading
- Play from here was incorrectly disabled in the visual editor on events with the `help_page_path == ""`


Jowan-Spooner:
Looks good to me. Thank you!

PR #2619, opened 2025-07-10
Fix spinbox issue: style setting for text_max_width can not be greater than 100

### LINKED ISSUE #2617: Style setting for `text_max_width` can not be greater than 100
## The problem

**Screenshots**

Image

The default `text_max_width` is `300`, so the `100` here is wrong. And users can not change this greater than 100. 
Many other properties also have this issue.

## Solutions

I find a bug here:

Image

line 439 should be `if property_info.hint_string == '':`
so I will do a PR.

jinyangcruise:
Fix: #2617


Jowan-Spooner:
Thank you very much. What an annoying change, this certainly worked at some point, not sure when it broke.

PR #2621, opened 2025-07-10
Fix text_bubble run _process() before starting dialog

### LINKED ISSUE #2620: text_bubble will run `_process()` before starting a dialog
## The problem

**Describe the bug**
If not any dialog triggered, the text_bubble's `_process()` still runs every process frame.
This is because it is not initialized to set_process(false). This will cause a lot of useless computing.

**To Reproduce**
modify `text_bubble.gd` and print something in the `_process()` function.
Run a scene with bubble dialog.

**Expected behavior**
`_process()` should not be called when dialog not shown.

 - OS: Windows 11
 - Godot Version: v4.5.beta2.official
 - Dialogic Version: 2.0 Alpha 17

## Solutions

I will do a PR.

jinyangcruise:
Fix:  #2620

### DIFF
--- a/addons/dialogic/Modules/DefaultLayoutParts/Layer_Textbubble/text_bubble.gd
+++ b/addons/dialogic/Modules/DefaultLayoutParts/Layer_Textbubble/text_bubble.gd
@@ -36,6 +36,7 @@ func _ready() -> void:
 
 
 func reset() -> void:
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
This makes sense. Thanks.

PR #2642, opened 2025-07-22
DialogTextNode: add a resource file checker before loading; there was…

endaye:
DialogTextNode: 
Add a resource file checker before loading. 
There was an error when `custom_bbcode_effects`(in `Settings/Text/Custom BBCode Effects`) is empty.

image
image


Jowan-Spooner:
This has been fixed in #2645. Thank you very much anyways!

PR #2643, opened 2025-07-23
Make StringNames also a valid type for timeline (which already accepts String).

fcolecumberri:
Make `StringNames` also a valid type for string valid variables.

I encountered that if a `StringNames` is passed as a timeline (instead of a `String`), the game crashes. Made no sense.


Jowan-Spooner:
Looks good to me. Thanks.

PR #2647, opened 2025-07-26
Add option to update CSV files without erasing content for default locale, original content is replaced instead (#2644)

timkrief:
This makes it possible to change update text for default locale directly from the CSV files.

It also makes it possible to change the default locale used in dialogic and update the text accordingly.

https://github.com/user-attachments/assets/221bfae5-c0ea-4ea8-be83-6ec4887aeb9a


timkrief:
would fix #2644

PR #2655, opened 2025-08-31
Fixes: The "move_to_foreground()" method is deprecated, use "grab_focus()" instead.

bobbens:
I believe it was deprecated in Godot 4.3.
Referenc: https://github.com/godotengine/godot/issues/82936

### DIFF
--- a/addons/dialogic/Editor/Common/reference_manager_window.gd
+++ b/addons/dialogic/Editor/Common/reference_manager_window.gd
@@ -163,7 +163,6 @@ func open() -> void:
 	DialogicResourceUtil.update_directory('dch')
 	DialogicResourceUtil.update_directory('dtl')
 	popup_centered_ratio(0.5)
-	move_to_foreground()
 	grab_focus()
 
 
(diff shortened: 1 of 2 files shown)


Jowan-Spooner:
Thanks for the PR!
4.3 is still supported though and this seems like not enough of a reason to drop it. Maybe you could leave a #TODO comment and add a warning_ignore instead of removing the old call?

bobbens:
From my understanding

> Thanks for the PR! 4.3 is still supported though and this seems like not enough of a reason to drop it. Maybe you could leave a #TODO comment and add a warning_ignore instead of removing the old call?

Unless I'm mistaken, it seems like `grab_focus` exists at least since 4.0 (see https://docs.godotengine.org/en/4.0/classes/class_window.html#class-window-method-grab-focus ). So this is a case of two functions doing the same thing, and one being retired in favour of another. I don't think it'll break anything and just quiet the warning on newer godots.

Jowan-Spooner:
Oh I see, sorry! In that case this is correct. I thought functionality was changed. Thanks for the clarification.

Jowan-Spooner (approved):

PR #2665, opened 2025-09-18
Godot 4.5 Compatibility: UID Fixes and remove SVGTexture references

uncleLukie:
## Summary:
This pull request resolves critical errors and warnings when using Dialogic with the newly released Godot 4.5. The primary fixes include updating the deprecated SVGTexture to the new DPITexture and correcting invalid resource UIDs.

## Changes:
- SVGTexture to DPITexture: In Godot 4.5, the SVGTexture class has been renamed to DPITexture. This was causing an error (Cannot get class 'SVGTexture') when loading scenes that used the old class name. This has been fixed in addons/dialogic/Editor/Events/Fields/field_file.tscn by replacing all instances of SVGTexture with DPITexture.

- Invalid UID Fixes: The timeline_editor.tscn file contained incorrect UIDs for the timeline_editor.gd and shortcut_popup.gd scripts. This has been resolved by updating the UIDs to match the actual file UIDs, which prevents any potential issues with script loading and scene integrity.

### DIFF
--- a/addons/dialogic/Editor/Events/Fields/field_file.tscn
+++ b/addons/dialogic/Editor/Events/Fields/field_file.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=7 format=3 uid="uid://7mvxuaulctcq"]
+[gd_scene load_steps=6 format=3 uid="uid://7mvxuaulctcq"]
 
 [ext_resource type="Script" uid="uid://buepm260xnmaa" path="res://addons/dialogic/Editor/Events/Fields/field_file.gd" id="1_0grcf"]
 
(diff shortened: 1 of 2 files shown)


CakeVR:
If `DPITexture` is not supported in older Godot versions, Dialogic would cease to work for a great range of users. Therefore, that class name change is not a solution we can accept.

Jowan-Spooner:
Hey @uncleLukie thanks for the PR! As cake said both DPI Texture and SVGTexture (which was only available for a short while during the 4.5 dev phase aren't recognized by older godot versions. But I made a commit just removing the icons from the scene as they are loaded at runtime anyways. I guess godot changed what texture type get_theme_icon() returns. This way this should be good to go.

Thanks again!

PR #2676, opened 2025-10-06
Cache defaults for unique scenes which use the same script

cliveglitch:
Hi, I got an annoying to find, but in context... 

I was playing around with dialogic so I can learn to use it and I'm structuring my project in such a way that I have a base character script which is shared between multiple character scenes, like these ones:

image
image

Found out while experimenting, when I have some props which I don't modify in dialogic's character menu, the first character who joins the timeline fills the cache metadata with their defaults, while whatever other characters join next will just use the first character's defaults and that's the basically the problem.

I tested putting a name string property for the characters, something like "@export var dialogicName: String", which would have the strings "Krauss" for Krauss and "Natsuhi" for Natsuhi, and while playing the timeline and checking with the debbuger, because Krauss joined first, Natsuhi was getting Krauss's defaults and getting "Krauss" name as the "dialogicName" prop.

The fix was simple, instead of saving the cache by script path, just save by scene node path... I don't know the dialogic's codebase fully so I'm open for other alternatives.

### DIFF
--- a/addons/dialogic/Core/DialogicUtil.gd
+++ b/addons/dialogic/Core/DialogicUtil.gd
@@ -36,7 +36,7 @@ static func autoload() -> DialogicGameHandler:
 
 #region FILE SYSTEM
 ################################################################################
-static func listdir(path: String, files_only:= true, _throw_error:= true, full_file_path:= false, include_imports := false) -> Array:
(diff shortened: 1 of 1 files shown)


Jowan-Spooner:
Thanks! This seems sensible, I wonder why I didn't do it like this. Will try to think it through once more later, but seems good rn.

PR #2679, opened 2025-10-10
fix: Bug where enum string exports would show as file exports because of handling of bitwise conditionals.

cliveglitch:
I'm playing with dialogic so I can develop a game in the future, and I found another issue where my character scenes had two string enum export props which were treated as file exports strangely enough. I went days finding out why and I found out, with the help of a debugger cause tbh I'm trying to understand bitwise operations, that some bitwise operation conditionals were behaving weird.

Bitwise operations, like bitwise AND, return a number not a boolean, and that number can be either a 0, which in a godot conditional would be treated as false, or anything else like 1, 2, 3, 50... which would be treated as true for godot.

image

In this case when you have a string enum property export that has a hint of PROPERTY_HINT_ENUM with value of 2 or 0010 in binary, and you make a bitwise AND operation with the global constant PROPERTY_HINT_DIR with value of 14 or 1110, the result is 2 or 0010 because the 3nd bit from left to right match... and godot takes that as true, so it treats the enum as a file prop.

You should compare the result with the global constant enum you want the usage to be, and in this case do something like "property_info['hint'] & PROPERTY_HINT_DIR == PROPERTY_HINT_DIR".

I went out of my way to change every bitwise operation so to standardize it, although if you ask how something like this didn't make any more bugs besides this... it's probably because you can mostly get away with it with certain global constant enums, for example:
PROPERTY_USAGE_EDITOR has a value of 4 or 100 in binary, which means the AND operation can ONLY be 0 or 4.
PROPERTY_USAGE_GROUP has a value of 64 or 1 000 000 in binary and can only be 0 or 64.
PROPERTY_USAGE_SUBGROUP value 256... well, let's just say it can only be 0 or 256.

Maybe in retrospect it's not that you're doing an improper comparison, the reason of this bug is because of how the global enum for PropertyHint is weird and doesn't go in exponentials of 2s like all the other enums which would absolutely make the comparison "value & ENUM = ENUM" unnecessary... well I guess compare it just in case there's a pesky bitwise enum that can make undesirable behaviors.

Dude, what a rabbit hole I went into...

---

Here is some testing, I used this export prop to break it:

@export_enum("Idle", "Run", "Jump", "Shoot")
var state: String = "Idle"

Here is an example broken:
image

Here is the example with the changes from this PR:
image

### DIFF
--- a/addons/dialogic/Core/DialogicUtil.gd
+++ b/addons/dialogic/Core/DialogicUtil.gd
@@ -305,7 +305,7 @@ static func apply_scene_export_overrides(node:Node, export_overrides:Dictionary,
 		return
 	var property_info: Array[Dictionary] = node.script.get_script_property_list()
 	for i in property_info:
-		if i['usage'] & PROPERTY_USAGE_EDITOR:
+		if i['usage'] & PROPERTY_USAGE_EDITOR == PROPERTY_USAGE_EDITOR:
 			if i['name'] in export_overrides:
 				if str_to_var(export_overrides[i['name']]) == null and typeof(node.get(i['name'])) == TYPE_STRING:
 					node.set(i['name'], export_overrides[i['name']])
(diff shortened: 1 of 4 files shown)


Jowan-Spooner:
Thanks for this fix @cliveglitch 🙏, I think this makes sense and is a great find. I use bitwise operations so rarely that it's not surprising I missed this detail 🙈.

PR #2686, opened 2025-10-16
Improve macOS shortcuts

PadraigK:
Fixes #2681 by handling macOS shortcuts keys. I tested this on my local Mac and it works well.

### DIFF
--- a/addons/dialogic/Editor/TimelineEditor/VisualEditor/timeline_editor_visual.gd
+++ b/addons/dialogic/Editor/TimelineEditor/VisualEditor/timeline_editor_visual.gd
@@ -1060,10 +1060,10 @@ func _input(event:InputEvent) -> void:
 			_add_event_button_pressed(DialogicTextEvent.new(), true)
 			get_viewport().set_input_as_handled()
 
-		"Ctrl+Shift+T", "Ctrl+Alt+T", "Ctrl+Option+T", "Shift+Command+T", "Option+Command+T", "Option+Command+A":  # Add text event with current or previous character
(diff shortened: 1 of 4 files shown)


addons/dialogic/Editor/TimelineEditor/VisualEditor/timeline_editor_visual.gd
```diff
@@ -1056,42 +1056,42 @@ func _input(event:InputEvent) -> void:
...
-		"Ctrl+T": # Add text event
+		"Ctrl+T", "Command+T":  # Add text event
 			_add_event_button_pressed(DialogicTextEvent.new(), true)
 			get_viewport().set_input_as_handled()
 
-		"Ctrl+Shift+T", "Ctrl+Alt+T", "Ctrl+Option+T": # Add text event with current or previous character
+		"Ctrl+Shift+T", "Ctrl+Alt+T", "Ctrl+Option+T", "Shift+Command+T", "Option+Command+T", "Option+Command+A":  # Add text event with current or previous character
 			get_viewport().set_input_as_handled()
 			var ev := DialogicTextEvent.new()
 			ev.character = get_previous_character(event.as_text() == "Ctrl+Alt+T" or event.as_text() == "Ctrl+Option+T")
```
Jowan-Spooner: This line needs to be adjusted as well I would think. From what I can tell, the "Ctrl+Option+T" thing is also just relevant for mac, so you can just replace it with "Command+Option+T", if that makes sense (and remove the Ctrl+Option+T above). 

Also curious why these all have command as the second part, I assume you tested it, so it must be right, but it feels a bit strange for godot to store it this way.

Jowan-Spooner (approved):

PadraigK:
> Thanks!

My pleasure! Thanks for your work on Dialogic ❤️

PR #2687, opened 2025-10-20
fix '+= operator does not work with strings '

### LINKED ISSUE #2601: += operator does not work with strings but + does
## The problem

**Describe the bug**
Using += to add to a string does not work and generates an unclear error message.

**To Reproduce**
1. In a timeline, assign text to a string variable
`set {var.var1} = "Hello"`
2. In a second line, add text using +=
`set {var.var1} += ", world."`
3. In a third line, add text using + inbetween two strings
`set {var.var1} = {var.var1} + " darkness, my old friend.`

Result: {var.var1} == "Hello darkness, my old friend." 

**Expected behavior**
+= and + should both concatenate strings

**Screenshots**
n/a

 - OS: MacO@
 - Godot Version: 4.4.1
 - Dialogic Version: 2.0 Alpha 17 WIP

## Solutions
**Workaround**

use the format {var} = {var} + "string"

**Possible fixes**
-

adrientremblay:
Changed the logic slightly for DialogicVariableEvents to work and not print an error when an event tries to add  a string to a string variable. The standard "Set Variable event failed because one value wasn't a float!" error will be printed if the user tries to add a non-string to a string variable.

Also updated the editor warnings to not show an error when a string is being added to a string.

_My changes work, but feels slightly invasive/messy. Thoughts? I can try and clean it up._

fixes #2601


adrientremblay:
Sick!

PR #2704, opened 2025-12-19
Simple fix for zero-value pitch errors in the typing-sounds node.

NGermany:
- Simple fix for zero-value pitch errors in the typing-sounds node

###  **Problem**
If the pitch variance for a sound file is set to anything greater than or equal to the base pitch value this will continuously produce errors in Godot 4, due to pitch being zero.  These errors flood the console and likely degrade performance, as well as prevent any sound from being played at random intervals. This is not the expected behavior.

I ran into this issue in my own projects but the fix is extremely simple. It is only one value in one line of code.

###  **Fix**
This PR is a fix to prevent the errors by setting the minimum value for pitch above zero. 
I took the value from Dialogic 1.x code ("0.0001") and applied it to current. Nothing more is in this PR.


Jowan-Spooner:
Thank you!

PR #2711, opened 2026-01-10
Fix Save Game Loading Bug

ootoovak:
A game using Godot was crashing when loading a saved game. This conditional check fixes the crash.

The game was being loaded with the following script:
```gdscript
extends Node2D

func _ready() -> void:
	if Dialogic.Save.has_slot(GlobalData.slot_to_load):
		await get_tree().process_frame
		Dialogic.Save.load(GlobalData.slot_to_load)
		GlobalData.slot_to_load = ""
	else:
		Dialogic.start("res://timelines/first-scene.dtl")
```

I don't anticipate this will be the final fix given if this is the pattern the is_character_joined branch could possibly also benefit from this extra check. Or, if this is just hinting at a deeper fix that needs to be done.

In any case I wanted to share what worked and use this PR as a starting point for feedback in hopes of eventually getting a fix into main.

(Using Godot 4.5 on macOS btw.)


zaknafean:
@Jowan-Spooner Is this one still needed after the changes in #2695

Issue #2424, opened 2024-10-03
Full quality backgrounds


mak448a:
**Is your feature request related to a problem? Please describe.**
When I load a background, it seems to be more pixelated than the original artwork. (I used AI artwork, but that's beside the point)

**Describe the solution you'd like**
Could you add a toggle in the UI to use the full quality version without pixelating the artwork?

**Describe alternatives you've considered**
Just leaving it. Honestly, it's not too much of a problem, but it would be nice to fix it.

**Additional context**
Here is the original picture and below is the dialogic picture.

![Screenshot_20241003_143924](https://github.com/user-attachments/assets/ab4d513d-f2cc-47fb-965d-81f3d375536a)
![Screenshot_20241003_143926](https://github.com/user-attachments/assets/3158aaaf-fc92-4ad7-a7a4-639b35d96bb0)

PR #2722, opened 2026-01-15
Background Layer options

zaknafean:
Fix for #2424
Added adjustable settings to the background layer to override project settings for texture filtering and texture repeating.

Its a pretty straightforward PR, but the style editor 'hacks' to make it work are a touch annoying.

Line 180 and line 185 in Subsytem_backgrounds do use a parent.get_parent() i don't quite like, but is pretty safe in this context, but if there is a cleaner way to get to the background layer I'd rather do that. In addition I could pass the parameters to the background_holder, but that seemed a bit more annoying to do. Thoughts?


Jowan-Spooner (approved):
Would be fine merging it like this, although I did not actually test this with images, so I just assume this actually fixes it, just looked over it from implementation/code standpoint.

PR #2730, opened 2026-01-22
Support and Hooks for custom subsystems

zaknafean:
Three additions all related to some Multiplayer work I'm dealing with, but have other broad applications as well

1.  Adds the ability to override base Subsystems with custom extensions 
2.  A warning if you accidentally name an extension a 'reserved' word
3.  Adds an RPC enabled hook to the Dialogic Game handler and a flag to identify it has been called 

So 1 and 2 are actual bugs that most people just haven't run into yet. Previous to this if you named your extension a 'reserved' subsystem, it would make a random name for the node, which could very well break your extension if you weren't ready for it. The warning is there to avoid that situation, but now if you DO have a naming conflict, it'll assume you're doing it on purpose and override the subsystem. This allows you to extend subsystems with custom functionality quickly and easily without messing with the subsystem code in the slightest!

3, by itself, just allows people to start a dialogic timeline for all or some connected peers if called properly (_Dialogic.start_mp.rpc(<timeline name>)_ ). With that in place you could use my multiplayer extension or build your own very easily. Since normal _Dialogic.start_ allows for Variants as an argument, we can't just make that RPC aware, we need to enforce primitives, so for combability a separate function is needed.

This one definitely needs a bit of review. The code is simple but it may need some supporting changes in other subsystems to be used cleanly, such as error messages concerning already connected signals, but I wanted it up and out there right now.

### DIFF
--- a/addons/dialogic/Core/DialogicGameHandler.gd
+++ b/addons/dialogic/Core/DialogicGameHandler.gd
@@ -66,6 +66,9 @@ var paused := false:
 ## By default this timeline only contains a clear event.
 var dialog_ending_timeline: DialogicTimeline
 
+## Flag to identify if current timeline is network enabled
+var is_mp_timeline : bool = false
+
 ## Emitted when [member paused] changes to `true`.
 signal dialogic_paused
 ## Emitted when [member paused] changes to `false`.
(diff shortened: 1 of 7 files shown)


zaknafean:
Better late then never! Implemented and tested the last selection, and the changes still work with the other typing changes that have been implemented since Jan. As a reminder this:

- Allows for the overriding of existing subsystems using the extension system (writing custom 
- Gives a warning if you try to name an event/subsystem to be the same as an existing Dialogic event or subsystem

PR #2735, opened 2026-01-29
Fix Wait Event not pausing when Dialogic is paused

salianifo:
As discussed on Discord, the Wait event did not pause when Dialogic was paused, as it did not use a subsystem. This PR moves the active part of the Wait event to a subsystem so it can hook into the `pause()` and `resume()` functions.


salianifo:
I had troubles with the Tween not having a good way to tell whether it was paused or stopped, so I switched to a Timer (not a SceneTreeTimer since it can't be paused directly).

PR #2754, opened 2026-03-26
Assign button sounds to the correct audio bus

mrspeaker:
Currently Button Sounds are played through the default (master) audio bus. Button Sounds play the typing sound when hovering over a dialog option. This PR assigns the bus to the same user setting as Type Sound (in `/Modules/Text/node_type_sound.gd`).

This fix was for my personal project where I noticed that muting the sound-effects audio bus did not stop the hover typing sounds in dialog options.

I'm not sure why the assignment to the bus needs to be guarded with `if bus == 'Master'`, but this is what was done in `node_type_sound.gd`.


zaknafean:
This should be changed and is consistent with the code in node_type_sound.gd so a merging it will go.

PR #2764, opened 2026-05-11
Objectdb debbuger snapshot null error fix

vrime:
In Godot 4.6 using ObjectDB profiler snapshot causing error by trying to access null variable
Screenshot_102

### DIFF
--- a/addons/dialogic/Modules/Variable/subsystem_variables.gd
+++ b/addons/dialogic/Modules/Variable/subsystem_variables.gd
@@ -178,6 +178,7 @@ func _set(property, value) -> bool:
 ## Allows to get dialogic built-in variables
 func _get(property):
 	property = str(property)
+	if not dialogic: return
 	if property in var_storage.keys():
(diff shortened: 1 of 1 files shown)


zaknafean:
Can you go into just a bit more detail on what this fixes exactly, and what circumstances cause the error to fire? Is it everytime you try to take a object db snapshot>

vrime:
Pretty much, yeah, not quite sure why is it causing an error, but every time i tried to took a ObjectDB snapshot it caused null error, as you can see on screenshot, at the moment of taking snapshot dialogic is null, so i tried to add check that verifies if dialogic currently null or not, and it seems to be working fine

zaknafean:
So this fix DOES fix what's in alpha 19, and the crash is confirmed within alpha 19, but the code was changed in the 'pre alpha 20'/main branch already and the bug is no longer reproducible there.

Thanks a lot for figuring out a fix for it, but it looks like its already fixed moving forward.

PR #2768, opened 2026-05-23
Added Emit Event to allow emitting Signals from Autoloaded classes

HelixTheSage:
Added new Emit Event - very similar to the Call Event, but it's structured to allow emitting Signals from Autoloaded classes rather than calling their methods. Added corresponding get_autoload_signal_suggestions to Core/DialogicUtil.


zaknafean:
Gonna leave this here for Jowan to review, but my gut says this should be an extension rather than in the core code.

PR #2769, opened 2026-05-25
fixed 'Not all code paths return a value.' error for godot 4.7

da-pwo:
the new godot versions now requires returning in all circumstances instead of having the interpreter figure it out by itself, so this just has them return what they wouldve returned automatically in previous versions and should also work with said previous versions as no new features are used

### DIFF
--- a/addons/dialogic/Modules/Text/node_name_label.gd
+++ b/addons/dialogic/Modules/Text/node_name_label.gd
@@ -23,3 +23,5 @@ func _set(property, what):
 		else:
 			name_label_root.show()
 		return true
+	else:
+		return false
--- a/addons/dialogic/Modules/Text/node_name_label.gd
+++ b/addons/dialogic/Modules/Text/node_name_label.gd
@@ -23,5 +23,4 @@ func _set(property, what):
 		else:
 			name_label_root.show()
 		return true
-	else:
-		return false
+	return false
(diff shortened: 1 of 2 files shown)


urssomansso (approved):

addons/dialogic/Modules/Text/node_name_label.gd
```diff
@@ -23,3 +23,5 @@ func _set(property, what):
 		else:
 			name_label_root.show()
 		return true
+	else:
```
coppolaemilio: I don't really think you need the else here, you can just go straight to the return at the end of the function

addons/dialogic/Modules/Variable/subsystem_variables.gd
```diff
@@ -183,6 +183,8 @@ func _get(property):
 			return VariableFolder.new(var_storage[property], property, self)
 		else:
 			return DialogicUtil.logical_convert(var_storage[property])
+	else:
```
coppolaemilio: same here, you can do without this else

addons/dialogic/Modules/Variable/subsystem_variables.gd
```diff
@@ -244,6 +246,8 @@ class VariableFolder:
 				return VariableFolder.new(data[property], path+"."+property, outside)
 			else:
 				return DialogicUtil.logical_convert(data[property])
+		else:
```
coppolaemilio: and here

zaknafean:
Made the suggested changes from Emi as this is a bit critical and I want to get it merged. Thanks for taking the initiative on fixing and testing this all~

PR #2783, opened 2026-07-14
Fix background subsystem changed emit with invalid argument field

Scrawach:
Regression #2695

I'm using a custom implementation for the background holder, which is connected to `background_changed` event. Everything worked fine in the `alpha-19` release, but there is a bug in the master branch: `background_changed` signal emit data with invalid `argument` field.


zaknafean:
Oh! Quite the copy paste error there. Good find and good fix. Thank you.

PR #2805, opened 2026-08-16
Fix crash on save load when a character has no portrait node

atirna:
Fixes #2801

Loading a save right after a scene change (or over a previously used slot) could crash with:

```
Invalid access to property or key 'test_cleo' on a base object of type 'Dictionary[String, Node]'
```

As the issue suspects, joined characters are tracked in two places: `portraits` (exported, so it is packed into saves and restored by `unpack_state`) and `character_nodes` (runtime only, rebuilt in `_load_state`). Those two can get out of sync:

- after a scene change the portrait nodes are gone, but the restored `portraits` entries say the character is joined, so `is_character_joined()` passes and every `character_nodes[...]` access afterwards hits a missing/freed key (this is the exact line from the issue report)
- in `_load_state`, `add_character()` can fail (e.g. the position container is not available after a scene change) and returns null, but the node was still accessed unconditionally
- `_clear_state` passed nodes that the scene change already freed into `_remove_character()`

What i changed, all in `subsystem_portraits.gd`:

- `is_character_joined()` now only reports a character as joined while it also has a valid entry in `character_nodes`. This is the single predicate most consumers already check (get_character_node, leave_character, highlighting, change/update methods), so stale entries now fail safely everywhere instead of just moving the crash one call further (a null node would previously reach `_animate_node` and crash there)
- `leave_character()` cleans up an entry that lost its node instead of leaving it behind
- `_load_state()` skips the portrait state restore when re-joining the character failed
- `_clear_state()` skips nodes that were already freed

Verification (headless Godot 4.5.1 + gdUnit4):

- on current main the new test suite fails with the exact reported error (`Invalid access to property or key ... Dictionary[String, Node]` in `get_character_node`) plus the follow-up null crash in `_animate_node`, and the runner aborts
- with this patch `Tests/Unit/subsystem_portraits_test.gd` passes (5 tests covering missing key, freed node, leave cleanup, failed re-join on load, clear with freed nodes), full existing `Tests/Unit` suite stays green (21 tests, 0 failures)

### DIFF
--- a/Tests/Unit/subsystem_portraits_test.gd
+++ b/Tests/Unit/subsystem_portraits_test.gd
@@ -1,0 +1,104 @@
+extends GdUnitTestSuite
+## Regression tests for dialogic-godot/dialogic#2801:
+## Dialogic.Save.load() crashes when a joined character has no portrait node
+## (e.g. after a scene change), because the portraits subsystem can hold a
+## joined-character entry whose portrait node is gone or was never created.
+
+
+var portraits
+
+
+func before_test() -> void:
+	DialogicResourceUtil.update()
+	portraits = Dialogic.Portraits
+	if portraits == null:
+		push_error("test setup failed: Dialogic.Portraits subsystem not available")
+		return
+	# Start from a clean subsystem state.
+	portraits.portraits.clear()
+	portraits.character_nodes.clear()
+
+
+func after_test() -> void:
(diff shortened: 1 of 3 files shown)


atirna:
closing this one out for now, its been sitting a while and i dont want to leave a stale branch on your queue. happy to reopen or rebase it if its still useful to you.

PR #2806, opened 2026-08-22
Fix textbox visibility update logic

pazalla:
Move text animation logic to show block in textbox visibility


Jowan-Spooner:
I think this is correct and was left there by mistake. 

🎉 Thanks @pazalla for the find and the fix!

PR #2807, opened 2026-08-24
Fix infinite loop when an event block has a Vector field with conditional visibility

Herrsch:
## Issue:
When creating a custom Dialogic event, if the event block has a parameter that is a Vector2, Vector3 or Vector4, and that parameter has a visibility condition, Godot will print 1049 errors to the console whenever that visibility condition becomes true.

It seems like the editor is getting caught in an infinite loop when calculating the parameter visibility for the timeline UI. The output prints this five times: 
```  ERROR: res://addons/dialogic/Editor/TimelineEditor/timeline_editor.gd:19 - Stack overflow. Check for infinite recursion in your script.
  ERROR: res://addons/dialogic/Editor/Events/EventBlock/event_block.gd:395 - Stack overflow. Check for infinite recursion in your script.
  ERROR: res://addons/dialogic/Editor/Events/EventBlock/event_block.gd:395 - Stack overflow. Check for infinite recursion in your script.
  ERROR: res://addons/dialogic/Editor/Events/EventBlock/event_block.gd:395 - Stack overflow. Check for infinite recursion in your script.
  ERROR: res://addons/dialogic/Editor/Events/EventBlock/event_block.gd:395 - Stack overflow. Check for infinite recursion in your script.
```
followed by this 1024 times: ```ERROR: modules/gdscript/gdscript.h:529 - Stack underflow! (Engine Bug)```

## Screenshots
<details>
<summary>Expand</summary>
image
image
image
</details>

## The Fix:
- Removed `_on_value_changed` signal from `_set_value` in `field_vector_base.gd`

Credit to @bazsupport for finding this fix.

I believe that when `event_block.gd` is looping in `recalculate_field_visibility`, it calls `_set_value` on the vector fields, which emits an `_on_value_changed` signal, which causes the UI to start updating again, bringing it back into the same loop.

The other `field_` scripts do not emit an `_on_value_changed` signal from `_set_value`, and in my testing, removing this line does not affect the Vector fields' ability to set and propagate their values.

### DIFF
--- a/addons/dialogic/Editor/Events/Fields/field_vector_base.gd
+++ b/addons/dialogic/Editor/Events/Fields/field_vector_base.gd
@@ -28,7 +28,6 @@ func _load_display_info(info: Dictionary) -> void:
 
 func _set_value(value: Variant) -> void:
 	_update_sub_component_text(value)
-	_on_value_changed(value)
 
 
 func _on_value_changed(value: Variant) -> void:


Jowan-Spooner:
Seems reasonable. Thank you for the find and fix @Herrsch and @bazsupport 🎉 !

PR #2810, opened 2026-08-29
Fix inherited style overrides and VN name label alignment

### LINKED ISSUE #2698: Something is broken with inherited styles
## The problem

**Describe the bug**
Hard to describe concisely, look at the 8th and 10th reproduction steps to see the description of some symptoms

**Steps to reproduce the behavior:**
Fresh Godot download, new blank project, fresh Dialogic install (pasted the addons folder from the zip straight into the project directory).
1. Make 2 characters, characterA and characterB
2. Make new premade VN style, let's call it styleA
3. Make inherited style from styleA, let's call the inherited style InheritedStyle (I did this by clicking +, Inherited Style while styleA is selected in the UI)
4. Inside styleA, in the VN textbox, set the alignment of the name label to "Right". Now InheritedStyle also shows that its name label is aligned to the right
5. Assign styleA to characterA, and InheritedStyle to characterB
6. Make the Display Name of characterB noticeably longer than for characterA
7. make new timeline where characterA speaks first, then characterB, then characterA. No character joining or anything, just 3 texts plain and simple
8. Click "Play Timeline" and witness the following:
- characterA speaks first, name label is on the right (as expected)
- characterB speaks next, name label on the left (should actually be on the right)
- characterA speaks last, name label is on the left (?) and also the size of the name label remains as large as characterB's display name, so there is extra empty space

**Expected behavior**
I expected that InheritedStyle would have the name label on the right, just like is written in its settings. Also expected that the name label of styleA wouldn't suddenly switch sides

Bonus bug that I just discovered:
9. In InheritedStyle, set the name label alignment to the left, so basically override the setting of the parent style (styleA)
10. Click "Play Timelie" and watch this happen:
- characterA speaks first, name label is on the right (as expected)
- characterB speaks next, name label is missing altogether (!)
- character A speaks last, but now the name label is wrongly positioned and is as wide as characterB's display name (see screenshot)
Image

These are the bugs that I encountered in this toy example, but in my main project I also witnessed a character not speaking in the custom font that I set for it, but then the character that speaks next speaks in that font. So the font got transferred to the character that speaks next, similar to my description above where the length of the name label got transferred to the character that speaks next.

**System**
 - OS: Windows 11
 - Godot Version: 4.5.1.stable
 - Dialogic Version: 2.0-Alpha-18 (Godot 4.3+)

## Solutions

**Workaround**
I encountered this bug in an ongoing project so I tried to remake all my characters' styles from scratch and reassign them to my characters. Didn't work.

**Possible fixes**
idk man, I'm stumped

Minato-chen:
## Problem

Inherited styles did not always save or apply their settings correctly.

If a child style was missing local layer data, changing a setting could fail without showing an error. The Style Editor also treated inherited values as if they were saved directly in the child style.

When switching between inherited styles during a dialogue, Dialogic could use the wrong layer ID. This caused some settings, such as name label alignment, to return to their default values.

The VN textbox name label also kept anchors and offsets from the previous style. After several style changes, the label could appear in the wrong position.

## Changes

- Missing local layer data is created before saving a child override.
- The Style Editor now keeps local overrides separate from inherited values.
- Runtime style changes use the complete inherited layer list.
- VN textbox name labels use Godot's layout presets to reset their anchors and offsets before applying their configured position.

## Testing

I tested:

- normal styles without inheritance;
- Base → Child → Grandchild inheritance;
- adding and resetting child overrides;
- repeated changes between Left, Center, and Right name label alignment;
- short and long character names;
- style changes after resizing the game window.

Fixes #2698

### DIFF
--- a/addons/dialogic/Modules/DefaultLayoutParts/Layer_VN_Textbox/vn_textbox_layer.gd
+++ b/addons/dialogic/Modules/DefaultLayoutParts/Layer_VN_Textbox/vn_textbox_layer.gd
@@ -203,13 +203,21 @@ func _apply_name_label_settings() -> void:
 	else:
 		name_label_panel.self_modulate = name_label_box_modulate
 	var dialog_text_panel: PanelContainer = %DialogTextPanel
-	name_label_panel.position = name_label_box_offset+Vector2(0, -40)
-	name_label_panel.position -= Vector2(
+	var label_offset := name_label_box_offset+Vector2(0, -40)
+	label_offset -= Vector2(
 		dialog_text_panel.get_theme_stylebox(&'panel', &'PanelContainer').content_margin_left,
 		dialog_text_panel.get_theme_stylebox(&'panel', &'PanelContainer').content_margin_top)
-	name_label_panel.anchor_left = name_label_alignment/2.0
-	name_label_panel.anchor_right = name_label_alignment/2.0
-	name_label_panel.grow_horizontal = [1, 2, 0][name_label_alignment]
+
+	var alignment_preset: Control.LayoutPreset = [
+		Control.PRESET_TOP_LEFT,
+		Control.PRESET_CENTER_TOP,
+		Control.PRESET_TOP_RIGHT,
+	][name_label_alignment]
+	name_label_panel.set_anchors_and_offsets_preset(
+		alignment_preset,
+		Control.PRESET_MODE_MINSIZE
+	)
+	name_label_panel.position += label_offset
 
 
 ## Applies all text settings to the scene.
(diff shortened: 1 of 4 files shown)


Jowan-Spooner:
Hi @Minato-chen thank you for this PR. It works correctly. However as I'm still working on a pretty substantial rework of styles (#2738) I think I won't merge this for now, instead I'll try to make sure that my PR also works correctly in this scenario. I'll leave it open until that PR makes it into main, just in case these changes still make sense like this afterwards.

Minato-chen:
@Jowan-Spooner Thanks for testing it and for the update. That makes sense.

I'll leave this PR open while #2738 is in progress. I'm also happy to test the same inheritance and runtime style-switching scenarios again once the rework is ready.

